PR #20, opened 2018-07-04
fix_typo

wakiyamap:
I think it is probably a mistake......


martinboehm:
Thank you for the fix!

PR #21, opened 2018-07-05
Viacoin

abllr:


romanornr:
Don't agree with merging this. 

Calling via instead of viacoin. (inconsistency in naming because altcoins are written fully with their name)

missing Auxpow  !

Please see https://github.com/trezor/blockbook/pull/23

PR #29, opened 2018-07-19
Ethereum classic, backend

petrkr:


martinboehm:
We have completely reworked the way debian packages are created:

1) there is only one file which defines the coin packages: `<coin>.json` in `configs/coins`. The file defines all coin, backend and blockbook specifics. The packages are then generated from this file.
2) we have added verification of the backend binary, so that the package is not created with faulty or malicious binary. The verification is controlled by `"verification"` and `"verification_source"` entries in the config json file.

Please merge the new build from `master` to your contribution.

What you will have to do:
- create `configs/coins/ethereum_classic.json` file (base it on `configs/coins/ethereum.json`) and fill in the specifics.
- remove all debian package config files in `contrib/backends`
- check that the verification of the binary is correct (the build will fail if it is not)

To make the packages, run `make all-ethereum_classic` in the root folder.

petrkr:
Did new backend regards to new way.

Check it. I already built deb package and seems to be working at my end.

martinboehm:
Looks good, thanks!

PR #30, opened 2018-07-22
[RFC] Groestlcoin support

yura-pakhuchiy:
This work-in-progress pull request adds Groestlcoin (GRS) support (breaking other coins). Purpose of this PR is to receive feedback how to proceed.

Changes required for GRS support are in https://github.com/trezor/blockbook/commit/b7962476a5d928d5411266e79c9f79059b71ae96:
1. Single SHA256 for tx hashing.
2. Double Groestl for checksum part of address. I have just disabled checksum check for now. Do we really need to implement groestl hash in go; or disabled check is enough? Blockbook fetches data directly from groestlcoind and Trezor will check address again if user signs transaction, validating address in blockbook seems a bit excessive.
3. groestlcoind uses the same network magic values as bitcoind. Blockbook does not like this. I worked around this by providing fake values. Looks like it only needs unique values, not necessarily correct. Am I missing anything?

Any advises how to implement above in mergable way (also note some code which requires changes is in `vendor/`)?

I've setup servers compiled from this branch:
https://209.250.247.173:9145/test.html - mainnet
https://209.250.247.173:19145/test.html - testnet

All operations from testpage seem to work, except:
1. subscribeAddressTxid(). I do not know why this happens ATM, ZMQ port is correct.
2. Fee estimation sometimes work, sometimes does not because of lack of transactions. groestlcoind returns in this case:
```
# /opt/coins/nodes/groestlcoin/groestlcoin-cli -conf=/opt/coins/nodes/groestlcoin/groestlcoin.conf estimatesmartfee 10
{
  "errors": [
    "Insufficient data or no feerate found"
  ],
  "blocks": 10
}
```

I've left some coins on SLIP-14 seed for both mainnet and testnet. Electrum is available at:
https://github.com/Groestlcoin/electrum-grs/releases/


yura-pakhuchiy:
I've also tried specifying https://209.250.247.173:9145/ as backend in wallet.trezor.io. It fails with "inconsistent state" error. Looks like wallet also checks base58 checksum on addresses and does not like it. Could anyone from wallet team to look on this? Disabled base58 checksum check for Groestlcoin on beta-wallet would be extremely helpful (and updated trezor-common to allow T2 firmware 2.0.8 for groestlcoin in wallet).

martinboehm:
Thanks for your contribution!
I will address the issues one by one:

1) if I understand your changes correctly, the transaction structure (or address structure) in Groestlcoin is different from Bitcoin and cannot be parsed by btcsuite/btcd library (without changes). The better approach in this case is to not to do tx parsing on the Blockbook side but use parsed transactions returned from backend. We have used this approach in Zcash, you can use it as a reference. In general, please do not touch sources of other coins or general blockbook sources, keep all your changes to Groestlcoin only.

2) network magic values: this is unfortunate and happens in case of other coins as well. Please change the network magic values to anything unique, Blockbook uses the magic values only as a unique identifier in a map.

3) estimate smart fee - it is fine if it returns error in case of lack of transaction. Wallet knows that it can happen (happens for other coins too) and handles it well.

4) about the wallet problems, I have passed your issue to the wallet developers, they will look into it

yura-pakhuchiy:
@martinboehm Thank you for your reply! I'm really appreciate it.

Could you please comment on single SHA in tx signing as well?  Relevant change (only one line):
https://github.com/trezor/blockbook/commit/b7962476a5d928d5411266e79c9f79059b71ae96#diff-f55ea7b60b3904bacca5b7f17799b589R314
It does not affect GRS addresses, but transaction hash. Blockblook calculates transaction hash from rawtx using this code. As result getAddressTxids() and getAddressHistory() are failing for GRS with unpatched btcsuite. How can I replace tx hash function in an acceptable way?

martinboehm:
Blockbook does not compute tx hashes. The code you point to is in the btcd/wire library, which I suggest not to use for  Groestlcoin. Please use the approach we took in Zcach - use parsed tx data returned from backend. The only disadvantage to this approach is that the initial synchronization is a bit slower, however Groestlcoin does not have as many transactions as Bitcoin so the performance will be acceptible.

prusnak:
Sorry, but this is just too much of changed code, including the vendored dependencies. :-/ You are free to experiment with blockbook, but currently we don't have resources to deal with such big change in our backend and/or our wallet.

yura-pakhuchiy:
@prusnak sure, I opened this PR only to get feedback. I will make new PR soon which does not touch vendor.

PR #34, opened 2018-08-04
fix blockbook package

wakiyamap:
With the following command `backend-bitcoin` is required at package creation, I can't install `blockbook` of altcoins.
`make all-<coin>`


martinboehm:
Thank you!

PR #40, opened 2018-08-21
Use github paths for import

karel-3d:
Without this, `go get github.com/trezor/blockbook` does not work

Generated by calling this

`find . -iname "*.go" -print0 | xargs -0 sed -i 's/\t"blockbook/\t"github.com\/trezor\/blockbook/'`

(with GNU sed)


martinboehm:
Sorry but this is not the way we started and now build the project, I will non merge your PR. See the branch `docs`, which contains updated readme how to build the project.

PR #53, opened 2018-09-15
update monacoin version

wakiyamap:
monacoin 0.16.2 has been released so it is an update pull request.
Should I pull request to v0.0.7 ?

I confirmed that it will move once.
https://blockbook.electrum-mona.org/


martinboehm:
I would recommend you to wait couple of days, we are almost ready to merge `v0.0.7` and integration `tests` branches to `master`. Then you will have a more stable codebase for your pull request.

PR #63, opened 2018-09-27
fix monacoinparser

wakiyamap:
Since wallet address was issued by btcutil(1-prefix) toward near 40000block, I changed it to use monautil(M-prefix).


martinboehm:
I think it is not necessary to reimplement TryParseOPReturn, you can call btc.TryParseOPReturn.

wakiyamap:
Surely. I fixed it.

PR #68, opened 2018-10-02
Add Myriad

wlc-:
This PR adds support for Myriad.
Running instance here: https://139.162.175.108:9143


wlc-:
Ahh, thanks. Must have missed that change.
The latest commit includes time and size in ParseBlock.

PR #102, opened 2019-01-05
Add support for PIVX cryptocurrency

rikardwissing:
This PR adds support for the PIVX cryptocurrency into blockbook.

- [x] Add coin implementation 
- [x] Add integration and unit tests 
- [x] Deploy public server (available here: https://pivx-blockbook1.coinid.org)

Also noticed an inconsistency with how the block view fetches transactions compared to the transaction view. In order to display zerocoinspend inputs as unparsed addresses in block view, PR #100 needs to be merged. It is merely a cosmetic issue though and this works fine without it.


martinboehm:
Thank you for your contribution.

I want to clarify some things though:
1) is your public server https://pivx-blockbook1.coinid.org implemented with PR #100 or without? It seems to me that it is without but I just want to be sure.

2) Is the reason for PR #100 that the explorer shows incorrectly "No Inputs (Newly Generated Coins)" instead of "Unparsed address"?

3) The PR #100 is replacing call to` db.GetTxAddresses` by `GetTransaction` in `api.GetBlock`, however the same logic is used also in `api.GetAddress`, where you have not replaced it (for example there is differently displayed input in tx ae87b23e17be8... in tx detail https://pivx-blockbook1.coinid.org/tx/ae87b23e17be8dd12492ea339cc08417a4bd2ec5651d7ad47f74c8d217c47d24 in comparison to address detail https://pivx-blockbook1.coinid.org/address/DQ5fZeW7wJohJWQiuAEW6UDHNkY8oJgfWd and block detail https://pivx-blockbook1.coinid.org/block/1569966). However, please do not continue with PR #100, I will not merge it. The call to GetTransaction is much more expensive than db.GetTxAddresses (Blockbook must do RPC call to backend for each not cached tx in block/address). I have a different solution for you in the next point.

4) You could implement your own `OutputScriptToAddressesFunc` in the `pivxparser` (see `bcashparser` for reference how to register it). Your implementation could recognise the anonymous addresses from the output script and return them in format "Anonymous [hex marker]" or something similar (like this explorer https://chainz.cryptoid.info/pivx/block.dws?1569966.htm does) and call the standard Bitcoin OutputScriptToAddressesFunc for other address types.

rikardwissing:
1. It is implemented without PR #100 in order to make it clear that it works fine without it.
2. Yes, exactly.
3. Ok, I'll have to check why GetAddress works then. But as you said it might be better to use your suggestion in point 4.
4.  This sounds like a good solution. Will try it out later tonight 👍

rikardwissing:
I've implemented OutputScriptToAddresses in pivxparser and deployed it to the test server.

See: https://pivx-blockbook1.coinid.org/tx/ae87b23e17be8dd12492ea339cc08417a4bd2ec5651d7ad47f74c8d217c47d24

Zerocoinspend inputs are however still displayed as unparsed (or in address and block view as Newly Generated Coins). This is acceptable in my view. But if you have any suggestion on how to handle these inputs to get them to be displayed as anonymous I would be glad to explore it. They do not have any txid as a normal input non coinbase transaction.

martinboehm:
The problem is that the txid of the anonymous inputs (`0000000000000000000000000000000000000000000000000000000000000000`) is not found and Blockbook ignores the input during the import.

I have prepared a solution to this, which I tested in our implementation of the Liquid coin.
You will have to rebase your code to the new Blockbook version, which is currently in branch `indexv4`. This new version is almost ready to be released, it should be stable enough. Apart from changing some dependencies, all the PR code should work fine. You will have to reindex the DB as the format is different and incompatible. But before reindex implement methods `GetAddrDescForUnknownInput` and `outputScriptToAddresses` in a similar way it is done in the Liquid, see code below. In `GetAddrDescForUnknownInput` you can use `OP_ZEROCOINMINT` instead of `OP_INVALIDOPCODE` but that is up to you.

https://github.com/trezor/blockbook/blob/f109513464b3506484bcffde4270334f6d1c8311/bchain/coins/liquid/liquidparser.go#L75-L106

Also, when showing the Anonymous address, you should show the hex without the initial part of the script, only the data part and I don't know if the braces are necessary...

rikardwissing:
OK, I'll rebase it to the new blockbook version in branch indexv4.

About showing the beginning I thought it was nice to show it in the same way as the chainz block explorer. But i guess I could skip the opcode part atleast.

Any thoughts about how large part of the sig to show? Currently I show the first 32 characters.

martinboehm:
Do it any way you want to but since the opcode part is always the same, it does not bring much information. 

You don't have to abbreviate the sig at all, the field in the browser has style `overflow: ellipsis`, it will do it automatically (although it will push the amount to the next line).

rikardwissing:
I've rebased with indexv4, implemented the required changes and deployed to: https://pivx-blockbook1.coinid.org/

Worked great!
Awesome solution on your part!

I'll create a PR to indexv4 once I've checked that the integration tests still work fine.

martinboehm:
OK then I will close PRs #100 and #102 and wait for your new PR.

PR #105, opened 2019-01-11
Add support for PIVX

rikardwissing:
Follow up on PR #102 

- [x] Add coin implementation 
- [x] Add integration and unit tests 
- [x] Deploy public server (available here: https://pivx-blockbook1.coinid.org)

I needed to do a small change outside the coin specific files because getTxAddresses returns nil on anonymous inputs in disconnectTxAddresses and therefore crashes blockbook. So I added a simple check there.


martinboehm:
Good catch with the null txAddress, thanks.
Also, I updated PIVX ports to next available.

PR #111, opened 2019-01-29
Convert JsonValue to ValueSat when `"parse": false`

bitspill:
When not using the wire decoder all `vout` were being stored with a 0 value instead of the proper amount


martinboehm:
Thank you. It's a good catch. We normally do not use the non wire decoder and this slipped our attention.

PR #112, opened 2019-01-30
decode transaction from json in GetTransactionForMempool if `"parse": false`

bitspill:
Similar to https://github.com/trezor/blockbook/pull/111 it looks as if the condition of parseBlocks being false is not considered in the GetTransactionForMempool function

While poking around it also appears as if a few coins have `"parse": true` set but look to override `GetBlock` and `GetTransactionForMempool` such that the wire decoding is not actually used

I _think_ groestlcoin, zcash, koto, and liquid could each have their *rpc.go files edited to remove their stub GetTransactionForMempool functions if their configs are changed to `"parse": false`

So far I've tested this against Groestlcoin and Flo which I'm working on adding, I do not have the capacity on this dev machine to test the others however.

If I've not overlooked any underlying issues here and this PR is merged then at a later time the other stub functions can be removed from said coins and parse set to false as appropriate

### DIFF
--- a/bchain/coins/btc/bitcoinrpc.go
+++ b/bchain/coins/btc/bitcoinrpc.go
@@ -657,12 +657,12 @@ func IsMissingTx(err *bchain.RPCError) bool {
 // GetTransactionForMempool returns a transaction by the transaction ID
 // It could be optimized for mempool, i.e. without block time and confirmations
 func (b *BitcoinRPC) GetTransactionForMempool(txid string) (*bchain.Tx, error) {
-	glog.V(1).Info("rpc: getrawtransaction nonverbose ", txid)
+	glog.V(1).Info("rpc: getrawtransaction", txid)
 
-	res := ResGetRawTransactionNonverbose{}
+	res := ResGetRawTransaction{}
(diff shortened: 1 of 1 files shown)


martinboehm:
Thanks a lot, you are absolutely right. 
Wouldn't it be easier to implement the change like this?
```
func (b *BitcoinRPC) GetTransactionForMempool(txid string) (*bchain.Tx, error) {
	if !b.ParseBlocks {
		return b.GetTransaction(txid)
	}
	...
```
However, I would recommend not to continue with this PR, we are going to change the mempool code in more fundamental way and the function GetTransactionForMempool will either disappear or will be modified. But we will take into account your findings and make sure that the ParseBlocks flag is processed correctly. Then we will do also the cleanup of coins you mentioned.

bitspill:
> Wouldn't it be easier to implement the change like this?

Yea, that would probably have been better -- guess I overlooked it :)



Is there any ETA on the mempool overhaul?


In the mean time I'll close this PR and utilize the same stub functions for FLO until the refactor later removes it from all of them

martinboehm:
ETA for the mempool overhaul is about 6 weeks, if everything goes right :-)

PR #135, opened 2019-03-14
Add NULS supported.

kiss1987f4:
Add NULS supported.


martinboehm:
Hi, thank you for your contribution.
Could you please use the available ports xx53 instead of already used xx32? See the [registry of used ports](https://github.com/trezor/blockbook/blob/master/docs/ports.md).

kiss1987f4:
Thank you very much for your reply, I have changed the port to XX53.

configs/coins/nuls.json
```diff
@@ -0,0 +1,67 @@
+{
+  "coin": {
+    "name": "NULS",
+    "shortcut": "NULS",
+    "label": "NULS",
+    "alias": "nuls"
+  },
+  "ports": {
+    "backend_rpc": 6001,
+    "backend_message_queue": 38332,
```
martinboehm: Please change also the `backend_rpc` and `backend_message_queue` ports to 8053 and 38353 if possible. It seems to me from the code that NULLS does not use mempool but change the `backend_message_queue` anyway so that it matches other coins.

kiss1987f4:
As you said,NULLS does not use mempool, I will change the backend_message_queue post to 38353. But NULS's rpc port is 6001, I am afraid this is not very easy to modify.

martinboehm:
OK, leave the NULS's rpc port as it is 6001, if it is too complicated.

kiss1987f4:
NULS will support the boot parameter configuration port number in the next version, and change the backend_rpc port to 8035.

martinboehm:
I tried to build the nulls backend and Blockbook packages using the command `make all-nuls`, however the backend build failed with some java dependencies problems, like
```
dpkg-shlibdeps: error: cannot find library libavcodec-ffmpeg.so.56 needed by debian/backend-nuls/opt/coins/nodes/nuls/jre/lib/amd64/libavplugin-ffmpeg-56.so (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '')
dpkg-shlibdeps: error: cannot find library libavformat-ffmpeg.so.56 needed by debian/backend-nuls/opt/coins/nodes/nuls/jre/lib/amd64/libavplugin-ffmpeg-56.so (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '')
```
Is the build using our makefile working for you?

nuls-io:
I fixed this problem and modified the backend_rpc port to 8053.

nuls-io:
Ask, I want to extend a LockTime field in the /api/utxo/ interface. What should I do to make it reasonable?
[](https://github.com/kiss1987f4/blockbook/commit/970756305cdf11dceb4dc105118cd9c9e37d6b87)
Is it possible to create a branch of nuls directly like this?

martinboehm:
Hi, it does make sense to return `LockTime` in utxo interface. 

However, we are considering to add a separate utxo index, probably by adding list of utxos with amounts to the `addressBalance` column, as the balance changes when utxos are spent anyway. Your request is a strong additional point for this implementation, as we would add the `LockTime` field there as well.

It has the following advantages:
1) at this moment, for addresses with a lot of txs, getUtxo can be slow and resource intensive - a special index would make it an O(1) request
2) the total utxo index would be relatively small, as only currently unspent spendable outputs would be kept

Your solution is adding 8 bytes to each output of already large `txAddresses` column and the `LockTime` is necessary to be indexed only for utxos.

I have the following suggestion: if you need the `LockTime` in the utxo interface, keep your solution for now without merging with upstream and when we implement the utxo index, we will add there the `LockTime` field and use the same API. You will be then able to switch to our solution.

martinboehm:
Also, I have been able to successfully build and run your Nuls backend and blockbook implementation. I have 2 points:
1) the sync integration tests are missing
2) the explorer does not show raw transaction (shows null) and API tx-specific request is also not returning anything

kiss1987f4:
> You will be then able to switch to our solution

Thank you, I am looking forward to your solution.

bchain/coins/nuls/nulsrpc.go
```diff
@@ -0,0 +1,616 @@
...
+		Height:        uint32(blockHeader.Data.Height),
+		Confirmations: blockHeader.Data.ConfirmCount,
+		Size:          blockHeader.Data.Size,
+		Time:          blockHeader.Data.Time,
+	}
+	return header, nil
+}
+
+func (n *NulsRPC) GetBlock(hash string, height uint32) (*bchain.Block, error) {
+	if hash == "" && height < 0 {
```
martinboehm: You give me a chance to nitpick, uint32 is never less than zero :)

kiss1987f4: Yes, I am wrong and have been modified.

martinboehm:
It seems to me that the block time (and also mined time in tx) is not properly parsed, see the screenshot:
image

Also, as you are not connected to any notification from the backend about new blocks, you depend on Blockbook's fallback mechanism to synchronise blockchain - polling. However, this polling is by default done only once per 15 minutes. You can override it by using parameter `-resyncindexperiod=30000` to do it every 30 seconds (or set it to any other reasonable value).

kiss1987f4:
If there are other questions, please let me know, thank you.

martinboehm:
I have a question about the `LockTime` in utxo interface. 

I realized, that it would make sense for Bitcoin-type coins to return `LockTime` only for transactions in mempool. The `LockTime` specifies the earliest time the transaction can be mined. For already mined transactions, the `LockTime` is not interesting.

What exactly is your use case with regards to `LockTime`?

kiss1987f4:
> I have a question about the `LockTime` in utxo interface.
> 
> I realized, that it would make sense for Bitcoin-type coins to return `LockTime` only for transactions in mempool. The `LockTime` specifies the earliest time the transaction can be mined. For already mined transactions, the `LockTime` is not interesting.
> 
> What exactly is your use case with regards to `LockTime`?

Each output of NULS has a LockTime field that indicates at what point in time or height the output is available; what is more embarrassing is that the NULS input carries the amount and LockTime field, and verifies that the amount and LockTime are correct; so I Need to return LockTime in the utxo interface.

martinboehm:
Thanks for the explanation. I will have to think about how to support your requirement without affecting other coins.

kiss1987f4:
This is good news for me, thank you very much for your help.

PR #173, opened 2019-05-06
ZelCash support

TheTrunk:
Hello,

ZelCash is a fork of Zcash and so I have reused current Zcash implementation as all of it is valid for ZelCash.

The contribution guide is not very clear if such an implementation is correct or a new folder zel with importing zec configuration should be used with adjusting tests to real data of ZelCash chain (although Zcash mocked ones work on ZelCash). 
ZelCash is also not currently supported by Trezor (not listed here https://trezor.io/coins/ ) and we are hoping for bringing ZelCash to Trezor and other wallets with this PR.

I have setup a Blockbook server accessible here: https://blockbook.zel.cash/

Thank you


martinboehm:
If the ZCash implementation including the unit tests works for ZelCash, you do not have to reimplement it, of course.
However, please add integration tests, using blocks after the ZelCash fork.
Also, please use ports xx58.

PR #180, opened 2019-05-18
mod Fujicoin binary_url

fujicoin:
Implementing PR each time Core is upgraded is a waste of time for both parties. When we update Core, we will update the contents of fujicoin-blockbook.tar.gz.

### DIFF
--- a/configs/coins/fujicoin.json
+++ b/configs/coins/fujicoin.json
@@ -23,7 +23,7 @@
     "package_revision": "satoshilabs-1",
     "system_user": "fujicoin",
     "version": "0.16.3",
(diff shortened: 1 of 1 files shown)


martinboehm:
Hi, when I tried to build the package, I got:
```
dpkg-shlibdeps: error: cannot find library libX11-xcb.so.1 needed by debian/backend-fujicoin/opt/coins/nodes/fujicoin/fujicoin-qt (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '')
```

You could probably put fujicoin-qt to exclude list.

fujicoin:
I see. I deleted fujicoin-qt from fujicoin-blockbook.tar.gz and recreated SHA256SUMS.asc Please recheck.

fujicoin:
tnx!

PR #184, opened 2019-05-24
Add Ritocoin support

Scotty0448:
Public Blockbook Explorer:
https://blockbook.ritocoin.org

Coin information:
https://ritocoin.org
https://github.com/RitoProject/Ritocoin


martinboehm:
Hi, it almost seems OK. However I noticed some inconsistency in parsing/displaying `transfer_asset` addresses.

For example, https://blockbook.ritocoin.org/address/BRTraysiJ1qi6Y1nB1q44NcfM1eENptH63 includes tx `da73fa69d10050c9a1bc9024a0a4f3a379f3827d1371a27a8d38b7abd0` but does not show the address `BRTraysiJ1qi6Y1nB1q44NcfM1eENptH63` in outputs.

I would merge your PR as it is but maybe you would like to fix the issue first.

Scotty0448:
The address that you couldn't see in the outputs is one of the ones showing
as "Unparsed address".  The Ritocoin chain, just like the Ravencoin chain,
accommodates the transfer of "assets" that are separate from the base coin
(RITO). Asset transfers are done using a different script type and the
associated addresses are included in a slightly different way. The amount
of the base coin will always be 0 for these asset transfer scripts though.
Most wallets would likely want to ignore these for now until they perhaps
decide to support assets in some way.

I could look into what it will take to parse those addresses so that they
show up. I imagine it could mean abandoning the faster binary decoding
though. But curiously the search by address that you initially tried
resulted in finding the transaction anyways, even though the address was
unparsed. So perhaps there's something else I'm not aware of that's
happening.

martinboehm:
I understand that it is a nonstandard (from Bitcoin point of view) script with OP_RITO_ASSET. I just got surprised that the address (script) got indexed in the addresses index as a standard script yet it is not later parsed correctly. It is really up to you if you want to investigate the issue more.

Scotty0448:
I'd say that we should go ahead and merge what we have. It looks like
everything involving the base coin is working fine, and that's all that is
needed at this time. I'll then revisit this at a later time and see what I
can do to parse the asset scripts in a more useful manner.

martinboehm:
Sure, I have no problem merging it, I just wanted to point out the weird issue.

Issue #196, opened 2019-06-14
Fork detection in blockbook


artall64:
That issue is actually question not issue. 
I just wondering how blockbook handling temporary network forks. Let's say on bitcoin, how I can see that via block book API ?
1) Will I see new different block with the same `height` ?
2) Will I be notified via Websocket API about new block in that case ?
3) May be some special not document API exists ?
4) I guess I can test that functionality if I attach blockbook indexer to my own testnet where I can do network fork. But maybe there is simpler simulate that network fork scenario

Thank you in advance.

martinboehm:
Hi. Blockbook always follows the main chain of the backend it is attached to. If there is a rollback-reorg in the backend, Blockbook will also do rollback. There is no special API for that in Blockbook. If you ask for block by height, you will always get the main chain block. If you ask for block by hash, you may get the block from another fork but it is not guaranteed (backend may not keep it).

To your questions:
1) yes, you may get a different block on the same height if there has been a fork in between
2) you will be notified via Websocket API about new blocks. If there is reorg, you will get a new block hash with the same (or even smaller if the reorg is deeper) height
3) there is no special undocumented API
4) yes, you can do that. Or if you want to test it, you can run Blockbook with coins where reorgs are quite frequent (for example Ethereum Testnet Ropsten is fairly easy to run)

artall64:
Thank you so much for detailed answer. If you mind I would add that in some where in the docs and open pull request

PR #198, opened 2019-06-14
Add notes on API in case of fork temporary fork of blockchain

artall64:
That request refers to the following issue:
https://github.com/trezor/blockbook/issues/196


martinboehm:
Thanks.

PR #207, opened 2019-06-18
Optimize Docker build

MrMebelMan:
I've chained `RUN` commands to decrease the number of created layers, so the resulting images take less space.

I've also changed the number of compiler threads to have a dynamic value by using `nproc`, this should speed up the compilation a little.

The comments in subshells look hacky, but I didn't find any better way to comment chained code :(

### DIFF
--- a/build/docker/bin/Dockerfile
+++ b/build/docker/bin/Dockerfile
@@ -2,6 +2,7 @@
 
 FROM debian:9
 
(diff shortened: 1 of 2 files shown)


MrMebelMan:
Discussed this with @martinboehm, looks like it's an unnecessary optimization that doesn't provide much benefits and even slows down the build due to rocksdb layer caching. I am closing this PR :|

PR #211, opened 2019-06-24
Fix the --vendor-only flag reference

dmigwi:
The `--vendor-only` dep flag was incorrectly referenced.

Since dependencies related to a specific coin are not added to the repo `Gopkg.toml`, running `dep ensure` to fetch all the dependencies creates `.vendor-new/` folder thus the reason why `.vendor*` was added to the `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 vendor/
+.vendor*
 data*/
 profile/
 blockbook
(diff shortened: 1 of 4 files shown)


martinboehm:
Hi, what issue fixes this PR? I am not aware of any problems in our build procedure.

dmigwi:
This change is probably not necessary.

PR #212, opened 2019-06-24
Less verbose on docker deployment

dmigwi:
`-qq` is a two level quiet level modified that hides most of the progress information for the `apt-get`.
The modifier doesn't log anything unless its an error.

Ref: `man apt-get` on linux or https://serverfault.com/questions/644180/what-does-qq-argument-for-apt-get-mean

`-nv` is the `wget` no verbose flag or `--no-verbose` Turn off verbose without being completely quiet (use -q for that), which means that error messages and basic information still get printed.

Ref: `man wget` on linux or https://www.cleancss.com/explain-command/wget/206217

### DIFF
--- a/build/docker/bin/Dockerfile
+++ b/build/docker/bin/Dockerfile
@@ -2,9 +2,9 @@
 
 FROM debian:9
 
-RUN apt-get update && \
-    apt-get upgrade -y && \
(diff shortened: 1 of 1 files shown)


martinboehm:
Thank you for your effort but I do not see any problem with build verbosity. It is not done so often to fill the disk with log files and at least it shows what it is doing.

Issue #229, opened 2019-07-12
decred: systemd type wrong


dajohi:
The backend systemd file for Decred has `Type=forking`.  `dcrd` runs in the foreground, resulting in systemd timing out the process.

PR #230, opened 2019-07-12
decred fixes

dajohi:
- set service type to simple as dcrd runs in the foreground.
- fix test rpclisten port
- set mainnet to false on testnet
- update mainnet explorer url.


dajohi:
Closes #229

PR #239, opened 2019-07-22
Use proper public key id values for Decred

dajohi:


martinboehm:
I see that there is also a difference in "slip44": 3 versus "slip44": 42 in the https://github.com/trezor/trezor-common/blob/master/defs/bitcoin/decred.json. I regenerated the decred.json file (we have a tool which synchronizes Blockbook configs with trezor-common). I can commit it, if you want to.

dajohi:
thanks for pointing that out.  PR on its way...

PR #245, opened 2019-07-26
WIP: decred: set proper base58 hasher (blake256).

dajohi:


dajohi:
marking WIP.  Adding and fixing tests.

PR #249, opened 2019-08-04
Fix decred's xpub decoding

dmigwi:
This PR fixes the extend public Key decoding  functionality that was returning an error before.

@dajohi This PR is still essential even if https://github.com/martinboehm/btcutil/pull/3 was merged.
It could even override that PR since implementation unique to decred's checksum hash generation is handled here fully.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -145,7 +145,7 @@
   branch = "master"
   name = "github.com/martinboehm/btcutil"
   packages = [".","base58","bech32","chaincfg","hdkeychain","txscript"]
-  revision = "a7b7a55592485f5972902f1caf076a434b079e29"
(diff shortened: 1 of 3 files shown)


martinboehm:
How is this related to this PR: WIP: decred: set proper base58 hasher (blake256). #245? Could you please consolidate this to one PR?

dmigwi:
PR #245 is being fixed so that I can work like this PR without needing new decred imports in this PR, but that requires additional modifications to martinboehm/btcutil which are not yet ready.

martinboehm:
OK, please tell me when you are ready to merge the decred modifications and in which single PR as I don't have time to analyze what does each of your (decred) PR do.

dmigwi:
@martinboehm You can now merge this PR instead of #245

PR #250, opened 2019-08-06
Implement the "feestats" endpoint

MrMebelMan:
* Add a FeeStats struct (TxCount, AverageFeePerKb, TotalFeesSat, Deciles).
* Add a TxSpecific struct with additional Vsize and Size fields.
* Refactor GetBlockInfoFromBlockID into a function.
* Add a test case for "feestats".


bchain/types.go
```diff
@@ -89,6 +89,13 @@ type Tx struct {
 	CoinSpecificData interface{} `json:"-"`
 }
 
+// Extends Tx with an additional Size and Vsize info
```
martinboehm: Make this type local to GetFeeStats function. It is very specific :)

api/worker.go
```diff
@@ -995,6 +989,110 @@ func (w *Worker) GetBlock(bid string, page int, txsOnPage int) (*Block, error) {
...
+		index := int(math.Floor(0.5+float64(k)*float64(n+1)/10)) - 1
+		if index < 0 {
+			index = 0
+		} else if index >= n {
+			index = n - 1
+		}
+		deciles[k] = feesPerKb[index]
+	}
+
+	return &FeeStats{
```
martinboehm: Please log how long it took to process the block, ideally with block hash and txcount

api/worker.go
```diff
@@ -972,13 +972,7 @@ func (w *Worker) GetBlocks(page int, blocksOnPage int) (*Blocks, error) {
...
 }
 
-// GetBlock returns paged data about block
-func (w *Worker) GetBlock(bid string, page int, txsOnPage int) (*Block, error) {
-	start := time.Now()
-	page--
-	if page < 0 {
-		page = 0
-	}
+func (w *Worker) GetBlockInfoFromBlockID(bid string) (*bchain.BlockInfo, error) {
```
martinboehm: Make this method package private, i.e. rename it to getBlockInfoFromBlockID

api/types.go
```diff
@@ -198,6 +198,13 @@ type Tx struct {
 	EthereumSpecific *EthereumSpecific `json:"ethereumSpecific,omitempty"`
 }
 
+type FeeStats struct {
+	TxCount         int       `json:"txCount"`
+	TotalFeesSat    *Amount   `json:"totalFeesSat"`
+	AverageFeePerKb int64     `json:"averageFeePerKb"`
+	Deciles         [11]int64 `json:"deciles"`
```
martinboehm: Maybe rename to DecilesFeePerKb, so that it is more selfdescriptive.

PR #252, opened 2019-08-07
Check if unindexed confirmed tx height exists

dmigwi:
For unindexed confirmed transaction, unnecessary API call is being made to fetch the height from the back-end.
Since the block height could be readily available in most coins when unmarshaling tx data, avoid making that API call to the backed.

### DIFF
--- a/bchain/types.go
+++ b/bchain/types.go
@@ -76,12 +76,13 @@ type Vout struct {
 // Tx is blockchain transaction
 // unnecessary fields are commented out to avoid overhead
 type Tx struct {
-	Hex      string `json:"hex"`
-	Txid     string `json:"txid"`
-	Version  int32  `json:"version"`
-	LockTime uint32 `json:"locktime"`
-	Vin      []Vin  `json:"vin"`
-	Vout     []Vout `json:"vout"`
+	Hex         string `json:"hex"`
+	Txid        string `json:"txid"`
+	Version     int32  `json:"version"`
+	LockTime    uint32 `json:"locktime"`
+	Vin         []Vin  `json:"vin"`
+	Vout        []Vout `json:"vout"`
+	BlockHeight uint32 `json:"blockHeight,omitempty"`
 	// BlockHash     string `json:"blockhash,omitempty"`
 	Confirmations    uint32      `json:"confirmations,omitempty"`
 	Time             int64       `json:"time,omitempty"`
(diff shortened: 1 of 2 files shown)


martinboehm:
But actually this is not how most coins work. Most coins fetch only unparsed tx blob from the backend and parse it on the Blockbook side. Unfortunately, this tx blob does not contain block height, therefore the necessity of additional call. But even with this call (which is in many cases not necessary) it is much faster.

I will not merge this PR as it could affect functionality of other coins. If you need to know the height, please make the additional call in your (decred) code.

dmigwi:
I wanted to add some flexibility on how to acquire the height used to index transactions with the code below.

```Golang
        // the transaction may not yet be indexed, in that case:
	if tx.BlockHeight > 0 {
		// Check if the tx height value is set.
		h = tx.BlockHeight
	} else {
		// Get the height from the backend's bestblock.
		h, err = c.chain.GetBestBlockHeight()
		if err != nil {
			return nil, 0, err
		}
	}
```
Is there any other way to set the accurate index height if the wrong value is set? Also if there is a way to stop

martinboehm:
Where do you want to place the code above? To the TxCache.GetTransaction?

dmigwi:
Yes. TxCache.GetTransaction is where the transactions indexing height is being set.

martinboehm:
The optimization you suggest is actually minimal, as the height is normally read from the DB using the call `c.db.GetTxAddresses(txid)`. The branch of code you want to modify is reached only in case when a user requests information about tx before Blockbook manages to index the block which contains the tx (which is usually less than a second after it is published by the backend).

dmigwi:
>The branch of code you want to modify is reached only in case when a user requests information about tx before Blockbook manages to index the block...

This is what I am hoping to achieve to because our mempool implementation has confirmed transactions due to the decred's PoS voting.


This is the reasoning as to why confirmed tx are present in our mempool implementation is:
- A decred block is considered to be valid if 3 of the 5 votes approve it.
- Blocks with less than 2 confirmations are not yet approved or disapproved. This implies that the decred chain's best block is always not voted for. This block yet to be voted for has its txs appearing in the mempool till its voted in or out.
- So in our implementation we only allow blocks which have been approved to be synced. By this we avoid txs in invalidated blocks being incorrectly identified as double spents.
- With this minor modification, we can correctly assign the indexing height to the confirmed but not yet voted for block in the mempool.

martinboehm:
I understand. However, I have to analyze/test if you change does not have some unintended side effects.

PR #253, opened 2019-08-07
Check unindexed confirmed tx height for decred

dmigwi:
This PR depends on #252 

- Decred's implementation mempool data shows confirmed transactions that are not yet indexed because of the decred's proof of stake implementation. (in blockbook, decred's tx with less than 2 confirmations are listed in the mempool)


dmigwi:
@martinboehm This is different from the xpub decoding problem mentioned earlier.

Since the blockbook API does not provide access to the back-end database so that transactions in blocks invalidated by PoS votes in the consecutive block can be dropped, we made all transactions in the mempool to be those with less than 2 confirmations.

It so happens that currently every confirmed transaction in the mempool is assumed to be mined in the bestblock which is not always accurate in our case.

martinboehm:
@dmigwi Hi, the PR #252 is accepted. What is the status of this PR?

dmigwi:
@martinboehm This ready for merging now.

PR #255, opened 2019-08-19
Fixes duplicated entry in `docs/rocksdb.md`

kr-mohak:
Under the description of the `addressBalance` column, in the list of data in the value field, `block_height` was duplicated.


martinboehm:
Thanks, I would never have thought that somebody would go so deep in the documentation. :)

PR #258, opened 2019-08-19
Disable wallet functionality for bitcoin-like daemons

vlddm:


martinboehm:
Sorry, we will not merge this PR. It could have unknown consequences to some of the coins that use this config and we do not want to retest all.

PR #260, opened 2019-08-19
Use fast sync over full for ethereum backend

vlddm:


martinboehm:
Hi, thank you for your effort. However, changes of base packages and/or coins which we (Trezor) maintain go through our internal testing and we generally do not accept PRs that affect them. See our [contribution guide](https://github.com/trezor/blockbook/blob/master/CONTRIBUTING.md). We will have to decide with our QA and admins if we accept your existing PRs. Please in the future, open an issue in such cases.

PR #276, opened 2019-08-30
Add Unobtanium

Flurbos:
Hopefully every things is OK.

Hosting a temporary version @ : blockbook.unobtanium.uno

Thanks in advanced :+1:


martinboehm:
Hi, I successfully built the backend package and run it. However, the backend synchronization seems to stop at block 542498 0000000000001312bc2678062b33a193af82b719ab899b29f006133169a42f97 and does not sync further. Should I just give it time or do something else? I run the backend for at least 3 days.

Flurbos:
> Hi, I successfully built the backend package and run it. However, the backend synchronization seems to stop at block 542498 0000000000001312bc2678062b33a193af82b719ab899b29f006133169a42f97 and does not sync further. Should I just give it time or do something else? I run the backend for at least 3 days.

There shouldn't be any reason for it to stop syncing at that block, do you still have some connections?

There is a node running at:
165.22.242.31
its fully synced and also another at blockbook.flurbo.xyz

Also is there any noticeable output in debug.log?

martinboehm:
There are a lot of messages like this but it does not progress anywhere:

```
2019-09-09 10:35:26 receive version message: /Unobtanium:0.10.1.1/: version 70002, blocks=1351662, us=88.208.115.69:44756, peer=847                                                                                                                                           
2019-09-09 10:35:31 socket recv error Connection reset by peer (104)                                                                                                                                                                                                          
2019-09-09 10:35:46 connect() to 162.243.155.228:65534 failed after select(): Connection refused (111)                                                                                                                                                                 ```

**Comment by Flurbos** [comment]
Try adding any of these (or all) using ```addnode IPHERE add```
```134.209.109.103
167.71.59.99
195.182.58.120
212.125.247.42
24.15.242.74
46.234.83.216
```
Or add them to the .conf ie.
```addnode=134.209.109.103```

Although generally shouldnt be required (i didnt do) 
perhaps itll help you pass that block

**Comment by martinboehm** [comment]
Thank you for the suggestion. However I would prefer to have a functional package, which works any manual configuration. 
Also, I looked at your configs/coins/unobtanium.json and noticed that you specify "reindex": 1. Doesn't it reindex the backend database after every restart of the service? At least it seemed to me that it is what it did when I restarted it.

**Comment by Flurbos** [comment]
> Thank you for the suggestion. However I would prefer to have a functional package, which works any manual configuration.
> Also, I looked at your configs/coins/unobtanium.json and noticed that you specify "reindex": 1. Doesn't it reindex the backend database after every restart of the service? At least it seemed to me that it is what it did when I restarted it.

Thats fair, i had 0 issues building 3 blockbook servers for uno, so still not sure why yours got stuck.

Our daemon requires a reindex on the first time running, so that config was added just for that purpose.

**Comment by martinboehm** [comment]
I restarted the backend service. The last log entries are:
```
2019-09-09 22:18:10 Before: 1a19056f 00000000000019056f0000000000000000000000000000000000000000000000
2019-09-09 22:18:10 After: 1a19cdf1 00000000000019cdf10df1ef86b781e5e37aab75d9d76e9f69cdcb62935dc1c0
2019-09-09 22:18:10 UpdateTip: new best=0000000000001312bc2678062b33a193af82b719ab899b29f006133169a42f97  height=542498  log2_work=68.988555  tx=667440  date=2015-01-07 11:25:04 progress=0.029371  cache=94796
2019-09-09 22:18:10 Loaded 230700 blocks from external file in 102056ms
2019-09-09 22:18:10 Reindexing finished
```
After that there is no progress, no additional line in the log.

**Comment by Flurbos** [comment]
It is unusual for this to happen, perhaps delete unontaniums peers.dat file before reindexing because it seems you may have picked up a bad peer i can only assume, and trying again with that same peers file may reproduce same results.

I will also try start up a new blockbook server and ensure it works out of the box for me but might not be until after tomorrow that i can get to do this.

**Comment by martinboehm** [comment]
I managed to synchronize the backend (after several attempts) and then Blockbook. I see two isseues:

1) it seems that notifications about new transactions/blocks are not working. You can reduce the polling interval for them by setting the following parameters in your config json.

https://github.com/trezor/blockbook/blob/35f32787718d99582e30a2c28a7e6759dbd0f24f/configs/coins/viacoin.json#L55

2) the integration fails:
```
--- FAIL: TestIntegration/unobtanium=main/rpc/GetBlockHeader (0.00s)
  rpc.go:358: hash b06166e7ec4da9759838415fe8d24a52d069e58731f1b926fce694f8a0b87229: -32601: Method not found
```

Flurbos:
I was unaware it failed that test, i removed it from tests.json and added params to reduce polling.
Hoping this resolves the issues you mentioned.

PR #338, opened 2019-12-19
Fiat rates & Balance history update

MrMebelMan:
* Input & output date values as Unix timestamps
* Store rates as `float64` instead of `json.Number` in the database
* If currency is not specified, return all available rates
* Document the new APIs

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -648,7 +648,7 @@ Example error response (e.g. rate unavailable, incorrect currency...):
 Returns a balance history for the specified XPUB or address.
 
 ```
-GET /api/v2/balancehistory/<XPUB | address>?from=<dateFrom>&to=<dateTo>[&fiatcurrency=<currency>&gap=<gap>&groupBy=<groupBySeconds>]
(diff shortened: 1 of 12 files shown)


docs/api.md
```diff
@@ -575,6 +578,109 @@ or in case of error
...
+GET /api/v2/balancehistory/<XPUB | address>?from=<dateFrom>&to=<dateTo>[&fiatcurrency=<currency>&gap=<gap>&groupBy=<groupBySeconds>]
+```
+
+Query parameters:
+- *from*: specifies a start date, format is YYYY-MM-DD.
+- *to*: specifies an end date, same format. 
+
+The optional query parameters:
+- *fiatcurrency*: if specified, the response will contain calculated fiat amounts at the time of transaction.
+- *gap*: TODO
```
MrMebelMan: @martinboehm, can you please help me documenting this parameter? :pray:

MrMebelMan: resolved IRL

Issue #341, opened 2019-12-30
Go Modules Support


dmigwi:
Decred (DCR) would like to upgrade their binary dependency from v1.4.0 to v1.5.0. We currently can't be able to successfully make that upgrade because some features of the go modules semantic versioning don't work very well with dep dependency management tool.

@martinboehm, a while back you gave this [response](https://github.com/trezor/blockbook/issues/205#issuecomment-503016557).  By now its clear that Go Modules do have a wide support since it simplifies a lot of the things that are currently done to set up and run a project. 

Has the issue priority changed since then? If not would you kindly consider adding some more priority to the issue? 

Go modules support will be very helpful to decred and many other supported coins.

martinboehm:
Yes, it is probably time to switch to Go Modules. I will let you know here when it is done.

dmigwi:
Hi @martinboehm, If no one is working on this task now, I could submit a PR with the needed fix for you to validate and merge.

martinboehm:
Great, go ahead!

martinboehm:
Blockbook now uses go modules.

PR #352, opened 2020-01-20
WIP: Add gomodules support to all packages

dmigwi:
This is in relation to #341

I thought it wise to make every coin implementation a sub-module to keep coin specific dependencies only limited to the specific coin implementation.


martinboehm:
Thanks a lot! The submodule for each coin looks like a good idea. However, would not it make it hard to upgrade a library used in all coins to a newer version? But probably, when thinking about it, the global text replace will work fine...

BTW when I try to compile the PR using go build, I get error:

dmigwi:
I have been trying to resolve that error for quite a while. The error is due to go-ethereum packages that don't support gomodules which is why the PR has a *WIP* in its title.. I think the global text replace could help fix that error.

martinboehm:
Hi, please, do not mix unrelated changes to this PR. The commits 0226a6f and 22ecf29 are not within the scope of this PR.
Actually, what is the reason for 0226a6f?
Also, why do you rebuild Blockbook when generating the config files in the contrib/scripts/build-blockchaincfg.sh? This is something I would not like to do.

dmigwi:
Closing this in favour of #374

PR #382, opened 2020-03-04
General Token support for Blockbook, integrate Syscoin SPT as first use-case

sidhujag:
1) Token support across all current major API's (xpub,balancehistory,asset,websocket,socketio)
2) Not restrictive to ERC20, create first implementation using SPT but not restrictive, aslong as tokens can be indexed via an int32 GUID it should work with only a little bit of work parsing each token platform asset individually
3) This PR depends on https://github.com/martinboehm/btcd/pull/1 which does serialization of SPT's in BTCD. Once that is merged-in, I can edit the code which points to syscoin/btcd to point to your btcd for SPT parsing.
4) General design constraint to ensure all API compatibility with existing coins

Who we are:

Blockchain Foundry is a blockchain development firm listed and traded on the CSE (BCFN:CSE). We are involved in many initiatives in the blockchain space including Digital Identity (part of the Decentralized Identity Foundation), as well as scalability (onchain / offchain crypto-crypto transactions) and interoperability (cross-chain consensus). Within each of these we have implemented solutions or are implementing solutions for enterprise clients. We have been around since 2013 and have developed a protocol (Syscoin) that puts these efforts out to the market:

1) Security: Syscoin is compliant with Bitcoin 100%. Merge-mined where BTC miners each Syscoin whilst securing Syscoin earning block subsidy. We built an opt-in asset protocol for users wishing to issue tokens on our chain. However base Syscoin remains under the same security domain as Bitcoin, while our Asset protocol has custom consensus, still adhering to Bitcoin standards but just with extensions for creating new coins through coloured coin transactions.
2) Performance: ZDAG protocol for on-chain point-of-sale crypto-to-crypto transfer capability. Opt-in and a feature of our Asset protocol (Syscoin Platform Tokens). https://syscoin.org/zdag_syscoin_whitepaper.pdf. Pretty much a formalization of Satoshi's snack-machine example: https://bitcointalk.org/index.php?topic=423.20. Together with TU-Delft we are researching ways to create payment channels (Lightning Networks) with our Asset protocol as a POC for assets + LN for offchain payments, supplemented obviously with our on-chain capability to do faster payments provides a unique experience to end-users.
3) Interoperability: Another opt-in feature for our Asset protocol. Sysethereum-bridge (https://github.com/syscoin/sysethereum-docs)  Syscoin essentially is at a cross between Bitcoin and Ethereum. Assets on Syscoin can interoperate with Ethereum ERC20 standard. Tokens burn and mint in either direction without middle-men or custodians. Ethereum contracts store Syscoin block headers for Syscoin SPV proofs, and Syscoin runs Geth light node to verify Ethereum SPV proofs.  Gas consumption remains constant around 100k or less for the good case on supplying optimized merkle root of an hour of headers to Ethereum whilst during challenges, it costs as little as 3M gas to prove the hour of blocks is valid for an honest submitter of blocks to Ethereum. This in contrast to BTC Relay and other relay contracts from coins to Ethereum that cost more than 30-40x this much for the equivalent security, on top of being only a one-way design, ours is two-way with ability to start on either side and move to the other side without dependency. In the future we are researching ways for zero-knowledge proofs to replace the SPV proof mechanism for general purpose cross-chain consensus.

We believe Blockbook is an integral part of the design for enabling our ecosystem for our clients, consumers and community. We will be developing Web interfaces into Trezor HW wallets (for eth/btc/sys) as well using web-usb and generally feel Blockbook will serve as an infrastructure to our future products and offerings to clients and community. I would love to work with you on improving Blockbook over the long-term as needed.

I am the CTO of Blockchain Foundry as well as core developer of Syscoin.

The PR recognizes a general purpose Token integration on top of the Bitcoin-type chain. This was my first attempt at using Go so there may be language related issues however the tests pass and we are currently integrated into our wallet without any issues so far. The work was done to be as generic as possible and also we have added ability for tokens to be fetched as part of balance history as well as xpub token filtering. Overall your code was fairly well-written and testable so integration was mostly straight forward. Parts of the code were re-organized so our general purpose approach would work, by moving some functionality to base classes/parsers. All Ethereum/Bitcoin API's should have remained constant whilst still providing functionality Syscoin needed from both API level as well as explorer public interfaces.


martinboehm:
Hi. You have done an impressive amount of work.
Unfortunately, the code you have created significantly diverges from our code, therefore we cannot merge your pull request.
I would recommend you to keep a separate fork, in which you can implement the changes required for the coins you want to support.

sidhujag:
Alot of it is refactoring your base to make it more compliant for a general purpose token integration. I made sure it was compliant with the existing code so there are no code paths that are different from original. However I understand and it is unfortunate, thanks for the read.

PR #392, opened 2020-03-14
Remove empty map on address unsubscribe

fanatid:
It is possible subscribe to a lot of not exists addresses and then unsubscribe (or disconnect). If repeat this in loop, `addressSubscriptions` will have a lot of empty maps what should cause OOM in the end.
(Not sure about this, never write code on go for real apps)


martinboehm:
Good catch, thank you!

Issue #397, opened 2020-03-27
Batched retrieve of pending (mempool) transactions


catenocrypt:
Blockbook keeps track of pending (unconfirmed, mempool) transactions, and it returns pending transactions as well when transactions of an address are retrieved.  However, pending transactions cannot be retrieved in a batch, independently of any address.

*Motivation:* In usage scenarios where a high number of addresses are followed, it is not efficient to poll for the transactions of each address.  Instead, it would be useful to be able to retrieve the set of pending transactions (similar how transactions of one block are returned), and then they can be filtered by addresses.

*Proposal:* an API to return current pending transaction, filtered by a timestamp range.
E.g.   `GET /api/v1/mempool/<start-time>`

Use case with periodical retrieval of blocks and pending transactions:
1. Retrieve blockbook status
1. If there is no new block (since last), and no new mempool tx (since last), sleep for some time, start over (Note: current `/api/` returns latest mempool tx timestamp `lastMempoolTime`).
1. If there is new block (or more), retrieve new block(s), process transactions, start over
1. If there is new block, but there are new pending transactions, retrieve new pending transactions (using new API), process them, start over.

martinboehm:
While it would be possible to create such API, I would suggest to use the websocket interface, where you can subscribe to be notified about new blocks and mempool transactions affecting selected addressed.

catenocrypt:
Is it possible to subscribe using websocket for any transactions, or only for those related to a particular address?  Because if it is only per-address, it is not practical to have 10K+ websocket connections for the 10K+ addresses to be watched.

martinboehm:
You do not need 10K+ websocket connections. Create exactly one connection and observe 10K+ addresses. The websocket api `subscribeAddresses` accepts multiple addresses as comma separated value list. 
Blockbook keeps one subscription per connection, it sends notifications about addresses subscribed in the last call to `subscribeAddresses.

vikmeup:
@martinboehm or even better just for subscribe to all incoming mempool transactions and then the client will filter out transactions/addresses he does not need. Having options for both would be great. 

Current use-case we explore on @trustwallet to notify users for incoming transfers from mempool, observing for millions of addresses might be tricky, for option for all addresses might work better in this case.

catenocrypt:
One connection watching 10K+ connections is fine, but at this number the updating of watched address list is in itself an issue (new addresses coming very often).

catenocrypt:
[...]

PR #417, opened 2020-05-07
Proposal for mempool API (issue #397)

catenocrypt:
Proposed implementation for mempool API, for retrieving mempool transactions, in one go, independent of affected addresses, filtered by time.  This allows periodic retrieval of new mempool transactions, to be able to notify users, in the usage case when a very large number of addresses are being watched.  Address-based watching is not feasible in this case, but full blocks are retrieved.  See issue #397 .


catenocrypt:
This solution is discouraged by dev team, as being not efficient, using too many calls to backend.  See discussion in #397 .  New websocket-based approach is in PR #420 .  Closing this.

PR #420, opened 2020-05-12
New subscribeAnyAddress() method on websocket (issue #397)

catenocrypt:
New subscribeAnyAddress() method on websocket.  With this it is possible to subscribe to any address.  A TX notification is sent to clients subscribed to that address, and/or to clients subscribed to any address.

This allows prompt notification of new mempool transactions for clients who need to observe a high number of addresses, whereby address-based watching is not feasible.

See issue #397 .

### DIFF
--- a/server/public_test.go
+++ b/server/public_test.go
@@ -1421,6 +1421,20 @@ func websocketTestsBitcoinType(t *testing.T, ts *httptest.Server) {
 			},
 			want: `{"id":"36","data":[{"time":1521514800,"txs":1,"received":"1","sent":"0","rates":{"eur":1301,"usd":2001}}]}`,
 		},
+		{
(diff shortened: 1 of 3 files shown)


martinboehm:
Hi, thank you for your PR.
I have reviewed it and it must be modified a bit. As it is implemented, it would send the notification of the same tx multiple times, once for each output address.

I will close this PR but I will do the following modifications based on your commit:
1) a new internal event onNewTx, fired only once per new mempool transaction (not once for each address in the new mempool transaction)
2) I will make this feature optional, controlled by command line parameter. It can lead to substantial network traffic if there are many subscriptions. We do not need this feature to be available in our production servers.
3) I will rename the subscription to subscribeMempool

PR #426, opened 2020-05-15
Add fiat rate(monacoin)

wakiyamap:
This is a pull request for fiat rates of monacoin.
I would appreciate it if you could let me know if you have any problems.

Postscript.
Thank you so much as always. Use it to pay for your beer!
https://blockstream.info/tx/864e14ccbddf100dad4faeac9a44e6d7ee6e7a812cd85a151ad18da1d97024da


martinboehm:
Thank you, this is very kind of you!

Issue #424, opened 2020-05-14
addToAddressesMap is too slow


xiaoyuer97:
I have a blockbook node syncing bsv is extremely slow recently, so I get the profile:

```
File: blockbook
Build ID: b7c75f24f454bc9a5c80e49afd0e4845f73342ac
Type: cpu
Time: May 14, 2020 at 6:27pm (CST)
Duration: 30.18s, Total samples = 30.30s (100.39%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 30.07s, 99.24% of 30.30s total
Dropped 47 nodes (cum <= 0.15s)
Showing top 10 nodes out of 24
      flat  flat%   sum%        cum   cum%
    24.95s 82.34% 82.34%     24.95s 82.34%  memeqbody
     3.44s 11.35% 93.70%     29.91s 98.71%  github.com/trezor/blockbook/db.addToAddressesMap
     0.77s  2.54% 96.24%      0.77s  2.54%  runtime.memequal
     0.75s  2.48% 98.71%     26.46s 87.33%  bytes.Equal (inline)
     0.16s  0.53% 99.24%      0.17s  0.56%  syscall.Syscall6
         0     0% 99.24%      0.24s  0.79%  github.com/trezor/blockbook/api.(*Worker).GetSystemInfo
         0     0% 99.24%     29.92s 98.75%  github.com/trezor/blockbook/db.(*BulkConnect).ConnectBlock
         0     0% 99.24%     29.92s 98.75%  github.com/trezor/blockbook/db.(*BulkConnect).connectBlockBitcoinType
         0     0% 99.24%      0.23s  0.76%  github.com/trezor/blockbook/db.(*RocksDB).DatabaseSizeOnDisk
         0     0% 99.24%     29.92s 98.75%  github.com/trezor/blockbook/db.(*RocksDB).processAddressesBitcoinType
(pprof) peek memeqbody
Showing nodes accounting for 34.13s, 100% of 34.13s total
----------------------------------------------------------+-------------
      flat  flat%   sum%        cum   cum%   calls calls% + context
----------------------------------------------------------+-------------
                                            24.96s   100% |   bytes.Equal
    24.96s 73.13% 73.13%     24.96s 73.13%                | memeqbody
----------------------------------------------------------+-------------
```
[...]

PR #427, opened 2020-05-16
[R4R] reduce time complexity of addToAddressesMap from o(n^2) to o(n)

yutianwu:
Resolves https://github.com/trezor/blockbook/issues/424

### DIFF
--- a/db/rocksdb.go
+++ b/db/rocksdb.go
@@ -501,6 +501,7 @@ type txIndexes struct {
 // each address contains a slice of transactions with indexes where the address appears
 // slice is used instead of map so that order is defined and also search in case of few items
 type addressesMap map[string][]txIndexes
+type addressTxIdxMap map[string]map[string]int
 
 type outpoint struct {
 	btxID []byte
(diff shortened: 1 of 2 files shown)


martinboehm:
Thank you for your help. 

Unfortunately the situation is not as simple.
You are replacing O(n^2) algorithm by approximately O(n*log(n)) algorithm, with much higher constant factor and also higher memory usage. 

The map algorithm will greatly help in situations where there are few addresses, with a lot of transactions each (as in BSV blockchain) but will perform worse in situations with a lot of addresses with few transactions each (for example BTC blockchain).

We have encountered this situation already in utxo logic. After a lot of tests, we implemented a hybrid algorithm, which uses a simple loop until there are `16` and more utxos for the address, after which it switches to map logic, see the code starting here https://github.com/trezor/blockbook/blob/b957ed66abc75bbef2dce3db085fb97b7417511f/db/rocksdb.go#L565 The number `16` was obtained by a number of tests as a sweet spot when the linear search alogrithm starts to get slower than the map algorithm.

I have actually run the initial sync of Bitcoin Testnet network with and without your PR. The inital import with the map algorithm is about 20% slower, all the other factors the same. From our experience, Bitcoin mainnet would suffer even higher slowdown.

But it is good that you pointed out the problematic place, I will run couple of tests and see if the hybrid algorithm makes sense, properly tuned could provide speedup to all coins.

For now, I would recommend to run BSV with your map algorithm but I will not merge your PR as is.

yutianwu:
Thanks for your tests, I did not test on other situations, but it did work for bsv.  maybe we can find a way to solve the problem for both cases finally.

PR #431, opened 2020-05-16
eth-like backends listen on localhost instead *

vlddm:
binding rpc to 0.0.0.0 is very insecure and there is no reason to do so at all. 
Bitcoind rpc listening on localhost works fine.


martinboehm:
Thank you very much! You are right, this is a historical leftover which is not necessary any more.

PR #441, opened 2020-06-17
You have to link against libdl on Linux

CryptoManiac:
Otherwise there will be linking error because rocksdb is importing ```dlclose```, ```dlopen``` and other functions from libdl.so.


martinboehm:
What distro/version is causing the problem as we do not have it in our builds?

CryptoManiac:
@martinboehm In Ubuntu 20.04, for example, there is simply no way to link the package without libdl. 

However, it's possible that in your environment some other library is being linked with libdl thereby effectively suppressing the linking error.

PR #455, opened 2020-07-10
Cryto miner2020 patch 1

CrytoMiner2020:


martinboehm:
Sorry, we are not using github workflows. Keep them in your downstream repo, please.

PR #507, opened 2020-11-05
Bump DeepOnion Version to v2.2

nezero:


nezero:
https://trezor.deeponion.org/

Backend up and running, still synching.

PR #528, opened 2020-12-03
Updating the API docs so it shows more information about the web socket communication

martin-key:
I needed some time to understand how does the websocket work for subscribing to an address, so I decided to add it to the documentation so it is clearer to other people using the service.


martinboehm:
Thank you!

PR #529, opened 2020-12-04
chore(static): show RBF in the transaction detail

### LINKED ISSUE #525: Replace RBF exclamation mark with something less dangerous
This is very nit-picky but I think the "danger" symbol that appears next to the transaction ID for RBF transaction is too "scary". I would like to replace it with smth less warning. I will discuss with @matejzak first.

image

tsusanka:
Closes #525. I did it "blindly", it builds but I did not see the result. So please check :).

### DIFF
--- a/static/templates/tx.html
+++ b/static/templates/tx.html
@@ -65,6 +65,18 @@
                 <td>Fees</td>
                 <td class="data">{{formatAmount $tx.FeesSat}} {{$cs}}</td>
             </tr>{{end -}}
+            {{- if $tx.Rbf}}
+            <tr>
+                <td><span title="Replace by fee">RBF</span></td>
+                <td class="data">
+                    {{- if $tx.Rbf}}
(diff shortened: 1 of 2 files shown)


martinboehm:
Thank you for the contribution. 

There is a reason why the RBF flag was shown in `txdetail.html`. The `txdetail.html` is included not only in the view of a single TX, but also in the Address and XPub views. Your implementation removes this functionality.

In addition to that the OFF️ part of the branch in the `tx.html` will never be executed as it is inside `if $tx.Rbf`. The correct condition would be probably `if not $tx.Confirmations` to show the RBF status only for mempool transactions and then the inner condition will work as intended.

But I do like you idea of showing the text RBF instead of an icon. So I suggest to keep the functionality in `txdetail.html` but replace the icon by the text RBF, with the explanatory tooltip.

We can of course keep the RBF info in the `tx.html` too after it is corrected.

PR #550, opened 2021-01-03
feat: Support Ethereum Goerli testnet

syrohei:
Hi guys, Just added a commit to support Goerli testnet.


martinboehm:
Hi. Thank you! Could you please move the ports to 1xx26 so that we can run Ropsten and Goerli at the same time?

wakiyamap:
@syrohei Hi. I think probably 1xx26 is talked it.
https://github.com/trezor/blockbook/blob/master/docs/ports.md

syrohei:
Hi @martinboehm , I pushed some commits. Could you review it thanks.

Issue #568, opened 2021-02-23
Ubuntu 20.04 - timezone choosing hanging


Perlover:
OS: **Linux Mint 20** (**Ubuntu 20.04** base)
commit: db597c1f660f857caf303808e788fefc09d28af9
If I do `make BASE_IMAGE=ubuntu:20.04 all-bitcoin` I got a hanging of scripts at timezone choosing:

```
Setting up libsasl2-modules-db:amd64 (2.1.27+dfsg-2) ...
Setting up tzdata (2021a-0ubuntu0.20.04) ...
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
Configuring tzdata
------------------

Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.

  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area: ^Cmake[1]: *** [Makefile:47: .bin-image] Прерывание
make: *** [Makefile:44: build-images] Прерывание
```
I try to enter string `8` and `Enter` and `Europe` and `Enter` but nothing happens.
I think there is some utility `dialog` problem which missings in Docker scripts.
I googled like same problem here: https://github.com/moby/moby/issues/27988#issuecomment-357243865

Perlover:
My problem was resolved by this hint:

https://github.com/trezor/blockbook/issues/45#issuecomment-666429319

I added the line `ARG DEBIAN_FRONTEND=noninteractive` to build/docker/bin/Dockerfile after `FROM $BASE_IMAGE`:

```
# initialize from the image defined by BASE_IMAGE
ARG BASE_IMAGE
FROM $BASE_IMAGE
ARG DEBIAN_FRONTEND=noninteractive
```

After the `make BASE_IMAGE=ubuntu:20.04 all-bitcoin` works here

PR #569, opened 2021-02-23
Ubuntu 20.04 doesn't work without this patch

Perlover:
The problem was described here:

https://github.com/trezor/blockbook/issues/568


martinboehm:
Thank you!

PR #574, opened 2021-03-09
Websocket new transaction

kaladinlight:
Add functionality to subscribe to all new pending transactions. This will allow the flexibility to either filter by address as already existed or receive all messages published to mempool depending on the use case.

Documentation and tests have been updated to reflect the new subscription.

The other option I thought about was passing a `*` wildcard string to the `subscribeNewAddresses` endpoint which would inform the filter to return all transactions, but adding a new endpoint to handle this case separately made more sense to me and felt cleaner.

Let me know your thoughts on this feature update. Thank you!

### DIFF
--- a/blockbook.go
+++ b/blockbook.go
@@ -73,6 +73,8 @@ var (
 
 	noTxCache = flag.Bool("notxcache", false, "disable tx cache")
 
+	enableSubNewTx = flag.Bool("enablesubnewtx", false, "enable support for subscribing to all new transactions")
+
 	computeColumnStats  = flag.Bool("computedbstats", false, "compute column stats and exit")
 	computeFeeStatsFlag = flag.Bool("computefeestats", false, "compute fee stats for blocks in blockheight-blockuntil range and exit")
 	dbStatsPeriodHours  = flag.Int("dbstatsperiod", 24, "period of db stats collection in hours, 0 disables stats collection")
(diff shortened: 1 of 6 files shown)


martinboehm:
Hi. Thank you. Although I haven't checked it yet in detail, I am not against it. 
However, we certainly do not want this functionality to be enabled on our production servers. 
Would you be able to make it disabled by default and enabled only by some command line option?

kaladinlight:
I can add a flag to disable this by default. Is there a specific reason for this? 

I just tapped into the already existing `OnNewTx` function so there will only ever be the single mempool transaction fetched and used for any subscriptions (NewTransaction or Addresses). The only added work would be the actual publishing of each mempool transaction to any subscribed clients.

My thought is that the websocket channels _should_ be able to handle this work, but I could be mistaken. Would the concern be that we are trying to publish too many messages and we end up in a channel overflow situation?

kaladinlight:
@martinboehm I have added the changes requested.

I chose to leave the endpoints, but only performing the actual (un)subscribe if the feature is enabled. There is an additional message sent back if either endpoint fails to inform of the flag to use.

I left the endpoints in the `test-websocket.html` as well just so they exist for further documentation, but this will obviously not successfully subscribe in your hosted production server assuming the default enable state.

martinboehm:
Thank you!

PR #583, opened 2021-03-19
OP_RETURN: upgrade to utf8

fujicoin:
It upgrades the display of OP_RETURN from ascii to utf8.

### DIFF
--- a/bchain/coins/btc/bitcoinparser.go
+++ b/bchain/coins/btc/bitcoinparser.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/hex"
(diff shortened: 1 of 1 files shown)


bchain/coins/btc/bitcoinparser.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/hex"
+        "unicode/utf8"
```
wakiyamap, line 7: I recommend deleting two spaces.

fujicoin: Changed spaces to TAB.

PR #605, opened 2021-04-23
Fixing the name of the Bitcoin Gold Testnet

martin-key:
The name should not have `-` symbol separating the words Bgold and Testnet


martin-key:
@martinboehm would you be able to check that PR?

martinboehm:
Thank you!

Issue #665, opened 2021-08-30
Build fails due to apt-get update timeout on debian 9


dylanseago:
```
> make deb-bitcoin
Building image blockbook-build from debian:9
Sending build context to Docker daemon  6.144kB
Step 1/26 : ARG BASE_IMAGE
Step 2/26 : FROM $BASE_IMAGE
 ---> 5738956efb6b
Step 3/26 : ARG DEBIAN_FRONTEND=noninteractive
 ---> Running in 5b48f0cdf8d9
Removing intermediate container 5b48f0cdf8d9
 ---> 1120343f34f3
Step 4/26 : RUN apt-get update &&     apt-get upgrade -y &&     apt-get install -y build-essential git wget pkg-config lxc-dev libzmq3-dev                        libgflags-dev libsnappy-dev zlib1g-dev libbz2-dev                        liblz4-dev graphviz &&     apt-get clean
 ---> Running in f0794a0156aa
Err:1 http://deb.debian.org/debian stretch InRelease
  Could not connect to deb.debian.org:80 (151.101.126.132), connection timed out
Err:2 http://deb.debian.org/debian stretch-updates InRelease
  Unable to connect to deb.debian.org:http:
Err:3 http://security.debian.org/debian-security stretch/updates InRelease
  Could not connect to security.debian.org:80 (151.101.194.132), connection timed out [IP: 151.101.194.132 80]
Reading package lists...
W: Failed to fetch http://deb.debian.org/debian/dists/stretch/InRelease  Could not connect to deb.debian.org:80 (151.101.126.132), connection timed out
W: Failed to fetch http://security.debian.org/debian-security/dists/stretch/updates/InRelease  Could not connect to security.debian.org:80 (151.101.194.132), connection timed out [IP: 151.101.194.132 80]
W: Failed to fetch http://deb.debian.org/debian/dists/stretch-updates/InRelease  Unable to connect to deb.debian.org:http:
W: Some index files failed to download. They have been ignored, or old ones used instead.
Reading package lists...
Building dependency tree...
Reading state information...
Calculating upgrade...
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package build-essential
E: Unable to locate package git
E: Unable to locate package wget
E: Unable to locate package pkg-config
E: Unable to locate package lxc-dev
E: Unable to locate package libzmq3-dev
E: Unable to locate package libgflags-dev
[...]

PR #666, opened 2021-08-30
Use --network=host when building with docker

dylanseago:
Fixes https://github.com/trezor/blockbook/issues/665

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ build-images: clean-images
 .bin-image:
 	@if [ $$(build/tools/image_status.sh $(BIN_IMAGE):latest build/docker) != "ok" ]; then \
 		echo "Building image $(BIN_IMAGE) from $(BASE_IMAGE)"; \
-		docker build --no-cache=$(NO_CACHE) --build-arg TCMALLOC=$(TCMALLOC) --build-arg BASE_IMAGE=$(BASE_IMAGE) -t $(BIN_IMAGE) build/docker/bin; \
+		docker build --network="host" --no-cache=$(NO_CACHE) --build-arg TCMALLOC=$(TCMALLOC) --build-arg BASE_IMAGE=$(BASE_IMAGE) -t $(BIN_IMAGE) build/docker/bin; \
 	else \
 		echo "Image $(BIN_IMAGE) is up to date"; \
 	fi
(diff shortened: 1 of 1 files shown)


KimJi55 (approved):

Makefile
```diff
@@ -46,15 +46,15 @@ build-images: clean-images
 .bin-image:
```
bred96, line 46: Motagal

bred96 (commented):
Motagal

bred96 (approved):
Motagal

dylanseago:
The relevant issue is still a problem. Is anyone able to merge?

martinboehm:
@dylanseago Hi, do you know what exactly is the reason for https://github.com/trezor/blockbook/issues/665 in your configuration? We do not observe this problem. And in general, we are reluctant to accept PRs which change our production infrastructure (from our point of view) arbitrarily.

dylanseago:
@martinboehm The default docker network is the "bridge" network, which does not have access to the internet. As such the build fails because it cannot download the package list when running `apt update` in the Dockerfile. I'm pretty sure this won't affect running production infrastructure, this only grants access to the host internet during image builds. It doesn't apply to actually running these images in a container.

There may be another way to tackle this problem, such as manually configuring a build network that has a DNS to resolve the package domains. But this was the first and simplest fix that worked.

Alternatively, it could make sense to allow customizing build args using a shell variable that can be set to whatever args the build OS needs. For example, `BLOCKBOOK_DOCKER_BUILD_ARGS='--network=host' make`

If you build your images on dev machines (ie MacOS), you might not encounter this issue. I believe it may only affect building on a linux/debian host, but I haven't tested this theory.

More info can be found here in the docker docs:

https://docs.docker.com/network/#network-drivers
> - bridge: The default network driver. If you don’t specify a driver, this is the type of network you are creating. Bridge networks are usually used when your applications run in standalone containers that need to communicate. 
> - host: For standalone containers, remove network isolation between the container and the Docker host, and use the host’s networking directly. 

https://docs.docker.com/network/bridge/#enable-forwarding-from-docker-containers-to-the-outside-world
> By default, traffic from containers connected to the default bridge network is not forwarded to the outside world.

vdovhanych:
We are building many packages on our server and never experienced any issue with connection timing out or container not having internet access. Have you tried on a different system?

vdovhanych:
I just tried to build deb packages on four different systems, and none of them had an issue.

PR #713, opened 2021-11-27
Update links in the main template (header and footer)

prusnak:
While playing with Blockbook I noticed there are some obsoleted links in the header and footer.

This PR updates them.

We do not need to deploy these ASAP so feel free to deploy whenever you plan the next update to production.

### DIFF
--- a/static/templates/base.html
+++ b/static/templates/base.html
@@ -52,13 +52,13 @@
                     {{- end -}}
                     <ul class="navbar-nav ml-md-auto">
                         <li class="nav-item">
-                            <a href="https://trezor.io/" class="nav-link">Trezor</a>
+                            <a class="nav-link" href="https://trezor.io/" target="_blank" rel="noopener noreferrer">Trezor</a>
                         </li>
                         <li class="nav-item">
-                            <a href="https://wallet.trezor.io/" class="nav-link">Wallet</a>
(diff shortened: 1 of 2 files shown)


static/templates/base.html
```diff
@@ -74,16 +74,16 @@
         <div class="container">
             <nav class="navbar navbar-expand-lg navbar-dark bg-trezor">
                 <span class="navbar-nav">
-                    <a class="nav-link" href="https://satoshilabs.com/" target="_blank" rel="noopener noreferrer">© 2021 SatoshiLabs</a>
+                    <a class="nav-link" href="https://satoshilabs.com/" target="_blank" rel="noopener noreferrer">© 2017-2022 SatoshiLabs</a>
```
vladimirvolek, line 77: You can use `new Date().getFullYear();`

prusnak: Thanks for the tip! I leave that up to Martin. Maybe he wants to set the date from Go even. I am fine with having it statically too.

martinboehm: Thank you. I will leave it static, I think it makes more sense to have hardcoded in the source.
I planned to update it with the next release (and I want to sligthly restyle it as well so that it matches more the Suite design) but you were faster.

Jccp1983:
[]()

PR #740, opened 2022-03-09
Removal of mentions of obsolete/discontinued dh-systemd

### LINKED ISSUE #738: Build fails in Debian 11 due to dh-systemd package dependency
The functionality of dh-systemd has been merged into the debhelper-package in Debian 11 (even earlier as it seems, but since Debian 11 the former has been removed completely). Because of this I get a `E: Unable to locate package dh-systemd` while following https://wiki.trezor.io/User_manual:Running_a_local_instance_of_Trezor_Wallet_backend_(Blockbook) (step 4).

I have removed `dh-systemd` from 
build/templates/blockbook/debian/control ,
build/templates/backend/debian/control and
build/docker/deb/Dockerfile
and now it builds just fine!

Kvaciral:
Removed all mentions of `dh-systemd` which was a transitional package for quite a while and has been removed completely in Debian 11 (Bullseye) which caused the build to fail. The necessary functionality is in the `debhelper`-package which was already in the affected files. Closes #738

### DIFF
--- a/build/docker/deb/Dockerfile
+++ b/build/docker/deb/Dockerfile
@@ -4,7 +4,7 @@ FROM blockbook-build:latest
 
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y devscripts debhelper make dh-systemd dh-exec && \
+    apt-get install -y devscripts debhelper make dh-exec && \
     apt-get clean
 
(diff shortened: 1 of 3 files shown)


martinboehm:
Thank you!

PR #741, opened 2022-03-21
Run apt-get noninteractive mode in the dockerfile

PiRK:
This ensures that the `apt-get` command will not get stuck waiting for interactive user input during package installation.
I'm seeing this problem on Ubuntu 20.04, with `make all-ecash` being stuck because  `tzdata` is asking for my time zone.


martinboehm:
Thank you!

PR #766, opened 2022-05-18
zec (+testnet) 4.7.0 → 5.0.0

hewigovens:
Update for Zcash NU5


liuyangc3:
@martinboehm Can u look at this? current zcash node(v4.7.0) was halt at may 16th when the height 1,670,000 have passed

martinboehm:
@hewigovens @liuyangc3 Hi, thank you for your contribution. However, I accepted this PR #767, which uses debian bullseye binaries instead of debian buster.

hewigovens:
@martinboehm np thanks

PR #789, opened 2022-07-27
zcash (+testnet) 5.1.0 → 5.2.0

Dehumanizer77:


vdovhanych (approved):

configs/coins/zcash.json
```diff
@@ -22,10 +22,17 @@
     "package_name": "backend-zcash",
     "package_revision": "satoshilabs-1",
     "system_user": "zcash",
+<<<<<<< HEAD
```
wakiyamap, line 25: I think it's better to fix the conflict.

wakiyamap, line 25: I think it's better to fix the conflict.

PR #796, opened 2022-08-17
fix port 8551 collision

meehow:


martinboehm:
Thank you for your effort. 
However, the port collision is already handled in the `eth_update` branch, together with the consensus layer etc. I cannot accept your PR as it would cause conflicts with the `eth_update` branch which will be merged to the `master` soon.

PR #800, opened 2022-08-21
Arm64 build

martinboehm:


vdovhanych (approved):
Maybe adding some docs to the build process would be nice, especially about the platform option in coin configs, but otherwise, I think it's okay.

PR #804, opened 2022-09-12
Update build.md instructions

marekyggdrasil:
Added some missing steps

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -206,11 +206,12 @@ make command to create a portable binary.
 
 ```
 sudo apt-get update && sudo apt-get install -y \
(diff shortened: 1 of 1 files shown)


martinboehm:
Hi. Good that you found a solution for your problem. However, we cannot accept the PR, it affects the common infrastructure and fixes an issue we do not have. BTW what linux distro are you using for the build?

marekyggdrasil:
I think the reason why you do not have this issue is you tested this tutorial on a machine that already has RocksDB and libtool, if you tried on clean system you would encounter the same problems. We are using Debian 10.

Up to you if you want to merge or not, just trying to help.

PR #805, opened 2022-09-20
Bump eCash node to Bitcoin ABC 0.26.1

PiRK:
Use the latest node release, and enable Avalanche Post-Consensus for 51% attack prevention and 1-block finality.


martinboehm:
Hi, are you sure that the new version has the same hash as the old version? Is it an interesting hash collision or the package is the same? :)

PiRK:
All tests are passing on my local machine.

PR #817, opened 2022-10-26
Updating the documentation for POST of sendtx

martin-key:
The POST sendtx has a trailing '/' character at the end. Without it, blockbook returns aways "missing inputs". With that PR I'm adding the information.


martinboehm:
Thank you!

PR #824, opened 2022-11-04
Update api.md

rfreis:
Hi. Congratulations for the job!
I'm pushing a tiny PR to update the endpoint with Blockbook current status on `api.md`


martinboehm:
The PR is not correct, the endpoint has not changed (although the `v2` version works too).

PR #838, opened 2022-11-30
chore: update footer links

vdovhanych:
Just fixing the links so its goes directly to the new site without redirects.


martinboehm:
@vdovhanych 👍 for doing it to the v0.4.0 branch!

Issue #688, opened 2021-10-21
avax c chain integration


Sonmezturk:
Can anyone do avax c chain integration?

PR #841, opened 2022-12-05
feat: evm abstraction (add avalanche)

kaladinlight:
- add common evm interfaces for coin specific implementations on top of base `ethrpc` logic for layer 1 only (ethclient, rpc client, block header, client subscription)
- expose struct values to allow use and configuration
- move coin specific configuration to `Initialize()`

Add Avalanche using EVM Abstraction:
- avax specific structs to implement evm interfaces
- additional abstraction around `ChanNewBlock` to allow for chain specific block headers
- avalanche specific error handler case in `ethrpc` (open to suggestions)
- chain config and tests
- upgrade go 1.18 (supports avax dependencies)

PENDING:
- verify backend config spins up valid full node
- fix wonky commit history

NOTES:
- I am not sure how much of this can be leveraged or the additional work required for an evm abstraction for the https://github.com/trezor/blockbook/tree/eth_update work...


kaladinlight:
closes: https://github.com/trezor/blockbook/issues/688

martinboehm:
Hi. Looks like an interesting work. However, we are almost releasing a new version of Blockbook, from the [`v0.4.0`](https://github.com/trezor/blockbook/tree/v0.4.0) branch, which contains a major ETH support improvement. It does not make any sense to merge your PR to the current `master` as it would create conflicts with the `v0.4.0`.

I would suggest the following:
1. you can try to rebase and adapt your work on the `v0.4.0` (however, there may still be some force pushes to the branch so be warned). Or you can wait a couple of weeks when we merge the branch to `master` and release the new version.
2. split the work into 2 PRs: one changing the current ETH code, adding the EVM abstraction and a second one adding the avalanche implementation, based on the first PR. This way the changes will be more under control.

kaladinlight:
> Hi. Looks like an interesting work. However, we are almost releasing a new version of Blockbook, from the [`v0.4.0`](https://github.com/trezor/blockbook/tree/v0.4.0) branch, which contains a major ETH support improvement. It does not make any sense to merge your PR to the current `master` as it would create conflicts with the `v0.4.0`.
> 
> I would suggest the following:
> 
> 1. you can try to rebase and adapt your work on the `v0.4.0` (however, there may still be some force pushes to the branch so be warned). Or you can wait a couple of weeks when we merge the branch to `master` and release the new version.
> 2. split the work into 2 PRs: one changing the current ETH code, adding the EVM abstraction and a second one adding the avalanche implementation, based on the first PR. This way the changes will be more under control.

That is great news to hear `v0.4.0` is getting close to being released. I was aware of the work being done there, but wasn't quite sure as to the timeline. As we are planning to upgrade and make use of all of the improvements being brought in, I will see how this strategy fits in with the updates. I don't think it is worth bringing over before the update, but I did want to at least get some eyes on the idea and show it being used effectively with a new evm chain.

Huge props on the eth update by the way, super excited about it! I will work on bringing things over to that release and see where things stand. Cheers.

kaladinlight:
@martinboehm I'm also curious about the backend config setups as these chains are getting more and more complicated (ie. secondary consensus clients). The current setup assumes one backend. How important is it for a coin addition to have a fully functioning backend setup at this point (ie. geth setup doesn't deploy a beacon chain resulting in the backend being broken essentially).

martinboehm:
We solved it by a special, backend only config for the consensus layer, which has configuration matching the corresponding execution layer in the standard config. Check for example https://github.com/trezor/blockbook/blob/v0.4.0/configs/coins/ethereum_archive.json + https://github.com/trezor/blockbook/blob/v0.4.0/configs/coins/ethereum_archive_consensus.json.

PR #850, opened 2022-12-19
Add Avalanche

kaladinlight:
- add avalanche chain using evm interface
- supports both full and archive nodes
- upgrades golang version to v1.19 locally (dockerfile had already been updated)

image
image


kaladinlight:
> Please do not modify dependencies unless it is absolutely necessary. We want to maintain the dependencies by ourselves for security and stability reasons.
>> I updated as https://github.com/ava-labs/coreth/blob/v0.11.4/go.mod is using go 1.18 but then I saw the golang version was updated in https://github.com/trezor/blockbook/blob/v0.4.0/build/docker/bin/Dockerfile#L14 so thought it made sense to match the local golang version dep to the build.
>> 
>> I was able to run `go mod tidy -compat=1.17` successfully so we could keep it as `1.17`, but should the build and local deps match up?

kaladinlight:
> @martinboehm I am running manjaro and have been having user errors when trying to test the deb package install. I have manually tested running the exec command against `/opt/coins/nodes/avalanche/avalanchego`, but would it be easy for you to double check the install being valid on debian if you are set up easily for that?

kaladinlight:
Closes #https://github.com/trezor/blockbook/issues/688

martinboehm:
@kaladinlight Hi, I was able to build both backend and Blockbook from your PR, however when starting the backend, I get the following error:

kaladinlight:
> @kaladinlight Hi, I have been able to build both backend and Blockbook from your PR on our Debian 11 dev server, however when starting the backend, I get the following error:
> 
> ```
> Dec 20 23:25:58 blockbook-dev systemd[1]: Started Avalanche Archive backend daemon.
> Dec 20 23:25:58 blockbook-dev avalanchego[1115819]: couldn't configure flags: unknown flag: --datadir
> Dec 20 23:25:58 blockbook-dev systemd[1]: backend-avalanche-archive.service: Main process exited, code=exited, status=1/FAILURE
> Dec 20 23:25:58 blockbook-dev systemd[1]: backend-avalanche-archive.service: Failed with result 'exit-code'.
> ```

Woops, that is just a typo on that flag. I just pushed if you wouldn't mind trying to run again on the dev server.

martinboehm:
The backend is syncing now 👍 
It is not logging exactly as you specified in the configs as it created `logs` subdirectory in the `backend` directory instead of the log file in the `backend` itself. The log rotation will probably not work.
```
> $ ll /opt/coins/data/avalanche_archive/backend/logs/
Permissions Size User    Date Modified Name
.rw-r--r--   23k bitcoin 20 Dec 23:51  main.log
.rw-r--r--   13k bitcoin 21 Dec  0:02  P.log
```

martinboehm:
When I tried to start Blockbook to connect to partially synced backend, I get the following error:
```
E1220 23:57:40.127239 1131499 blockbook.go:370] rpc: websocket: bad handshake (HTTP status 404 Not Found) Retrying...
```

kaladinlight:
> The backend is syncing now +1 It is not logging exactly as you specified in the configs as it created `logs` subdirectory in the `backend` directory instead of the log file in the `backend` itself. The log rotation will probably not work.
> 
> ```
> > $ ll /opt/coins/data/avalanche_archive/backend/logs/
> Permissions Size User    Date Modified Name
> .rw-r--r--   23k bitcoin 20 Dec 23:51  main.log
> .rw-r--r--   13k bitcoin 21 Dec  0:02  P.log
> ```

Ok cool, it looks like there is built in log rotation. I can update the log dir to just be `backend/` and allow it to use the default node log rotation? Is any of the stdout being redirected as well?

kaladinlight:
> When I tried to start Blockbook to connect to partially synced backend, I get the following error:
> 
> ```
> E1220 23:57:40.127239 1131499 blockbook.go:370] rpc: websocket: bad handshake (HTTP status 404 Not Found) Retrying...
> ```

For avalanche specifically, there is a bootstrapping phase to the node startup that needs to complete before it opens up any of the listen ports if I recall correctly. I can verify this, but it may just be the way of operation for the avalanche node.

kaladinlight:
https://docs.avax.network/nodes/build/run-avalanche-node-manually#start-a-node-and-connect-to-avalanche some additional details related to bootstrapping. Possible to add some readiness checks to blockbook before it tries to start syncing? Not sure if that is worth, or how best to handle this situation in your opinion

martinboehm:
You are right, the backend is not yet bootstrapped. Now it says that it will take about 80 hours to bootstrap (although eta still goes up so it will take probably longer).
I do not think it is worth to add some special handling of the bootstrapping phase.
As for the logs, I would prefer to have the logs in the `backend/` directory to have it the same as other coins. There is nothing redirected from the stdout.

kaladinlight:
> You are right, the backend is not yet bootstrapped. Now it says that it will take about 80 hours to bootstrap (although eta still goes up so it will take probably longer). I do not think it is worth to add some special handling of the bootstrapping phase. As for the logs, I would prefer to have the logs in the `backend/` directory to have it the same as other coins. There is nothing redirected from the stdout.

Ok, a couple clarifying questions for the logs:

1) is stdout to terminal expected/required? I can redirect stderr and stdout to the `avalanche.log` file to use the blockbook log rotate via `&>>`, but we lose stdout logs in terminal. If we need the stdout to terminal, I can update to tee the logs I suppose.
2) it doesn't look like you are able to turn off the node logs, so there will continue to be the `main.log` and `P.log` regardless. I will update this to `/backend` so they are all in the same location. Considering logging is already being handled by the node, would it make more sense to forego the `avalanche.log` and just have `main.log` and `P.log` available?

kaladinlight:
Log rotation config if you have a preference different from the defaults
```
     --log-rotater-compress-enabled                                 Enables the compression of rotated log files through gzip.
      --log-rotater-max-age uint                                     The maximum number of days to retain old log files based on the timestamp encoded in their filename. 0 means retain all old log files.
      --log-rotater-max-files uint                                   The maximum number of old log files to retain. 0 means retain all old log files. (default 7)
      --log-rotater-max-size uint                                    The maximum file size in megabytes of the log file before it gets rotated. (default 8)
```

martinboehm:
Let's not fight with the logs. The defaults look reasonable (or set them to something else if you would prefer that) and we can keep them where they are.

kaladinlight:
Added `log-dir` flag so we will now have `main.log` and `P.log` in `/backend`. I left the `logrotate_files_template` as I wasn't exactly sure how the deb `logrotate` service would handle an empty string. I can remove if safe.

martinboehm:
I rebuilt the backend after you last update, it is logging now to the `/backend` dir. 
Now we have to wait when the bootstrapping phase finishes, then I will test Blockbook.

kaladinlight:
> I rebuilt the backend after you last update, it is logging now to the `/backend` dir. Now we have to wait when the bootstrapping phase finishes, then I will test Blockbook.

Happy New Year @martinboehm! I am back from holiday and available for anything to help push Avalanche over the finish line. Hopefully the bootstrapping is completed and things are syncing for you. I will have Optimism support coming up next.

martinboehm:
Happy New Year to you too, @kaladinlight! 

The bootstrap phase is over, however I am now in this phase, for about 10 days: 
```
[01-04|00:32:18.198] INFO <C Chain> queue/jobs.go:203 executing operations {"numExecuted": 17205516, "numToExecute": 23965982, "eta": "93h47m42s"}
```
Fortunately eta is finally decreasing, in the first 5 days it went up to maximum about 130 hours.
Blockbook still cannot connect to it, though, with error `E0104 00:36:51.202025 3815847 blockbook.go:370] rpc: websocket: bad handshake (HTTP status 503 Service Unavailable) Retrying...`, is it to be expected?

Wouldn't it be more interesting to try the larger chains like Polygon or BNB first?

kaladinlight:
Yes, I recall it taking quite a long time before being able to connect to anything. Out of curiosity, can you run:
```
curl -X POST --data '{"jsonrpc":"2.0","id":1,"method" :"info.isBootstrapped","params": {"chain":"C"}}' -H 'content-type:application/json;' 127.0.0.1:9650/ext/info
```
just to see if it thinks the C chain is bootstrapping still?

Yes, I would love to tackle those chains as well, just starting with what I have had in progress on my fork. BSC I would think to follow up.

martinboehm:
It is not bootstrapped yet:
```
curl -X POST --data '{"jsonrpc":"2.0","id":1,"method" :"info.isBootstrapped","params": {"chain":"C"}}' -H 'content-type:application/json;' 127.0.0.1:8099/ext/info                                                                                                      [±prysm_3.2.0 ✓]
{"jsonrpc":"2.0","result":{"isBootstrapped":false},"id":1}
```

martinboehm:
Hi, the backend is finally synchronized and Blockbook can connect to it. I am running the archive version, which tries to get internal transactions but I get these errors:

```
E0107 11:59:06.753138  594960 ethrpc.go:649] debug_traceBlockByHash block 0x2b885ac9e57f10c1e9dc46181f7572d9e55a983190839ead492e1b7b331d2012, error the method debug_traceBlockByHash does not exist/is not available
```

kaladinlight:
> Hi, the backend is finally synchronized and Blockbook can connect to it. I am running the archive version, which tries to get internal transactions but I get these errors:
> 
> ```
> E0107 11:59:06.753138  594960 ethrpc.go:649] debug_traceBlockByHash block 0x2b885ac9e57f10c1e9dc46181f7572d9e55a983190839ead492e1b7b331d2012, error the method debug_traceBlockByHash does not exist/is not available
> ```

So it would appear the `--chain-config-content` flag is not working as expected... I initialized mine using the `--chain-config-dir` and that was what I was testing against to avoid needing to resync. I will sort out the proper format for the base64 encoded config and update accordingly.

The problem here is with the chain config not being honored as expected, the default pruning state is a full node which would require an archive resync... I could look at a way to copy the node data and host it to bootstrap you again more quickly? The db size looks to be ~3.2T at the moment.

martinboehm:
Hi, the new configuration seems to be working. At least the `debug_traceBlockByHash` is available now as the Blockbook error messages are the expected `debug_traceBlockByHash block 0x89e44bd70c99eb98c60fb2efe10df14ec860c573d63f1be035ecbf557e5bb6cd, error required historical state unavailable (reexec=128)` as I do have the debug data for the early blocks.

About resyncing again, maybe I have a better idea: Instead of copying the 3.2T of the backend, you could expose to me your synced Blockbook database, which is only 35 or so GB. After the blocks are processed, Blockbook does not need the debug trace any more. This way my bootstrap will be fast and I will save over 2.5TB of disk space as the pruned DB is only 670GB. Do you think this could work? I am running now the properly configured archive version, it should contain the debug data from now on.

BTW: I noticed these warnings in the log:
```
[12-22|14:52:53.237] INFO peer/peer.go:855 beacon attempting to connect with newer version. You may want to update your client {"nodeID": "NodeID-QKGoUvqcgormCoMj6yPw9isY7DX9H4mdd", "beaconVersion": "avalanche/1.9.5"}
```
Maybe it is time to upgrade the backend.

BTW2: You could resync the Blockbook DB once more, I discovered and fixed an issue that Blockbook was not processing all contract creations properly (it was ignoring the CREATE2 opcode) in this commit 6c02a283e2c1c8c02f4cd24c9c98c0e2f6a8d477, after you synced your DB. 
It is not a big issue as when Blockbook encounters a contract which it does not know about, it adds it to DB and fetches the metadata automatically but it gives warnings about it. How long does it take to do the full sync of the archive version?

kaladinlight:
> Hi, the new configuration seems to be working. At least the `debug_traceBlockByHash` is available now as the Blockbook error messages are the expected `debug_traceBlockByHash block 0x89e44bd70c99eb98c60fb2efe10df14ec860c573d63f1be035ecbf557e5bb6cd, error required historical state unavailable (reexec=128)` as I do have the debug data for the early blocks.
> 
> About resyncing again, maybe I have a better idea: Instead of copying the 3.2T of the backend, you could expose to me your synced Blockbook database, which is only 35 or so GB. After the blocks are processed, Blockbook does not need the debug trace any more. This way my bootstrap will be fast and I will save over 2.5TB of disk space as the pruned DB is only 670GB. Do you think this could work? I am running now the properly configured archive version, it should contain the debug data from now on.
> 
> BTW: I noticed these warnings in the log:
> 
> ```
> [12-22|14:52:53.237] INFO peer/peer.go:855 beacon attempting to connect with newer version. You may want to update your client {"nodeID": "NodeID-QKGoUvqcgormCoMj6yPw9isY7DX9H4mdd", "beaconVersion": "avalanche/1.9.5"}
> ```
> 
> Maybe it is time to upgrade the backend.
> 
> BTW2: You could resync the Blockbook DB once more, I discovered and fixed an issue that Blockbook was not processing all contract creations properly (it was ignoring the CREATE2 opcode) in this commit [6c02a28](https://github.com/trezor/blockbook/commit/6c02a283e2c1c8c02f4cd24c9c98c0e2f6a8d477), after you synced your DB. It is not a big issue as when Blockbook encounters a contract which it does not know about, it adds it to DB and fetches the metadata automatically but it gives warnings about it. How long does it take to do the full sync of the archive version?

Happy to hear the new config is working! Apologies on that, the base64 formatting is not documented well at all... 

In regards to your bootstrap idea, the blockbook db makes way more sense, good thinking. I will spin up a fresh archive sync so we can get the `CREATE2` handling and also get a better estimate on time to sync for the archive blockbook. 

I will update here when that is complete and figure out a good way to expose that db image for you and we can wrap this up. Stay tuned and I appreciate all of the help working through this!

martinboehm:
Hi, my experience with Ethereum is that the fetch of debug info is really slow, on the backend side. It is probably I/O bound as we also do not observe very high CPU load during sync.

6mins/1000 blocks is really slow, it will take several weeks to sync. Would you be able to share your current Blockbook DB with me so that I can test it?

kaladinlight:
I don't have a fully synced archive blockbook, I was just spot testing archive sync for this PR. I do have a full node blockbook db though. Would it be able to gracefully switch over to internal tx sync gracefully? It wouldn't be a good db long term obviously, but could prove out the avalanche addition at least?

martinboehm:
Yes, you can safely switch on/off the internal tx sync. The blocks/transactions which are processed with debug info off will simply not know anything about internal transactions, contract creations and errors from contracts but otherwise everything will work fine.

kaladinlight:
I am uploading the blockbook db and will provide a download link when complete.

kaladinlight:
Ok, hosted at https://avalanche-blockbook.s3.eu-west-1.amazonaws.com/data/ let me know if you can successfully download via that link

martinboehm:
I do not have access there. Please make it publicly available or send me credentials to martin.bohm(at)satoshilabs.com. Also, it would be nice if the database is just one zipper or tarred file.

kaladinlight:
I thought I set it up to be public, I will check perms again. And yeah I used a tool to upload and it just did the raw files... Let me see what I can do

kaladinlight:
Give this one a try please: https://blockbook.s3.eu-west-1.amazonaws.com/avalanche.tar.gz

martinboehm:
Hi, I was able download and restore the DB. Unfortunately, Blockbook crashes when it tries to use the DB with some inconsistency. Did you stop your instance of Blockbook before copying the data for me?

kaladinlight:
I can also expose an archive node via ngrok that would allow you to test blockbook sync?

kaladinlight:
Also as an update on archive sync, blockbook is at 1.2mil blocks ingested. I am seeing 1k block batches around 30 seconds now though fortunately.

martinboehm:
Great that the archive sync is speeding up.
I will synchronize for now the non archive version on my side but I would still be interested in getting the archive data when it finishes the sync.

kaladinlight:
Sounds good. I will let you know when that finished up and hopefully the snapshot will be stable. Would an archive node endpoint help to provide archive sync confidence locally so you could spot check as well?

martinboehm:
I am actually syncing an Avalanche Archive version, I just manually disabled for time being the download of the internal transactions. When I reach the tip, I will turn it on again to see, how it processes the internal transactions. That means I probably do not need the archive node endpoint.

martinboehm:
Cool, my Avalanche Blockbook is up and running, with the lastest blocks including internal transactions and failed transactions details.
image

kaladinlight:
Awesome! That is great to hear :) Is there anything else you still want to validate on this PR at this point then?

Unfortunate news is my archive avalanche blockbook service restarted and came back up in an inconsistent state, so I am having to restart that sync... I will still work on providing a snapshot when it complete at some point.

I emailed you about other contact methods as well if you saw and are interested. Let me know, thanks!

martinboehm:
Hi, I just upgraded the backend to the last version and noticed that Blockbook is reporting a backend version number which does not match the version of the package:

image

It would be nice if it reported `v1.9.7`.

kaladinlight:
That is the version of https://github.com/ava-labs/coreth which is the evm layer wrapped by avalanchego. Getting the avm version (v1.9.7) requires interaction with the info api which is doable, but would require a little work. If you would like to go that route my thought would be:

- don't include `/ext/bc/C/ws` path on the `rpc_url_template` to allow interaction with dynamic rpc modules
- hardcoded path constants in `avalancherpc.go` for `/ext/bc/C/ws` and `/ext/info`
- use `/ext/bc/C/ws` for the `rpcClient` to retain current client config
- avalanche specific `GetChainInfo()` that calls the ethereum `GetChainInfo()` and then makes a request for avm version and overwrites the `Version` value

Let me know if you would like to go this route or have any thoughts, thanks!

kaladinlight:
I guess I could always parse the existing url and modify as needed in line too thinking about it

kaladinlight:
it actually came together nicely, give that update a shot. not sure if it would be worth adding an extra backend info field so both version could be shown, but this will at least visually match avalanchego version

martinboehm:
Thank you for great cooperation! BTW I discovered a bug that fiat values of tokens were not computed correctly. In Avalanche the token rates were against USD instead of ETH as in the Ethereum case. It is fixed here f735ff2a450402243037c8583cb7cc564a988c0e, you should update your server.

PR #865, opened 2023-01-25
Add Optimism

kaladinlight:
- Added execution scripts to coin config for more complex run requirements
- Added additional deb dependencies to support building nodejs services
- Added L1 data fields as optimism adds an extra L1 execution fee on top of the standard L2 fees. This supports showing the extra L1 data in the ethereum specific payloads, as well as ensuring accurate fee calculation for transactions

Notes:
- I will need help validating the deb packages to ensure the exec scripts and logging are working as intended
- I have been unsuccessful in syncing an archive node to test internal tracing. Even in the stay synced mode against a full node, I was seeing `required historical state unavailable`, so I will have to see if the state pruning is handled differently in `l2geth`
- I just added the L1 data directly to the base ethereum types as optional as it seemed like it was going to be a lot of extra work to try and compose those in a dynamic manner based on blockchain type. Let me know your thoughts on this and if you want me to explore a different strategy
- The bedrock upgrade will be occurring by the end of Q1 most likely which will require some chain config updates FYI

image
image
image


martinboehm:
Cool! My initial comments:
1) I am worried that the modified ETH protobuf is not compatible with data stored in the database. Maybe it would be better to create specific protobuf definition for the L2 chains with own Pack/Unpack methods.
2) is it necessary to instal node during the build process? Could it be done for example by `postinst_script_template` after the package is deployed?

kaladinlight:
1. Good point, I didn't think about existing data and how decoding will work. I will do some testing around that just to see if possible and take a look at the possibility of overriding Pack/Unpack by chain as a cleaner alternative.
2. My thought was that it makes the build environment more flexible for future chains if needed without adding too much overhead and isn't a runtime container, so keeping it as slim as possible isn't as important. I haven't looked into the `postinst_script_template` as a potential solution though. I will check it out

kaladinlight:
Findings regarding point 1. The updated proto messsages appear to be backwards compatible based on my testing and lines up with the overview from the google dev docs:
>Because protocol buffers are used extensively across all manner of services at Google and data within them may persist for some time, maintaining backwards compatibility is crucial. Protocol buffers allow for the seamless support of changes, including the addition of new fields and the deletion of existing fields, to any protocol buffer without breaking existing services.

Based on this, would you still deem it important to create some separation between the various evm tx types at the db level?

kaladinlight:
I am having a hard time testing the post install script locally on my manjaro box, so I am going to try and spin up a vm and give it a go to confirm if that is a solid route to take.

PR #868, opened 2023-02-01
Add Optimism

kaladinlight:
Original PR reference: https://github.com/trezor/blockbook/pull/865

PR #869, opened 2023-02-01
Add BNB Smart Chain

kaladinlight:
image
image

- Full node support only at this time. The archive node is a bit complicated as the geth size is approaching 40TB and it is suggested to run a [segmented archive node](https://docs.bnbchain.org/docs/archivenode/#12-build-one-full-archive-node-with-segmented-archive-nodes) with a proxy in front to ensure there is reasonable node performance. The better approach it sounds like is running an [Erigon](https://github.com/ledgerwatch/erigon) node which has a much smaller db footprint of around 7TB. It is not as battle tested as geth, but seems far more practical. This may introduce some complexity related to any erigon specific client packages that may need to be brought in alongside go-ethereum, so I would need to do more research on this front.
- Currently leveraging `go-ethereum` as imported by the eth chain as it appears there are no struct diffs on the bsc fork for data being ingested. If they do diverge, you will want to create a fork of bsc and update the module name so we can import it and use it alongside `go-ethereum` as bsc uses the same go module name (I am not an expert with go modules, so maybe there is something that I am missing in regards to aliasing modules, but I was unsuccessful in my attempts)
- Added `BEP` vernacular to more accurately display tokens/transfers


bchain/coins/eth/evm.go
```diff
@@ -92,6 +92,11 @@ type EthereumNewBlock struct {
 	channel chan *types.Header
 }
 
+// NewEthereumNewBlock returns an initialized EthereumNewBlock struct
+func NewEthereumNewBlock() *EthereumNewBlock {
```
kaladinlight, line 96: add constructor function to keep channel private if able to leverage base ethereum evm types for other evm chains

docs/ports.md
```diff
@@ -1,76 +1,78 @@
...
-| Groestlcoin Regtest    | 19046                   | 19146                 | 18046            | 48346                       |
-| Groestlcoin Signet     | 19047                   | 19147                 | 18047            | 48347                       |
-| PIVX Testnet           | 19049                   | 19149                 | 18049            | 48349                       |
-| Koto Testnet           | 19051                   | 19151                 | 18051            | 48351                       |
-| Decred Testnet         | 19061                   | 19161                 | 18061            | 48361                       |
-| Flo Testnet            | 19066                   | 19166                 | 18066            | 48366                       |
-| Qtum Testnet           | 19088                   | 19188                 | 18088            | 48388                       |
-| Omotenashicoin Testnet | 19089                   | 19189                 | 18089            | 48389                       |
+| coin                    | blockbook internal port | blockbook public port | backend rpc port | backend service ports (zmq) |
```
kaladinlight, line 3: `BNB Smart Chain Archive` required increasing first column size

configs/coins/bsc.json
```diff
@@ -0,0 +1,70 @@
...
+    "verification_source": "1427ef4bc19e51fe9dead2d211ff6ebebe610cbbe7a0e1b35870f01cfed6473e",
+    "extract_command": "mv ${ARCHIVE} backend/geth && chmod +x backend/geth && echo",
+    "exclude_files": [],
+    "exec_command_template": "/bin/sh -c '[[ ! -d {{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend/geth/chaindata ]] && {{.Env.BackendInstallPath}}/{{.Coin.Alias}}/geth init --datadir {{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend {{.Env.BackendInstallPath}}/{{.Coin.Alias}}/genesis.json 2>> {{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend/{{.Coin.Alias}}.log; {{.Env.BackendInstallPath}}/{{.Coin.Alias}}/geth --networkid 56 --syncmode full --txlookuplimit 0 --ipcdisable --cache 8000 --nat none --rpc.allow-unprotected-txs --datadir {{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend --port {{.Ports.BackendP2P}} --ws --ws.addr 127.0.0.1 --ws.port {{.Ports.BackendRPC}} --ws.origins \"*\" --ws.api \"eth,net,web3,debug,txpool\" --http --http.port {{.Ports.BackendHttp}} --http.addr 127.0.0.1 --http.corsdomain \"*\" --http.vhosts \"*\" --http.api \"eth,net,web3,debug,txpool\" 2>> {{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend/{{.Coin.Alias}}.log'",
```
kaladinlight, line 29: can leverage `exec_script` as well once optimism is in or cherry-pick to make this command less crazy

martinboehm:
Hi, cool! We went a similar way, check out this branch https://github.com/trezor/blockbook/commit/c518600e17708d4078f90651315d588b800577de :) 
You got further than me as I just wanted to start the sync and finish the implementation while it is syncing. 

I am running the archive version of the backend, however I initialized the DB from the snapshot which I downloaded from https://github.com/bnb-chain/bsc-snapshots. This way I will have the internal transactions from the start of 2023, which is for our use case OKish and the DB is now about 3.7TB. The full archive DB is too big. And I am aware of Erigon, however we could experiment with it later.

My comments to you PR:
1. maybe you could cherry pick the archive config from my commit. My `exec_command_template` is a bit simpler than yours, please evaluate which version is more correct. I am also keeping the name `geth_linux` so that it is not that simple to confuse it with ETH geth. Also, could you use for your non archive version the port 9064, I am using 9065? I called the chain name simply BSC, it is shorter and easier to use. Does it make sense or is it a wrong name for the chain?
2. I used the same approach as you with `go-ethereum`, with a bit dirty trick to not to duplicate the initialization code code https://github.com/trezor/blockbook/blob/c518600e17708d4078f90651315d588b800577de/bchain/coins/bsc/bscrpc.go#L39. Maybe it could be rewritten so that it is not dirty and yet does not duplicate the code.
3. I intended to do exactly the same with the reinitialization of the `EthereumTokenTypeMap`, good work! However, I think the code in the `txdetail_ethereumtype.html` could be made a bit more generic - I would extend the `TemplateData` struct by 3 fields (FungibleToken, NonFungibleToken and MultiToken) and set them in `newTemplateData` to the values from the `EthereumTokenTypeMap`. Then in the `txdetail_ethereumtype.html` you could compare/display the values directly from these fields and get rid of the hardcoded names. Btw bscscan is using ERC721 and ERC1155 instead of BEP721 and BEP1155 but it is probably OK to use the latter.

kaladinlight:
Oh haha, nice! I was going to start polygon next, have you begun that as well by chance?

Good to know on the archive snapshot and that you are working on testing that side of things. I agree Erigon can be pushed off as a follow up addition/improvement.

1. I will cherry-pick your commit and update the `exec_command_template`. Yours works since you restored from the snapshot, but it is necessary to initialize the chain from the `genesis.json` on a fresh sync. I may bring over the `exec_script` just to simplify things as long as I'm in the area. Much easier to parse a shell script than a one liner command. I can update the ports if you like. I was just adding to the end, but can add into that available gap. I went back and forth a bit on the naming. I chose to follow the existing pattern `eth` package and `EthereumRpc`, but I feel like `Bsc` is well enough known and understood that we can use it to keep things more succinct if you'd like. No big preference either way and think both are plenty valid.
2. I personally don't mind the explicit initialization just to keep a consistent evm derivative pattern. Would also need to pull over more of the logic for testnet network switch if that is important to support anyways. That is my 2c, but I will defer to your call.
3. Cool! I am not super familiar with the templating engine, so that is good feedback. I def just shimmed it in there as I could but will explore this more and get myself more familiar. And I say we are one upping bscscan with proper BEP vernacular for the other tokens as well then then haha :P Easy enough to change if you'd like

martinboehm:
1. OK, name it `BNB Smart Chain`, just in the case of the archive version the `label` could be without the Archive suffix, I think it is not necessary to display in the explorer header that it is the archive version
2. sure, the explicit init is fine
3. let's keep BEP :)

martinboehm:
I am not working on polygon, it is all yours :D

kaladinlight:
Ok, review comments addressed. I did go ahead and add the `exec_script` logic, though I still haven't gotten a debian vm set up and haven't been able to fully test the flow. Manually executing all of the steps works as expected however.

kaladinlight:
Also, I am using `bnb` for `platformVsCurrency`, though you had `usd`. I haven't looked into how that is being used. Can you provide a quick explanation? Saw ethereum was using `eth` which is why I went with `bnb`, but not much more thought than that...

martinboehm:
Thanks, the template looks great! However, you should also update `static/templates/address.html` as there are also hardcoded token types.

As for the `platformVsCurrency`, using `bnb` is fine. In Avalanche you used `usd` as it is not in the supported vs currencies. Blockbook is now able to compute the token exchange rates from any currency, which is in the https://api.coingecko.com/api/v3/simple/supported_vs_currencies list.
In general it should be a currency against which we have the maximum possible history of exchange rates so I would not use some of the smaller fiat currencies from the list but otherwise it does not matter.

kaladinlight:
Good catch! I just pushed those changes. I am not on the best internet atm, so did as best of a spot check as I could and it was looking ok, but could use another check probably or can wait until I am back in better service.

Thanks for the rundown, that makes sense!

build/tools/templates.go
```diff
@@ -354,3 +359,23 @@ func writeBackendClientConfigFile(config *Config, outputDir string) error {
 	_, err = io.Copy(out, in)
 	return err
 }
+
+func writeBackendExecScript(config *Config, outputDir string) error {
+	out, err := os.OpenFile(filepath.Join(outputDir, "backend/exec.sh"), os.O_CREATE|os.O_WRONLY, 0777)
```
martinboehm, line 364: Just thinking, is it necessary to open the output file if the `config.Backend.ExecScript` is empty?

configs/coins/bsc_archive.json
```diff
@@ -0,0 +1,73 @@
...
+      "mempool_workers": 8,
+      "mempool_sub_workers": 2,
+      "block_addresses_to_keep": 600,
+      "additional_params": {
+        "address_aliases": true,
+        "mempoolTxTimeoutHours": 48,
+        "processInternalTransactions": true,
+        "queryBackendOnMempoolResync": false,
+        "fiat_rates": "coingecko",
+        "fiat_rates_vs_currencies": "AED,ARS,AUD,BDT,BHD,BMD,BRL,CAD,CHF,CLP,CNY,CZK,DKK,EUR,GBP,HKD,HUF,IDR,ILS,INR,JPY,KRW,KWD,LKR,MMK,MXN,MYR,NGN,NOK,NZD,PHP,PKR,PLN,RUB,SAR,SEK,SGD,THB,TRY,TWD,UAH,USD,VEF,VND,ZAR,BTC,ETH",
```
martinboehm, line 64: You removed the `fourByteSignatures` entry. Do you think the signatures do not work for BSC?

kaladinlight: Bad copy, good catch. Although I did notice there looks to be issues handling conflicting sighashes when resolving the name field. Seems difficult outside of best guess without having abi available...
image

kaladinlight: https://github.com/trezor/blockbook/pull/869/commits/b3da7aa10789a4fc596838a5267fe562938a693d

build/tools/templates.go
kaladinlight: So it looks like due to the way these backend files are renamed during [dh-exec-install](https://github.com/trezor/blockbook/blob/1f4398aa99c17b340744d75cd677e7e4260ca61d/build/templates/backend/debian/install) it is expected the files exist. I tried a couple different attempts at conditional rename without any luck. Let me know if you have an idea, otherwise going to leave it in line with existing conf files

configs/coins/bsc_archive.json
martinboehm: What method should it be? What tx is it?

build/tools/templates.go
martinboehm: Well, you could make https://github.com/trezor/blockbook/blob/1f4398aa99c17b340744d75cd677e7e4260ca61d/build/templates/backend/debian/install#L6 conditional :) I would actually prefer it as I do not want to have unnecessary files around if the exec script is not used.

martinboehm: To be more specific, include the line only if the exec script exists.

kaladinlight: I was attempting to make conditional with
```
[ -f exec.sh ] && exec.sh => {{.Env.BackendInstallPath}}/{{.Coin.Alias}}/{{.Coin.Alias}}_exec.sh
```
but it didn't seem happy with bash evaluation in the dh-exec script. I will fiddle with it a bit more

configs/coins/bsc_archive.json
kaladinlight: an example tx would be: https://bscscan.com/tx/0xcdc4a1495c6734af04abc1955fedff6837d09aa013e9d7fbb927be28b663cc23

so just a standard `transfer` method. it may just be related to the 4 byte storage as I am looking. I haven't had long term syncs during dev testing, so possible I just haven't stored that sighash yet from the 4bytes directory.

build/tools/templates.go
martinboehm: Wha I mean is that the file `build/templates/backend/debian/install` is a go template - you could put there something like
```
{{if .Backend.ExecScript}}exec.sh => {{.Env.BackendInstallPath}}/{{.Coin.Alias}}/{{.Coin.Alias}}_exec.sh{{end}}
```

configs/coins/bsc_archive.json
martinboehm: It seems plausible that your 4 byte directory is not yet synced as the data look like a standard transfer.

build/tools/templates.go
kaladinlight: doh :man_facepalming: yeah that totally works haha

configs/coins/bsc_archive.json
kaladinlight: For posterity, it finally populated:
image

kaladinlight:
Let me know if there are any other changes you would like me to make, but I think it is in a good state for merge at this point :) I will rebase optimism after we get this in with the exec script updates

martinboehm:
Could you please add `bsc_archive` integration test? I think it can just a copy of the non archive version. I cannot run the non archive version as the ports of my archive backend do not match what the test expects.

Otherwise the PR looks good. 
I am already running Blockbook from this PR - it is syncing (I expect it to be syncing for about 10 days more). After that I will verify the functionality of the synced Blockbook and install the backend to check the exec script functionality.

kaladinlight:
Just pushed up those copy archive tests for you.

That's great to hear! Let me know if anything comes up with those couple remaining verification steps :)

configs/coins/bsc_archive.json
```diff
@@ -0,0 +1,74 @@
...
+    "package_revision": "satoshilabs-1",
+    "system_user": "bsc",
+    "version": "1.1.19",
+    "binary_url": "https://github.com/bnb-chain/bsc/releases/download/v1.1.19/geth_linux",
+    "verification_type": "sha256",
+    "verification_source": "ad49b00a5df44fb67e0dc166a7cf02a02c67ab296d25e4ac3d56134567577c41",
+    "extract_command": "mv ${ARCHIVE} backend/geth_linux && chmod +x backend/geth_linux && echo",
+    "exclude_files": [],
+    "exec_command_template": "/bin/sh -c '{{.Env.BackendInstallPath}}/{{.Coin.Alias}}/bsc_exec.sh 2>> {{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend/{{.Coin.Alias}}.log'",
+    "exec_script": "bsc.sh",
```
martinboehm, line 30: Hi, I just built the archive backend. Unfortunately the config seems to me a bit messed up, fortunately in a way that the backend did not start the non archive version :D 

When starting the service I got the error `/bin/sh: 1: /opt/coins/nodes/bsc_archive/bsc_exec.sh: not found`. In the folder `/opt/coins/nodes/bsc_archive/` there is file `bsc_archive_exec.sh`, which contains the non archive script :D

I think the `exec_command_template` should start `bsc_exec_archive.sh` instead of `bsc_exec.sh` and the `exec_script` should be `bsc_archive.sh` instead of `bsc.sh`.

martinboehm: Yep, this works:
```
    "exec_command_template": "/bin/sh -c '{{.Env.BackendInstallPath}}/{{.Coin.Alias}}/bsc_archive_exec.sh 2>> {{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend/{{.Coin.Alias}}.log'",
    "exec_script": "bsc_archive.sh",
```

kaladinlight: Man I really botched the copy of the archive config... Sorry for all the issues there. Just pushed up that change

martinboehm: Don't worry, it just seemed funny to me that two independent bugs averted potentially larger issue for me (I do not know what would happen if I started the non archive backend to my archive DB).
And in general, the exec script functionality works really well, good work!

kaladinlight: Ha yeah for real. Glad your archive db remained safe, that would have sucked.. Glad the exec scripts are working, and I assume the post install worked even though you aren't using the genesis file since you restored from a snapshot?

martinboehm:
@kaladinlight Hi, my archive Blockbook is finally synchronized. However, I see a small problem - it seems to me that the subscribe to `newHeads` is not really working. While BSC is getting a new block every 3 seconds (as I see in the backend log), Blockbook remains stuck on one block for quite some time (minutes) before fetching the new head and synchronizing and then being stuck again. Could you please verify it?

kaladinlight:
@martinboehm I will look into this today and see if I can track down the issue.

Out of curiosity what kind of sync speeds were you seeing for archive sync against bsc? Was it a full archive sync starting at block 0 (I can't remember if you mentioned)?

kaladinlight:
Right yeah that makes sense. In regards to the debug trace, did you compare performance between trace block vs trace transaction by chance? I am wondering if there could be any performance squeeze by concurrently tracing transactions when ingesting blocks. Feels like nodes are good at handling high request throughput versus the more time/compute intensive requests like debug block. I even tried increasing the worker count and ended up getting a lot of context deadline timeouts due to it overloading the node with those heavy requests

kaladinlight:
Just pushed some changes to the node config:

- use `config.toml` provided in release. this is required for proper gas price oracle configuration (accurate gas prices) if using node for estimating and broadcasting transactions. also specifies bootnodes and static nodes for peering and configures logging to file `bsc.log`
- log redirect removed since redirecting via node directly
- `maxpeers` bumped to 100. this appears to make peers more stable and keep the chain in sync and broadcasting new heads as expected. if there are still issues surrounding this, it sounds like the next update would be overwriting `StaticNodes` in the `config.toml` with the peers provided by https://api.binance.org/v1/discovery/peers.

Let me know if this stabalizes your `newHeads` situation or if there is still something else going on. I am not seeing any issues on the blockbook side of things though based on my testing today

martinboehm:
Your config changes fixed the issue!

It is running fine now. I observed some minor issues with a blockchain with so many blocks/minute:
- the status now shows a lot that it is out of sync (which it is as it is almost all the time syncing). But this is misleading as it is actually syncing fine. I think it will be necessary to use couple of seconds delay before setting the status to out of sync
- Blockbook now computes db directory size after each mined block, which is fine for BTC with block every 10 minutes but is ridiculous for BSC. This has to be modified.

However these problems are not to be resolved in this PR. I will let it run for a day or so and if there are no further problems, I will merge this PR. Really great work 👍, thank you!

I find it hard to believe that individual transactions can be much faster. It is now able to process about 5000 blocks/hour and with about 100 transactions/block on average it is about 140 transactions parsed/second. There would be some network overhead for each transaction and I do not have a server with hundreds of CPUs so parallelism would also not help.

kaladinlight:
Awesome! Happy to hear that resolved the issue and things are looking solid at this point. Both of those follow up points make sense to me.

Thanks for another solid review and working through this integration with me! Optimism is ready if you wanted to start on that one next? I have polygon up, but need to verify the build a bit more still and am seeing some sync performance issues I need to diagnose.

martinboehm:
What is the expected size of the Optimism backend/blockbook db, so that I know what to expect?

kaladinlight:
l2geth: ~200gb
data-transport-layer: ~100gb
blockbook: ~20gb (note I am still running a legacy fork here, so footprint will be a bit bigger on v0.4.0)

I am going to start a v0.4.0 sync now that I am talking about it so I can get you a more accurate size. Full node, not archive btw.

Also remember to point to a free tier alchemy node as described here to avoid getting rate limited: https://github.com/trezor/blockbook/pull/868#discussion_r1107766602

martin-key:
Hey! Have you tried doing a full sync of the chain? I get some weird error messages and it could be just a local problem. @martinboehm I have some nodes for internal use but I can share you a link if you need them for testing. Cheers!

martinboehm:
@martin-key Hi, yes, I have fully synced Blockbook. My backend runs in archive mode now, however contains blocks with full archive info only from December 2022 (I restored it from the snapshot) and Blockbook has internal transactions starting this time (I disabled fetching of internal transactions while it was syncing). I do not see any weird error messages apart from very occasional timeout of getting debug info.

kaladinlight:
> Hey! Have you tried doing a full sync of the chain? I get some weird error messages and it could be just a local problem. @martinboehm I have some nodes for internal use but I can share you a link if you need them for testing. Cheers!

Can you provide any of the error messages your are seeing? Are these related to the bsc node or blockbook sync specifically?

martinboehm:
@kaladinlight Hi, I think the `logrotate_files_template` is not correct, it still expects the previous way of logging using stdout redirection. I actually preferred the old way a bit more. But either way, the log files should be archived and rotated.

Now:
```
lrwxrwxrwx     57 bitcoin  7 Mar  8:00  bsc.log -> /opt/coins/data/bsc_archive/backend/bsc.log.2023-03-07_08
.rw-r--r--   1.7M bitcoin 28 Feb  0:59  bsc.log.2023-02-28_00
.rw-r--r--   4.0M bitcoin 28 Feb  1:59  bsc.log.2023-02-28_01
.rw-r--r--   3.0M bitcoin 28 Feb  2:59  bsc.log.2023-02-28_02
...
.rw-r--r--   2.1M bitcoin  7 Mar  7:59  bsc.log.2023-03-07_07
.rw-r--r--   902k bitcoin  7 Mar  8:36  bsc.log.2023-03-07_08
```
The old way:
```
.rw-r--r--      0 bitcoin  1 Mar  0:00  bsc_archive.log
.rw-r--r--   119k bitcoin 28 Feb  0:33  bsc_archive.log.1.gz
.rw-r--r--   9.8M bitcoin 27 Feb 23:59  bsc_archive.log.2.gz
.rw-r--r--    14M bitcoin 26 Feb 23:59  bsc_archive.log.3.gz
```

kaladinlight:
Ok, I will take a look. Only reason I switched was so we could use the included `config.toml` which provides config not supported by cli flags, but it also set log file config. If it takes too much time to sort out, I am just going to sed replace the log config as I did in my own node setup, was just trying to avoid that in the blockbook infra.

I am also tracking down a sync issue I am seeing on bsc:
```bash
getBlockWorker 2 connect block error eth_getLogs blockNumber 0xf006c6: websocket: read limit exceeded. Retrying...
```

I am running into this on multiple syncs at the same block. I have confirmed the log payload looks to be 17mb when the go-ethereum websocket rpc client has a 15mb read limit set that isn't configurable resulting in the error. I am confused how your sync didn't seem to run into this same issue though... I have been able to replicate locally. Any chance you can try to replicate against your running node just for a quick sanity check?

My thought is to add the concept of an overloaded http rpc client that can be used in place of the websocket rpc client for verbose payload calls as http doesn't have the same read limit (or just straight up swap `eth_getLogs` with http client, no overload/config). I don't want to go the route of hand rolling a websocket client just for configurable read limit, or it looks like there may be an unsafe way to manipulate private fields, but that just sounds like a bad idea. Let me know if you have any thoughts on the matter, thanks!

martinboehm:
Logs: maybe just the correct setup of the `logrotate_files_template` will fix the log rotation. I haven't had time to try it though.

Payload limit: I fixed it long time ago using much more hacky way :-) You are not building Blockbook using the docker build, right? Because I have hit this limit in Ropsten and Goerli and did the following:
https://github.com/trezor/blockbook/blob/1bcdcc6a0a29eb839a9b08cb51653bc68edf2a4c/build/docker/bin/Makefile#L41.
I actually hit this limit twice, once I created a go-ethereum PR which fixed the issue (the limit was common for rest and ws requests and much lower than 15MB), however they rejected it and fixed it a bit differently, IMHO less flexibly. And the 2nd time I did not bother :D

kaladinlight:
> Logs: maybe just the correct setup of the `logrotate_files_template` will fix the log rotation. I haven't had time to try it though.
> 
> Payload limit: I fixed it long time ago using much more hacky way :-) You are not building Blockbook using the docker build, right? Because I have hit this limit in Ropsten and Goerli and did the following:
> 
> https://github.com/trezor/blockbook/blob/1bcdcc6a0a29eb839a9b08cb51653bc68edf2a4c/build/docker/bin/Makefile#L41
> 
> .
> I actually hit this limit twice, once I created a go-ethereum PR which fixed the issue (the limit was common for rest and ws requests and much lower than 15MB), however they rejected it and fixed it a bit differently, IMHO less flexibly. And the 2nd time I did not bother :D

Haha, that is a cool hack I didn't think of. And no I have a bit of a custom setup and never saw that sed replace which explains why you didn't run into it and I am having issues.

I did add: https://github.com/trezor/blockbook/pull/869/commits/a50db41ae417a2faed383de19ee87785d84a917e as a fix if we want to keep it? there is some inferrence based on `rpc_url` and additional `http_port`, so I will leave it up to you to keep or just continue relying on sed hack (which I will update my setup to include)

martinboehm:
I would prefer to not to use https://github.com/trezor/blockbook/commit/a50db41ae417a2faed383de19ee87785d84a917e. It fixes only the `eth_getLogs` case but the overflow could happen in other requests too. For us the hacky fix works better.

kaladinlight:
That sounds fair. I am wondering if it would be worth adding same hack for `github.com/ava-labs/coreth` just in case? Or suppose we can wait if/when there is an occurrence and add at that point.

Dropping https://github.com/trezor/blockbook/commit/a50db41ae417a2faed383de19ee87785d84a917e

I update the log template, but haven't tested it myself yet.

martinboehm:
Thank you! I will test the log rotation right now.

martinboehm:
Thank you!

martinboehm:
@kaladinlight Hi, I am not really happy with the way bsc logs are rotated now. Example:
```
.rw-r--r--     0 bitcoin 14 Mar  0:00  bsc.log.2023-03-12_23
.rw-r--r--     0 bitcoin 16 Mar  0:00  bsc.log.2023-03-12_23.1.gz
.rw-r--r--  242k bitcoin 12 Mar 23:59  bsc.log.2023-03-12_23.1.gz.1.gz
```
I would probably prefer to revert to the previous, redirected logs. I tried to use sed to disable it in this commit https://github.com/trezor/blockbook/commit/e8be05a1ec8bc802bcc56a2e23d404d4c24feffa, unfortunately bsc is complaining that it does not know the disabled key and does not start.

kaladinlight:
> @kaladinlight Hi, I am not really happy with the way bsc logs are rotated now. Example:
> 
> ```
> .rw-r--r--     0 bitcoin 14 Mar  0:00  bsc.log.2023-03-12_23
> .rw-r--r--     0 bitcoin 16 Mar  0:00  bsc.log.2023-03-12_23.1.gz
> .rw-r--r--  242k bitcoin 12 Mar 23:59  bsc.log.2023-03-12_23.1.gz.1.gz
> ```
> 
> I would probably prefer to revert to the previous, redirected logs. I tried to use sed to disable it in this commit [e8be05a](https://github.com/trezor/blockbook/commit/e8be05a1ec8bc802bcc56a2e23d404d4c24feffa), unfortunately bsc is complaining that it does not know the disabled key and does not start.

Apologies for that, here is the sed I am using locally:
```sh
sed -i -e '/\[Node.LogConfig\]/,+5d' config.toml
```

Would you like me to PR the fix, or want to just try to swap that out? I am not aware of a specific `disable` syntax for a key (although would think it just wouldn't parse it...)

martinboehm:
@kaladinlight  Thank you, that worked perfectly! I got a bit lazy to search the right sed magic to remove lines. The change is already in master.

And additional thing: how about checking the need to do init chain using the presence of the folder `geth/lightchaindata` instead of `geth/chaindata`. I was setting up the backend from the ancients and it did not get initialized as the folder `geth/chaindata` was present (but containing only `geth/chaindata/ancients`) and the init was still necessary.

kaladinlight:
Glad that sed worked, I know that feel with sed magic haha.

And yeah, might as well check that both `chaindata` and `lightchaindata` exist. Wasn't really familiar with that setup, so didn't even think about it.

martinboehm:
Hi, FYI, I am running into issues described here https://github.com/bnb-chain/bsc/issues/1289, I am trying to increase the number of peers to 200 as suggested in the issue.

kaladinlight:
> @kaladinlight Hi, FYI, I am running into issues described here [bnb-chain/bsc#1289](https://github.com/bnb-chain/bsc/issues/1289), I am trying to increase the number of peers to 200 as suggested in the issue.

Hey! I have been having sync stability issues pretty constantly thus far. I have tried the static nodes provided by binance as well. Going to give `maxpeers=200`another shot along with you. Let me know if that resolves or helps things on your end 🤞

Issue #882, opened 2023-03-03
blst error - question about development build


GBBx:
Hi,

I've been using Blockbook for a long time.
I used to be able to follow the [manual build](https://github.com/trezor/blockbook/blob/master/docs/build.md#manual-build) steps but I run into an issue since I updated to 0.4.0.
When I start blockbook I get this error message:
```
Caught SIGILL in blst_cgo_init, consult <blst>/bindings/go/README.md.
```

I found this [discussion](https://github.com/community/community/discussions/43153). They suggest to add `CGO_CFLAGS=-O -D__BLST_PORTABLE__`, so I tried this:
```
CFLAGS="-fPIC" CXXFLAGS="-fPIC" CGO_CFLAGS="-O -D__BLST_PORTABLE__" make release
```
when building rocksdb but it did not solve the issue.

I also tried to set `CGO_CFLAGS="-O -D__BLST_PORTABLE__" make release` globally but that failed, too.

I undertand `blst` is a dependency of a dependency of Blockbook but are you maybe familiar with this problem?

martinboehm:
Hi, sorry, we do not have this build problem, neither in the manual build nor in the docker build. What version of go are you using? As it seems related to Avalanche dependencies, maybe @kaladinlight will have an idea.

GBBx:
Hi @martinboehm , thanks for coming back to me.
I try to build it with Go 1.19.2 on Ubuntu 20.04 (I also tried Ubuntu 22.04 - didn't help).

kaladinlight:
I have an i7 and not seeing any issues. I am using `go version go1.20 linux/amd64` locally if you want to give that a try? There was an additional `-lzstd` LDFLAG added with v0.4.0 as well just to make sure everything matches up.

GBBx:
hi @kaladinlight, thanks for the hints.
I've tried:
- setting CGO_CFLAGS_ALLOW
- setting CGO_CFLAGS before building rocksdb, and after
- updating kernel
- adding -lzstd
All failed.

@XK4MiLX's finding is interesting. There are some comments about architecture at the [bottom of the page](https://github.com/supranational/blst/tree/master/bindings/go#build) I haven't tried yet.

kaladinlight:
I would be curious about building and testing the blst go binding directly to reduce the scope of the problem a bit. https://github.com/supranational/blst/tree/master/bindings/go. According to https://github.com/supranational/blst#platform-and-language-compatibility it appears to have extensive os/platform support...

[...]

PR #883, opened 2023-03-06
fix: avalanche blst dependency issue

kaladinlight:
- use base coreth rpc client over info client to reduce additional dependencies which also caused blst build errors as described: https://github.com/trezor/blockbook/issues/882

image


martinboehm:
@kaladinlight Thank you! That removed a lot of dependencies, great! Were you able to run this version? I do not have a running Avalanche at this moment, I cannot check that it works.

kaladinlight:
Yeah, crazy how much that brought in. I will keep an eye on that moving forward... The screen shot above is local run with this change. I can build the image and verify the build as well if you would like.

PR #905, opened 2023-03-30
Use sync.Map to improve EVM sync performance

kaladinlight:
Sync speed issues have become really apparent when syncing Polygon where 1k block chunks can take upwards of 3 hours.

Golang pprof exposed the bottleneck in blockbook related to `big.Int.Cmp()` in `addToContract` as shown here:
image

To remedy this problem, I switched to use `sync.Map` giving a O(1) lookup instead of the O(n) `big.Int.Cmp()`. Maps are not thread safe which is why I chose `sync.Map` over `map` and managing our own mutex lock, etc.

The biggest downside to `sync.Map` is the lack of `len()` support and key/val being typed as `any` which requires some extra type casts and loops. There is also the conversion of the map key from `string` to `big.Int` back and forth.

I have validated the NFT/MultiTokens are being updated as expected, but could use a second set of eyes just to make sure I am not missing something.

I also left in the error logs on all of the casts for posterity, but realistically there shouldn't be a problem there since we are controlling the types on all sides. It would catch if a type is updated and not propagated through correctly however.

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -957,20 +957,47 @@ func (w *Worker) getEthereumContractBalance(addrDesc bchain.AddressDescriptor, i
 				}
 			}
 		} else {
-			if len(c.Ids) > 0 {
-				ids := make([]Amount, len(c.Ids))
-				for j := range ids {
-					ids[j] = (Amount)(c.Ids[j])
+			ids := []Amount{}
+			c.Ids.Range(func(key, val any) bool {
+				k, ok := key.(string)
+				if !ok {
+					glog.Errorf("invalid nft id: %v (type: %T)", key, k)
+					return true
 				}
+				id, ok := new(big.Int).SetString(k, 10)
+				if !ok {
+					glog.Errorf("failed to convert nft id: %s", k)
+					return true
+				}
+				ids = append(ids, (Amount)(*id))
+				return true
+			})
+			if len(ids) > 0 {
 				t.Ids = ids
 			}
-			if len(c.MultiTokenValues) > 0 {
-				idValues := make([]MultiTokenValue, len(c.MultiTokenValues))
-				for j := range idValues {
-					idValues[j].Id = (*Amount)(&c.MultiTokenValues[j].Id)
-					idValues[j].Value = (*Amount)(&c.MultiTokenValues[j].Value)
+
+			multiTokenValues := []MultiTokenValue{}
+			c.MultiTokenValues.Range(func(key, val any) bool {
+				k, ok := key.(string)
+				if !ok {
+					glog.Errorf("invalid multi token id: %v (%T)", key, key)
+					return true
 				}
-				t.MultiTokenValues = idValues
+				id, ok := new(big.Int).SetString(k, 10)
+				if !ok {
+					glog.Errorf("failed to convert multi token id: %s", k)
+					return true
+				}
+				v, ok := val.(*big.Int)
+				if !ok {
+					glog.Errorf("invalid multi token value: %v (%T)", val, val)
+					return true
+				}
+				multiTokenValues = append(multiTokenValues, MultiTokenValue{Id: (*Amount)(id), Value: (*Amount)(v)})
+				return true
+			})
+			if len(multiTokenValues) > 0 {
+				t.MultiTokenValues = multiTokenValues
 			}
 		}
 	}
--- a/db/rocksdb_ethereumtype.go
+++ b/db/rocksdb_ethereumtype.go
@@ -22,9 +22,9 @@ type AddrContract struct {
 	Type             bchain.TokenType
 	Contract         bchain.AddressDescriptor
 	Txs              uint
-	Value            big.Int                  // single value of ERC20
-	Ids              []big.Int                // multiple ERC721 tokens
-	MultiTokenValues []bchain.MultiTokenValue // multiple ERC1155 tokens
+	Value            big.Int  // single value of ERC20
+	Ids              sync.Map // multiple ERC721 tokens
+	MultiTokenValues sync.Map // multiple ERC1155 tokens
 }
 
 // AddrContracts contains number of transactions and contracts for an address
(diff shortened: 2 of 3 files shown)


db/rocksdb_ethereumtype.go
```diff
@@ -53,19 +53,58 @@ func packAddrContracts(acs *AddrContracts) []byte {
 			l = packBigint(&ac.Value, varBuf)
 			buf = append(buf, varBuf[:l]...)
 		} else if ac.Type == bchain.NonFungibleToken {
-			l = packVaruint(uint(len(ac.Ids)), varBuf)
+			i := 0
+			ids := []*big.Int{}
+			ac.Ids.Range(func(key, val any) bool {
```
kaladinlight, line 58: Could probably reduce to single loop by skipping the length byte, then adding back to whatever index it would have been at since we have to increment `i` during `Range` for length. Figured, keep it simple to avoid possibly messing something up..

db/rocksdb_ethereumtype.go
```diff
@@ -53,19 +53,58 @@ func packAddrContracts(acs *AddrContracts) []byte {
...
+				k, ok := key.(string)
+				if !ok {
+					glog.Errorf("invalid nft id: %v (type: %T)", key, k)
```
kaladinlight, line 61: to reiterate, since we know the types we are dealing with throughout the pipeline, we could shortcut the error handling to make the code less verbose, but figured the safety net to catch unintended later changes might be beneficial. easy to get lost in this area of things (at least for me haha)

db/rocksdb_ethereumtype.go
```diff
@@ -100,24 +139,21 @@ func unpackAddrContracts(buf []byte, addrDesc bchain.AddressDescriptor) (*AddrCo
 			buf = buf[ll:]
 			ac.Value = b
 		} else {
-			len, ll := unpackVaruint(buf)
+			length, ll := unpackVaruint(buf)
```
kaladinlight, line 142: update to avoid shadowing `len`

api/worker.go
```diff
@@ -957,20 +957,47 @@ func (w *Worker) getEthereumContractBalance(addrDesc bchain.AddressDescriptor, i
...
 				t.Ids = ids
 			}
-			if len(c.MultiTokenValues) > 0 {
-				idValues := make([]MultiTokenValue, len(c.MultiTokenValues))
-				for j := range idValues {
-					idValues[j].Id = (*Amount)(&c.MultiTokenValues[j].Id)
-					idValues[j].Value = (*Amount)(&c.MultiTokenValues[j].Value)
+
+			multiTokenValues := []MultiTokenValue{}
+			c.MultiTokenValues.Range(func(key, val any) bool {
```
kaladinlight, line 980: I am noticing some multi token contracts are marked as invalid (proxy contract it looks like at first glance), resulting in no tokens being displayed. seems like this can be handled differently to just show contract address (no additional contract details since we don't have), and the multi tokens. Separate issue though, so raising here before I forget.

martinboehm:
Interesting. Why do we not see the bottleneck in other EVM chains? Is Polygon so NFT heavy?

kaladinlight:
I was wondering the same thing and I have to assume that is the case. Not only NFTs but also multi token traffic as well?

Even though polygon was the extreme case, this change should still improve all evm chain sync speed to some extent. I have not spent the time fully profiling all of the evm chains to measure the difference though.

martinboehm:
My assumption was that there will be only a few ids per address and contract (which holds at least for ETH), in which case the simple O(n) algorithm can outperform the more complicated map approach. However it is obviously not true at least in case of Polygon, you approach to fix it is correct. Thank you!

martinboehm:
The unit tests `go test -tags=unittest ./...` are failing
```
db/rocksdb_ethereumtype_test.go:768:12: cannot use []big.Int{…} (value of type []big.Int) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:780:25: cannot use []bchain.MultiTokenValue{…} (value of type []bchain.MultiTokenValue) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:897:17: cannot use []big.Int{…} (value of type []big.Int) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:926:17: cannot use []big.Int{…} (value of type []big.Int) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:955:17: cannot use []big.Int{…} (value of type []big.Int) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:989:17: cannot use []big.Int{…} (value of type []big.Int) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:995:25: cannot use []bchain.MultiTokenValue{…} (value of type []bchain.MultiTokenValue) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:1038:17: cannot use []big.Int{…} (value of type []big.Int) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:1044:25: cannot use []bchain.MultiTokenValue{…} (value of type []bchain.MultiTokenValue) as type sync.Map in struct literal
db/rocksdb_ethereumtype_test.go:1091:17: cannot use []big.Int{…} (value of type []big.Int) as type sync.Map in struct literal
```
The initialization of the test data must be adapted to the sync.Map. 
Would you fix the tests please?

kaladinlight:
Oh yeah, sure thing, I will get to that in the morning!

martinboehm:
There is also an additional issue, the order of ids is undefined. I am not decided yet if it makes sense to sort the ids in the worker.go as the sort operation would be quite costly. 
You may have the same issue in the unit tests when comparing the expected value to the received value.

kaladinlight:
> There is also an additional issue, the order of ids is undefined. I am not decided yet if it makes sense to sort the ids in the worker.go as the sort operation would be quite costly. You may have the same issue in the unit tests when comparing the expected value to the received value.

Yeah, the tests have been more difficult than expected due to `sync.Map` being a memory address reference so `DeepEqual` doesn't work.

I am seeing the order issue as well. I was thinking initially that `sync.Map` kept track of order based on store, but I think that was just wishful thinking. An alternative to sorting would be keeping an associated ordered array of keys. This would actually probably make the length checks and ranging more straight forward as well while keeping order in tact. What do you think of that strategy?

kaladinlight:
Eh, that would pretty much undo the performance improvements having to manage the array again.... Putting some more thought into it.

martinboehm:
Thank you for the fix of the test, it was a pain to do looking at the code :)
Sorting only in the worker would not affect the speed of sync and would slow down only the client responses a bit. I think it is the only reasonable way how to do it but I am still not sure if it is worth it.

kaladinlight:
Haha yeah... I was not expecting that. Yeah I am not sure if the sort is worth and just leave it up to a client to decide if they care about sorting and do it accordingly for their use case. Easy enough to add if needed later on if there is a good reason or enough chatter over it, but I think response times would take precedence in this case imo.

The only annoyance will be the unit tests would need another update to handle the unordered nature unless you are ok with intermittent fails which I doubt would be the case haha

kaladinlight:
I am not coming up with any other reasonable way to handle it either...

martinboehm:
Of course, there is another potential possibility, which is to keep the O(n) algorithm but use some hacky way of comparing the big.Ints as we are interested in equality and not which one is bigger (like comparing only the zeroth element of the `abs` value - i.e. the least significant digits and only if they are the same do the full `Cmp` or possibly `CmpAbs`, which does not do the sign check). 
```
type nat []Word
type Int struct {
	neg bool // sign
	abs nat  // absolute value of the integer
}
```
But I am not sure if it would really make the difference.

kaladinlight:
I was even thinking of how much of a difference a cast + string comparison vs big int cmp would be, but as you are implying I feel like while there may be some marginal improvements with a more clevery/hacky comparison, it will still be problematic.

I can try to profile things more to compare, but it might take a little bit.

martinboehm:
It depends on how many ids has a typical address in Polygon. The Map approach has fairly big overhead (converting to string and the hashmap algorithm itself). If it is in tens then it can work, if it is in hundreds or thousands then O(n) is hopeless.

kaladinlight:
There are a handful of the top tokens with thousands and with thousands of transactions over the past 24hrs, which would totally destroy things. I am also curious if/when bull market comes back and more use of nft/multi-tokens if ethereum or the other evm chains would feel it as well

martinboehm:
Ha, I have just realized what is probably the best approach (Complexity Theory 101 :) )

Lets get rid of the maps and use always sorted arrays of ids and look up the ids using binary search. That will be O(log n) both for removal and insertion (we will have to always insert to the right place). As a bonus it will keep the arrays sorted even for the client.

The old algorithm has O(1) for insertion - just append and O(n) for removal but unfortunately does not guarantee sorted order, only constant order. Which will mean there will have to be a migration which will have to sort all stored addresseContracts, however it will have to run only once.

The map algorithm has O(1)-O(log n) both for insertion and deletion, however a huge overhead of string conversions (with memory allocations etc.) and especially for small number of ids is suboptimal. And does not guarantee any order.

What do you think?

kaladinlight:
Big brains, I like it! I will rework it with that strategy and report back here with the findings in the morning. Thanks for the brainstorming help!

kaladinlight:
Have been busy and haven't had a chance to get around to this obviously, but hoping to pick it back up start of next week just as an update

martinboehm:
No problem. Maybe keep this PR for a reference and create a new PR. 
Also, I could give you a hint with the migration or do it myself if it would help.

kaladinlight:
Yeah if you wouldn't mind pointing me in the right direction, that would be helpful. Happy to learn how to do it.

kaladinlight:
Ok, think I have the migration figured out. Testing to make sure it is working as intended and then will profile the new binary search performance to make sure we are in a good place

martinboehm:
Sorry if we both did the same thing, we can compare who has done better :)

kaladinlight:
Haha all good. Largely the same, though I did it as a db version migration, didn't see the more feature specific migration flags. Also didn't do any of the swap logic as I was thinking ids should be unique for contract anyways:

https://github.com/kaladinlight/blockbook/commit/45280220e76b7b595916a0eaa72af0905ce9d496

kaladinlight:
With all of this though and doing the binary search insert/update/delete, while it is faster than the previous range implementation, things are still taking way too long... I feel like I was seeing in the 3-5min for range, ~40s for binary, <250ms with sync map.

image

kaladinlight:
Update: I may have been seeing slower times due to mismanagement of the array refs resulting in unnecessary copies of data. I am going to update my db migration and retest against my db state to re validate. Stay tuned...

PR #910, opened 2023-04-11
perf: leverage binary search for managing for Ids and MultiTokenValues

kaladinlight:
Closing: https://github.com/trezor/blockbook/pull/905 in favor of this to reduce overhead and unsorted nature of `sync.Map`

### DIFF
--- a/db/rocksdb_ethereumtype.go
+++ b/db/rocksdb_ethereumtype.go
@@ -38,8 +38,7 @@ func (s *Ids) sort() bool {
 func (s *Ids) search(id big.Int) int {
 	// attempt to find id using a binary search
 	return sort.Search(len(*s), func(i int) bool {
-		cmp := (*s)[i].CmpAbs(&id)
-		return cmp == 1 || cmp == 0
+		return (*s)[i].CmpAbs(&id) >= 0
 	})
 }
 
@@ -79,8 +78,7 @@ func (s *MultiTokenValues) sort() bool {
 // search for multi token value using a binary seach on id
 func (s *MultiTokenValues) search(m bchain.MultiTokenValue) int {
 	return sort.Search(len(*s), func(i int) bool {
-		cmp := (*s)[i].Id.CmpAbs(&m.Id)
-		return cmp == 1 || cmp == 0
+		return (*s)[i].Id.CmpAbs(&m.Id) >= 0
 	})
 }
 


blockbook.go
```diff
@@ -196,6 +196,17 @@ func mainWithExitCode() int {
...
 	}
+
+	// sort addressContracts if necessary
+	if !internalState.SortedAddressContracts {
+		err = index.SortAddressContracts(chanOsSignal)
+		if err != nil {
+			glog.Error("sortAddressContracts: ", err)
+			return exitCodeFatal
+		}
+		internalState.SortedAddressContracts = true
```
kaladinlight, line 207: you had a comment about sync here that I wasn't exactly sure what you meant if you want to provide additional detail

db/rocksdb_ethereumtype.go
```diff
@@ -1202,3 +1267,56 @@ func (d *RocksDB) DisconnectBlockRangeEthereumType(lower uint32, higher uint32)
...
+		addrDesc := it.Key().Data()
+		buf := it.Value().Data()
+		if len(buf) > 0 {
+			ca, err := unpackAddrContracts(buf, addrDesc)
+			if err != nil {
+				glog.Error("failed to unpack AddrContracts for: ", hex.EncodeToString(addrDesc))
+			}
+			for i := range ca.Contracts {
+				c := &ca.Contracts[i]
+				if sorted := c.Ids.sort(); sorted {
```
kaladinlight, line 1306: went ahead and added the swap logic to track updates

blockbook.go
martinboehm: I did not want to set the migration as complete until the sorting is implemented in the `addToContract` function. If I just sort and then run the old version, new data will be inserted in wrong order...

db/rocksdb_ethereumtype.go
```diff
@@ -17,14 +19,103 @@ import (
...
+		}
+		return isLess
+	})
+	return sorted
+}
+
+func (s *Ids) search(id big.Int) int {
+	// attempt to find id using a binary search
+	return sort.Search(len(*s), func(i int) bool {
+		cmp := (*s)[i].Cmp(&id)
```
martinboehm, line 41: How about using `CmpAbs`. I do not think there are any negative ids and `CmpAbs` is slightly more performant.

blockbook.go
kaladinlight: Oh right, makes sense thanks.

db/rocksdb_ethereumtype.go
```diff
@@ -1202,3 +1267,56 @@ func (d *RocksDB) DisconnectBlockRangeEthereumType(lower uint32, higher uint32)
...
+				}
+				if sorted := c.MultiTokenValues.sort(); sorted {
+					multiTokenValuesSortedCount++
+				}
+			}
+			buf := packAddrContracts(ca)
+			wb.PutCF(d.cfh[cfAddressContracts], addrDesc, buf)
+		}
+	}
+	if err := d.WriteBatch(wb); err != nil {
```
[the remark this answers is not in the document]
martinboehm: I had issues with resources consumption with a huge iteration. However, it was a long time ago and maybe by now the RocksDB resolved the issue. On the other hand, it logs progress, which is nice when it takes quite some time. Also, I was updating only rows which were resorted (and row after row). I am not really sure if it is a good idea to create a single batch which updates 200M rows.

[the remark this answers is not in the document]
martinboehm: Sorry, I was answering a question you did not ask :) It was about my way of splitting the iteration to multiple smaller parts.

martinboehm:
Is the performance now acceptable? I would hope so as I would expect that converting `big.Int` to decimal (or you used a bit more performant hexadecimal, I do not recall?) string must be at least as expensive as 5-10 `CmpAbs` operations, not taking account additional  Map overhead. The O(log n) algorithm should in 5-10 steps cover fairly large arrays of up to 1000 elements.
But as I did not measure it, these are only my assumptions...

kaladinlight:
Performance is acceptable now at ~1m-1.5m per 1k blocks which is pretty standard historical if I recall correctly. The `sync.Map` still outperforms even with the extra overhead required at ~30s per 1k blocks, but this seems like the right approach all things considered.

martinboehm:
ETH mainnet has this stats:
```
{
                "name": "addressContracts",
                "version": 6,
                "rows": 279932025,
                "keyBytes": 5598640500,
                "valueBytes": 9853527243,
}
```
That is why I think it is not a great idea to update it in one huge batch.

kaladinlight:
Ok, write batch updated per `AddrContracts` and progress logging every 100k rows https://github.com/trezor/blockbook/pull/910/commits/47e16e2fb3c997160554f5dbafa484eff46bbcd2

db/rocksdb_ethereumtype.go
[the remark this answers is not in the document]
kaladinlight: https://github.com/trezor/blockbook/pull/910/commits/47e16e2fb3c997160554f5dbafa484eff46bbcd2

martinboehm:
BTW here are the stats from my version of sorting of Goerli:
```
I0410 22:52:55.512316   10845 rocksdb_ethereumtype.go:1246] SortAddressContracts: starting
I0410 22:53:01.101915   10845 rocksdb_ethereumtype.go:1258] SortAddressContracts: row 5000000, sorted 22423
I0410 22:53:06.572258   10845 rocksdb_ethereumtype.go:1258] SortAddressContracts: row 10000000, sorted 45019
I0410 22:53:11.998873   10845 rocksdb_ethereumtype.go:1258] SortAddressContracts: row 15000000, sorted 67467
I0410 22:53:17.445040   10845 rocksdb_ethereumtype.go:1258] SortAddressContracts: row 20000000, sorted 90004
I0410 22:53:23.074838   10845 rocksdb_ethereumtype.go:1258] SortAddressContracts: row 25000000, sorted 112486
I0410 22:53:25.140106   10845 rocksdb_ethereumtype.go:1290] SortAddressContracts: finished, scanned 26868153 rows, sorted 120904 entries
```
I think to log every 5M rows is kind of optimal from the number of log lines and times between them and also you can see that less than 1% of rows needed to be updated, only a minority of addresses own a nft.

martinboehm:
And thank you for the great work you are doing on Blockbook!

kaladinlight:
5M it is, and thank you! Happy to be helping out such a great project and I appreciate all of the team work :)

kaladinlight:
And actually looking at the sync speed after deploying to aws, I am seeing <15s per 1k blocks in the same block range that `sync. Map` was ~30s. And it looks to have cut mem usage in half almost! So you're assumptions are looking spot on and complexity theory mvp award goes to you haha. :trophy:

jollad (commented):
Good

PR #937, opened 2023-06-07
Bump GRS to 25.0

gruve-p:


gruve-p:
Thanks for merging @martinboehm
I have added [extended index option](https://github.com/trezor/blockbook/commit/d52832f6f7e8cfce0812bfe6c64673d017d7f5a1) for GRS mainnet and GRS testnet however due to this parameter the GRS daemon is no longer starting. Is this parameter only working for BTC (and should it be removed for GRS mainnet and GRS testnet) or is this a bug?

martinboehm:
Hi @gruve-p. The extended index should work for any bitcoin type coin, however it requires new sync of DB as the data in the non extended index are not compatible with extended index. Blockbook will not start using incompatible DB. I would turn it off if you do not need it.

PR #952, opened 2023-07-15
Add Golomb block filters

grdddj:
WIP - see the commit message


grdddj:
@martinboehm I think it is ready for review/takeover

PR #960, opened 2023-08-04
Change symbol for Sepolia

AdamSchinzel:
Adding a new testnet to Suite and after a discussion with guys from product (@sime, @Hannsek) we would like to have `tETH` symbol for Sepolia testnet.


AdamSchinzel:
Closing, not needed now

PR #962, opened 2023-08-07
Change symbol for Sepolia

AdamSchinzel:
Adding a new testnet to Suite and after a discussion with guys from product (@sime, @Hannsek) we would like to have `tSEP` symbol for Sepolia testnet.


vdovhanych (approved):
This doesn't really affect anything, so let's merge it. What do you think @martinboehm ?

PR #967, opened 2023-08-24
Ignore Ordinals in Golomb filters

grdddj:
Builds on top of block-filters branch.

Adds logic to recognize whether a transaction is ordinals-related and in that case not including it into the Golomb filter calculation.

To make it possible, adds a `Witness` field into `Vin`.


grdddj:
Possibly fixes https://github.com/trezor/blockbook/issues/968 and fixes https://github.com/trezor/blockbook/issues/969

Last two commits - [b7f5467](https://github.com/trezor/blockbook/pull/967/commits/b7f54670bfb5e4071f562211fcaeb4cc3403b98a) and [760a1c4](https://github.com/trezor/blockbook/pull/967/commits/760a1c460e617fe5626c6697608ed81a4cdacaca)

@onvej-sl @marekrjpolak @martinboehm 

Two things `Suite`-affecting:
- 1 - We allow for the `Golomb` key to be an array of zeroes
- 2a - We enforce that each request (be in `GET` or `websocket` has a `scriptType` parameter (e.g. `taproot`, `taproot-noordinals`), which must always match the filtering-type used on the `Blockbook` server
- 2b - We add optional `paramM` parameter into all these requests - currently, nothing is happening with it (should we even include it then?)
- 2c - We add `paramP` and `paramM` into the response of all these functions (meaning the simple responses of just one field got more complex by including more fields - and it will break the current `Suite` test parsing)

onvej-sl:
I discussed this with Andrew, and he came up with another suggestion: If we want the API to be really future-proof, the response should include the key as well (provided the key is constant). What do you think?

marekrjpolak:
And what is the hypothetical reason to change the (constant) key in the future?

PR #976, opened 2023-09-19
Remove ropsten

AdamSchinzel:
Ropsten testnet is depracted as of Q4 2022 so no need to have it (and it's also not in Trezor Suite).


martinboehm:
Thank you. We removed the configs but this completes the cleanup 👍

PR #978, opened 2023-09-19
Update README

AdamSchinzel:
I think that it would be better to have `for Trezor Suite` (this line of code was not changed for 5 years)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Blockbook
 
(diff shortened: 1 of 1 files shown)


martinboehm:
Good catch! Please change it to `Blockbook is a back-end service for Trezor Suite. The main features of Blockbook are` to slightly fix English while we are at it :)

PR #982, opened 2023-10-23
btc (+testnet) 25.0 → 25.1

fanquake:


vdovhanych (approved):

Dehumanizer77 (changes requested):
You forgot to change version in all configs. Other than that, it looks fine.

fanquake:
Can you clarify which configs? This is the same as #935.

configs/coins/bitcoin.json
```diff
@@ -23,9 +23,9 @@
         "package_revision": "satoshilabs-1",
         "system_user": "bitcoin",
         "version": "25.0",
```
Dehumanizer77, line 25: This needs to be changed to 25.1 as well as in all subsequent files.

Dehumanizer77 (approved):

PR #986, opened 2023-11-06
Use Erigon for Ethereum Sepolia backend

AdamSchinzel:
Geth for still used for Sepolia archive backend. This changes it to Erigon.


martinboehm:
@AdamSchinzel  Thank you! For the mainnet we already use v2.53.2 of Erigon. Is there a reason to use an older version?

AdamSchinzel:
Sorry, I was using your old commits and there were before the bump, will fix it

PR #994, opened 2023-11-22
perf: add contract info cache for spam token/nft transactions

kaladinlight:
This change adds an in memory cache for contract info on a per transaction basis. The cache fixes a big performance issue when calling `getTransactionFromBchainTx` for spam token/nft transactions. Instead of trying to fetch contract info and failing for thousands of transfers, we can just cache the state of the contract info which should not change during that single transaction fetch (specifically for invalid contracts since they are never saved to the db).

Example transaction: https://polygonscan.com/tx/0xb5b2790f6f38d2a87c28f420515bfb7e6d8a4866e123525ce6285a3c09969bfe

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -36,14 +36,8 @@ type Worker struct {
 	metrics           *common.Metrics
 }
 
-// contractInfoWithValid contains the contract info and whether it is a valid contract or not
-type contractInfoWithValid struct {
-	*bchain.ContractInfo
-	Valid bool
-}
-
 // contractInfoCache is a temporary cache of contract information for ethereum token transfers
-type contractInfoCache = map[string]*contractInfoWithValid
+type contractInfoCache = map[string]*bchain.ContractInfo
 
 // NewWorker creates new api worker
 func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool, txCache *db.TxCache, metrics *common.Metrics, is *common.InternalState, fiatRates *fiat.FiatRates) (*Worker, error) {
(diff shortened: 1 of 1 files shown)


martinboehm:
Thank you, I am going to test it.

martinboehm:
Looks good. Maybe we could do a small optimization in `getEthereumTokensTransfers`, slightly reorder it and add a condition so that it does not allocate the cache map if there are no tokens:
```
func (w *Worker) getEthereumTokensTransfers(transfers bchain.TokenTransfers, addresses map[string]struct{}) []TokenTransfer {
	tokens := make([]TokenTransfer, len(transfers))
	if len(transfers) > 0 {
		sort.Sort(transfers)
		contractCache := make(contractInfoCache)
		for i := range transfers {
	...
```
What do you think?

kaladinlight:
> Looks good. Maybe we could do a small optimization in `getEthereumTokensTransfers`, slightly reorder it and add a condition so that it does not allocate the cache map if there are no tokens:
> 
> ```
> func (w *Worker) getEthereumTokensTransfers(transfers bchain.TokenTransfers, addresses map[string]struct{}) []TokenTransfer {
> 	tokens := make([]TokenTransfer, len(transfers))
> 	if len(transfers) > 0 {
> 		sort.Sort(transfers)
> 		contractCache := make(contractInfoCache)
> 		for i := range transfers {
> 	...
> ```
> 
> What do you think?

Looks good to me! https://github.com/trezor/blockbook/pull/994/commits/1bf88451b6432c287a01f716f96b884dade2221b

martinboehm:
@kaladinlight while looking at the code a bit more, maybe I would modify it slightly more:
1. check if the contract is in the cache directly in the `getEthereumTokensTransfers`
2. if it is there, use it and do not set the cache value again (it is unnecessary, it is already there)
3. only if it is not in cache, call `getContractInfo` (without any caching) and set the cache afterwards

kaladinlight:
@martinboehm that makes sense. I went back and forth a bit on where to add, but this looks cleaner as you suggested: https://github.com/trezor/blockbook/pull/994/commits/819ce3e00b67de1606940b5bc185ce0ffcd7c4c4

martinboehm:
Thank you, you have done it exactly as I wanted 👍 
I hope you will not hate me :) but I just realized that we are not using the Valid flag at all. The cache could be directly `map[string]*bchain.ContractInfo`.

kaladinlight:
> Thank you, you have done it exactly as I wanted 👍 I hope you will not hate me :) but I just realized that we are not using the Valid flag at all. The cache could be directly `map[string]*bchain.ContractInfo`.

This I was aware of, but chose to cache the full state in case that would be used in the future. If there is no plan or intention of using this ever I can remove. If that is the case, would it be worth updating the function sig of `getContractInfo` to be more explicit?

Let me know how you want me to proceed :)

martinboehm:
I would keep the function sig of getContractInfo as it is but as the caching logic is local to `getEthereumTokensTransfers`, we can cache only what we really need.

martinboehm:
Great, thank you! 👍

PR #1005, opened 2023-12-12
Bump Viacoin 0.16.3 to 0.16

romanornr:
Bump viacoin to 0.16.3

add deprecatedrpc="estimatefees"
add coingecko Viacoin rates 

*** I followed the 900-second rule like everyone else, but there's a rate limit on Coingecko API requests. I am not sure if it makes sense to have such a fast refresh rate


martinboehm:
You will get rate limit errors from Coingecko but Blockbook is able to work with that. Or you can purchase API key from Coingecko and provide it to Blockbook using `COINGECKO_API_KEY` env variable.

PR #1007, opened 2023-12-13
Groestlcoin: Bump to 26.0

gruve-p:


martinboehm:
Hi @gruve-p. Are you sure that need the block filter settings? They are used for a very specific use case of Trezor Coinjoin. Do you need this functionality?

gruve-p:
Hi @martinboehm We do not need this functionality however it can not harm to keep it in right? Are there any disadvantages to keep it in?

martinboehm:
@gruve-p Apart from slightly higher load of the server (but probably nothing significant) there is not issue. If you want to, I will merge it as it is.

gruve-p:
@martinboehm Please merge it then as it is. Thank you!

PR #1011, opened 2023-12-18
Refactor in initialization NewCoinGeckoDownloader

romanornr:
1. Moved the creation of `allowedVsCurrenciesMap` to a dedicated function `getAllowedVsCurrenciesMap`. This function takes a `currenciesString` input and returns a map of allowed vs currencies.

2. Updated the `httpTimeout` and `httpClient.Timeout` initialization to use constants (`DefaultHTTPTimeout`) for consistency and to make it easier to adjust the default timeout in the future.

I actually have an idea for the `makeReq` method, but let's see if this is fine first

### DIFF
--- a/fiat/coingecko.go
+++ b/fiat/coingecko.go
@@ -17,6 +17,11 @@ import (
 	"github.com/trezor/blockbook/db"
 )
 
+const (
+	DefaultHTTPTimeout     = 15 * time.Second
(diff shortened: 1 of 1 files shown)


martinboehm:
Hi @romanornr. Thank you for your contribution. Normally we do not accept PRs that modify common code but this PR seems simple enough and is making the code more readable.

PR #1014, opened 2023-12-20
Refactor retry logic in Coingecko API

romanornr:
Refacto retry logic in Coingecko API

- I modified the HTTP status code check to use 'http.StatusOK' for clarity
- Introduced a counter 'attempt' for tracking retry attempts in 'makeReq' function
- Replaced the fixed 60-second wait time with an exponential backoff for retrying after throttling errors

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

Yes, I've seen the message `Normally we do not accept PRs that modify common code`
But I think it's still worth considering it or modifying what I have. 

Unfortunately, I did import math, not sure maybe you guys have a better way

### DIFF
--- a/fiat/coingecko.go
+++ b/fiat/coingecko.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


martinboehm:
@romanornr Hi, thank you for your contribution. However, I am against the change as I think it would slow down the download of the rates. If Blockbook is competing against others on the same IP, it will not help that it voluntarily slows down the download.

PR #1055, opened 2024-03-22
Remove Goerli

AdamSchinzel:
Goerli will be removed from Suite in the next release and overall support for this testnet is coming to an end.

Prepared this so we can then easily get rid of it completely.


martinboehm:
Thanks!

PR #1150, opened 2024-11-01
Add Base Support

kaladinlight:
This PR adds support for Base (includes docker image backend logic from `add-arbitrum`).

I also made some slight changes regarding `addrContracts` storage structure and caching. Happy to discuss this further, but it stemmed from my initial attempts at improving performance for Gnosis chain due to the large amount of spam on chain. The main bottleneck was the storing and fetching of `addrContracts`. To help I made two changes:
- Use protobuf to (de)serialize the data as opposed to manual (un)pack of byte slices under the assumption that protobuf is more performant than the manual byte slice manipulation. It also just modernizes the format to align with some of the other stored types. I did leave the legacy (un)pack functions to ensure everything was backwards compatible so legacy data can be read and then stored back in the new protobuf format.
- Extend the `addressContracts` cache to be supported for both bulk connect and the regular synchronous sync modes. This allows synchronous sync to utilize the cache instead of constantly reading from db to help improve performance. There is a config flag for setting the max size of the contracts map as well for tweaking based on desired memory usage. The cache will auto prune when full prioritizing keeping the larger address contract sets considering those would take the most time to read from the db.

I know this isn't really a direct requirement for Base and didn't even fully solve the Gnosis sync issues, but they still seemed like useful improvements to me, but feel free to let me know if you aren't keen about the changes and I will rip them out.


db/rocksdb_ethereumtype.go
```diff
@@ -347,9 +431,15 @@ func (d *RocksDB) addToAddressesAndContractsEthereumType(addrDesc bchain.Address
 		if ac == nil {
 			ac = &AddrContracts{}
 		}
-		addressContracts[strAddrDesc] = ac
+		if addressContracts != nil {
```
martinboehm, line 434: Is this condition necessary? Why would we pass nil `addressContracts`? I do not see any change in the calling of `addToAddressesAndContractsEthereumType`

kaladinlight: The reason for this is due to `addressContracts` being moved into the `RocksDB` struct and used for both `bulkconnect` and `rocksdb_ethereumtype`. This condition is to handle `nil` being passed from `connectBlockEthereumType` as we don't need a second map being assigned for bulk connect. The local `addrContracts` variable is solely to track `addrContracts` for a single block in the serial sync flow.

kaladinlight: In other words, the `RocksDB.addressContracts` is the cache state variable used by `bulkconnect` always and only, and can be used by the serial sync if configured via `-dbmaxaddrcontracts`. The `addressContracts` param is used only for serial sync to track and store the associated address contracts for each block.

martinboehm: OK, got it. Thank you.

build/templates/backend/scripts/base_archive.sh
```diff
@@ -0,0 +1,47 @@
...
+
+{{define "main" -}}
+
+set -e
+
+GETH_BIN={{.Env.BackendInstallPath}}/{{.Coin.Alias}}/geth
+DATA_DIR={{.Env.BackendDataPath}}/{{.Coin.Alias}}/backend
+
+CHAINDATA_DIR=$DATA_DIR/geth/chaindata
+SNAPSHOT=https://mainnet-archive-snapshots.base.org/$(curl https://mainnet-archive-snapshots.base.org/latest)
```
martinboehm, line 11: Unfortunately they stopped providing archive snapshots for Base for geth. We would have to switch to reth. But for our use case, it would be enough to download non archive snapshot and run it in archive mode. Geth in that case starts to store the archive data for the new blocks.

kaladinlight: d6a94e766bd886c35baca0abbdaa1b7c74469f56

PR #1153, opened 2024-11-05
fix: retry estimatefee with more blocks when getting negative value

grdddj:
Fix for dogecoin, when we receive a negative value from estimatefee, we increase the block query by one and try again ... with specified max retries (3 now)


cranycrane:
hey, thanks for the PR! I have done a follow-up here - https://github.com/trezor/blockbook/pull/1780

PR #1155, opened 2024-11-09
Enable API Before Sync and Return 503 When Not Synced

cranycrane:
## Description:
This PR introduces two changes to improve API behavior during synchronization:

1. **Return HTTP 503 When Not Synced**
API endpoints now return a 503 status if Blockbook is unsynced, helping clients detect incomplete synchronization. Status of the Blockbook in the response was preserved.

2. **Allow API Access Before Sync Completes**
Added an `enableAPIBeforeSync` flag to optionally allow API access before full sync, useful in read-only scenarios.

## Changes:

Updated `jsonHandler` to return 503 if not synced.
Added `enableAPIBeforeSync` flag for optional API access during sync.
## Impact:
Improves API flexibility for clients by providing better control over sync-dependent responses.

### DIFF
--- a/blockbook.go
+++ b/blockbook.go
@@ -415,7 +415,7 @@ func startInternalServer() (*server.InternalServer, error) {
 
 func startPublicServer() (*server.PublicServer, error) {
 	// start public server in limited functionality, extend it after sync is finished by calling ConnectFullPublicInterface
-	publicServer, err := server.NewPublicServer(*publicBinding, *certFiles, index, chain, mempool, txCache, *explorerURL, metrics, internalState, fiatRates, *debugMode)
+	publicServer, err := server.NewPublicServer(*publicBinding, *certFiles, index, chain, mempool, txCache, *explorerURL, metrics, internalState, fiatRates, *debugMode, *enableAPIBeforeSyncFlag)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


cranycrane:
@martinboehm hello, can you please take a look at this?

martinboehm:
Sorry, I cannot accept the pull request, for the following reasons:
1.  I think it is very bad idea to enable public interface when Blockbook is in bulk sync mode. The database is in this case in an inconsistent state and the results are unpredictable
2. the detection of the sync mode is done in a very costly way. You are calling api.GetSystemInfo, which internally calls backend. And this is done for every single API call. You could instead use `s.is.InitialSync` to detect initial sync state. However with a current refactor it is a bit more complicated...
3. there is a merge conflict, due to a refactor from our side

The handling of the initial sync state is not really an issue for us as we make our Blockbooks available only after they finish the syncing.

Also, please be aware that the PR has not complied with our contribution guide https://github.com/trezor/blockbook/blob/master/CONTRIBUTING.md

cranycrane:
Hi @martinboehm,

Thank you for pointing out the issues with my pull request and for referring me to the contribution guide. I apologize for not fully complying with the guidelines. 

Back to the point, I’d like to simplify my proposal to implement only the functionality where the API returns a 503 status code when synchronization is incomplete. This would make clearer communication with clients without major changes in the current architecture when clients interpret 2xx status codes as correct behaviour but do not receive expected data.

Would this adjustment make the PR acceptable?

martinboehm:
Thank you. We have already implemented something similar, in this commit 95e965d5dfd12b3bdeeeaa6278f04f2c9ee08945

PR #1185, opened 2025-01-13
Rename Token type to standard

enjojoy:
Renamed token type to standard, leaving token as deprecated


martinboehm:
Merged as a part of bigger change

PR #1198, opened 2025-02-03
Publish confirmed transactions by address

kaladinlight:
Currently in order to push confirmed transaction notifications to a client, you have to:
- Subscribe to new blocks
- Fetch block from blockbook on new block
- Check if any subsccribed addresses are associated with any transactions in the block

Historically this has worked without any noticeable issue, however in integrating Base, I have noticed this no longer holds up due to the fast block times and high transaction throughput on chain. The node gets oversaturated with network requests causing it to fall out of sync along with blockbook. Even when they are staying in sync, I was seeing calls to get a block taking more than 30 minutes. I did make many attempts at increasing resources, cache sizes, etc. to improve performance without avail.

To reduce most overhead related to this scenario, I chose to add an optional `newBlockTxs` param to `subscribeAddresses` which will allow subscribing to addresses for mempool AND confirmed transactions now. By leveraging the mostly parsed sync blocks, we can detect all associated addresses for a transaction to determine if a confirmed transaction should be published to a client and only then, fetch the evm transaction receipt to ensure an `api.Tx` is sent to the client. This update cuts down nearly all of the previously seen network requests and completely fixed the performance issues seen in the above scenario.

(I went with `newBlockTxs` just to tie into the new block vernacular, but could update to `confirmedTxs` if that sounds better.)


server/websocket.go
```diff
@@ -955,9 +958,60 @@ func (s *WebsocketServer) onNewBlockAsync(hash string, height uint32) {
 	glog.Info("broadcasting new block ", height, " ", hash, " to ", len(s.newBlockSubscriptions), " channels")
 }
 
+func (s *WebsocketServer) publishNewBlockTxsByAddr(block *bchain.Block) {
+	s.addressSubscriptionsLock.Lock()
+	publishNewBlockTxs := func() bool {
```
martinboehm, line 963: Why is this a function? So that you can break the inner loop by return? 
I am a bit worried that it could be fairly slow if there are many channels with many subscribed addresses and only few of them subscribe to publishNewBlockTxs and this code is executed for every block. How about having a counter of publishNewBlockTxs subscriptions, when on subscribe we increment the counter and on unsubscribe (or on channel close) we decrement the counter. Then on new block the condition is simple (and we could do it even before starting the goroutine).

server/websocket.go
```diff
@@ -955,9 +958,60 @@ func (s *WebsocketServer) onNewBlockAsync(hash string, height uint32) {
...
+				for i, vin := range tx.Vin {
+					vins[i] = bchain.MempoolVin{Vin: vin}
+				}
+				subscribed := s.getNewTxSubscriptions(vins, tx.Vout, tokenTransfers)
+				if len(subscribed) > 0 {
+					if csd, ok := tx.CoinSpecificData.(bchain.EthereumSpecificData); ok {
+						receipt, err := s.chain.EthereumTypeGetTransactionReceipt(tx.Txid)
+						if err != nil {
+							glog.Error("GetTransactionFromMempoolTx error ", err, " for ", tx.Txid)
+							return
```
martinboehm, line 992: I think there are copy/paste issues... In addition to that, I am a bit surprised that the receipt is not already in the CoinSpecificData as it must be fetched during block processing to get the token transfers.

kaladinlight: Good catch, fixed: 61bd16f7f12de217bc6b5800bec1b59c554dce57

The receipt contains the logs relevant for processing transfers so we are able to detect addresses without any extra rpc calls required, however, the constructed receipt from `ethrpc.GetBlock` is missing the gas/status/fee data which is required to be fetched from the node to return the full `api.Tx`

server/websocket.go
kaladinlight: 68d930ae98cec403808d5aaa3e0f1704c8b6e628

etimofeeva:
@kaladinlight is this a finished PR ready for review?

kaladinlight:
> @kaladinlight is this a finished PR ready for review?

@etimofeeva yep, it is ready to go!

docs/api.md
```diff
@@ -1035,6 +1035,19 @@ Example for subscribing to an address (or multiple addresses)
...
 
+Example for subscribing to an address (or multiple addresses) including new block (confirmed) transactions
+
+```javascript
+{
+  "id":"1",
+  "method":"subscribeAddresses",
+  "params":{
+    "addresses":["mnYYiDCb2JZXnqEeXta1nkt5oCVe2RVhJj", "tb1qp0we5epypgj4acd2c4au58045ruud2pd6heuee"]
+    "newBlockTxs" true,
```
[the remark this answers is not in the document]
kaladinlight: 7c1818977963d9ce668bbf5b3f39ec845acbd558

server/websocket.go
```diff
@@ -1061,48 +1124,51 @@ func (s *WebsocketServer) sendOnNewTxAddr(stringAddressDescriptor string, tx *ap
...
 			}
-		}
-		addrDesc, err = s.chainParser.GetAddrDescFromAddress(tx.TokenTransfers[i].To)
-		if err == nil && len(addrDesc) > 0 {
-			sad := string(addrDesc)
-			as, ok := s.addressSubscriptions[sad]
-			if ok && len(as) > 0 {
-				subscribed[sad] = struct{}{}
+		} else if s.chainParser.GetChainType() == bchain.ChainBitcoinType {
+			processVout(vouts[vins[i].Vout])
```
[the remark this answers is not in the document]
kaladinlight: 7c1818977963d9ce668bbf5b3f39ec845acbd558

server/websocket.go
```diff
@@ -1014,9 +1029,54 @@ func (s *WebsocketServer) onNewBlockAsync(hash string, height uint32) {
...
+				if csd, ok := tx.CoinSpecificData.(bchain.EthereumSpecificData); ok {
+					receipt, err := s.chain.EthereumTypeGetTransactionReceipt(tx.Txid)
+					if err != nil {
+						glog.Error("EthereumTypeGetTransactionReceipt error ", err, " for ", tx.Txid)
+						return
+					}
+					csd.Receipt = receipt
+					tx.CoinSpecificData = csd
+				}
+				atx, err := s.api.GetTransactionFromBchainTx(&tx, int(block.Height), false, false, nil)
```
pragmaxim, line 1191: Confirmed block transactions are treated as mempool (confirmations=0, blocktime reset).
   - References:  `api/worker.go:489`, `api/worker.go:516`
   - Why it matters: For Bitcoin-type blocks parsed via `ParseBlock`, `bchain.Tx.Confirmations` stays at 0. The websocket path calls `GetTransactionFromBchainTx` which then pulls `GetTransactionSpecific` and overwrites blocktime using mempool time, emitting a “confirmed” tx with `confirmations=0`.
   - Suggested fix (set confirmations/blocktime before conversion):
     ```go
     for i := range block.Txs {
         tx := block.Txs[i]
         if tx.Confirmations == 0 {
             tx.Confirmations = 1
             tx.Blocktime = block.Time
             tx.Time = block.Time
         }
         // existing publish flow...
         atx, err := s.api.GetTransactionFromBchainTx(&tx, int(block.Height), false, false, nil)
         // ...
     }
     ```
     Alternative: guard the mempool-only path in `GetTransactionFromBchainTx` so `height > 0` does not trigger the mempool-specific flow.

server/websocket.go
```diff
@@ -1014,9 +1029,54 @@ func (s *WebsocketServer) onNewBlockAsync(hash string, height uint32) {
...
+	for _, tx := range block.Txs {
+		var tokenTransfers bchain.TokenTransfers
+		var internalTransfers []bchain.EthereumInternalTransfer
+		if s.chainParser.GetChainType() == bchain.ChainEthereumType {
+			tokenTransfers, _ = s.chainParser.EthereumTypeGetTokenTransfersFromTx(&tx)
+			esd := tx.CoinSpecificData.(bchain.EthereumSpecificData)
+			if esd.InternalData != nil {
+				internalTransfers = esd.InternalData.Transfers
+			}
+		}
```
pragmaxim: unsafe type assertion can panic for non-ethereum txs with missing coin-specific data.
   - Suggested fix:
     ```go
     if s.chainParser.GetChainType() == bchain.ChainEthereumType {
         tokenTransfers, _ = s.chainParser.EthereumTypeGetTokenTransfersFromTx(&tx)
         if esd, ok := tx.CoinSpecificData.(bchain.EthereumSpecificData); ok && esd.InternalData != nil {
             internalTransfers = esd.InternalData.Transfers
         }
     }
     ```

server/websocket.go
```diff
@@ -1014,9 +1029,54 @@ func (s *WebsocketServer) onNewBlockAsync(hash string, height uint32) {
...
+			go func(tx bchain.Tx, subscribed map[string]struct{}) {
+				if csd, ok := tx.CoinSpecificData.(bchain.EthereumSpecificData); ok {
+					receipt, err := s.chain.EthereumTypeGetTransactionReceipt(tx.Txid)
+					if err != nil {
+						glog.Error("EthereumTypeGetTransactionReceipt error ", err, " for ", tx.Txid)
+						return
+					}
+					csd.Receipt = receipt
+					tx.CoinSpecificData = csd
+				}
```
pragmaxim: receipt fetch failure drops the notification entirely.
   - Suggested fix (fallback to sending without receipt):
     ```go
     if csd, ok := tx.CoinSpecificData.(bchain.EthereumSpecificData); ok {
         receipt, err := s.chain.EthereumTypeGetTransactionReceipt(tx.Txid)
         if err != nil {
             glog.Error("EthereumTypeGetTransactionReceipt error ", err, " for ", tx.Txid)
             // continue without receipt
         } else {
             csd.Receipt = receipt
             tx.CoinSpecificData = csd
         }
     }
     ```

etimofeeva (approved):

PR #1217, opened 2025-03-20
Add UnmarshalJSON method to Amount

kedazo:
Hey, I've tried to use the api.Address type in an application I write,
however the api.Amount was missing the JSON unmarshal function:

`json: cannot unmarshal string into Go struct field Address.balance of type api.Amount`

This small mod (+test) solves this issue.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -3,8 +3,10 @@ package api
 import (
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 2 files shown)


martinboehm:
Blockbook does not need this functionality but for the sake of completeness it makes sense..

PR #1259, opened 2025-05-23
Support no-compression on rocksdb for Windows

cpuchainorg:
This occurs for Windows builds when built on dynamic linking


martinboehm:
Thank you!

PR #1273, opened 2025-06-16
Add Tron Network Support to Blockbook

cranycrane:
The focus of my Bachelor's thesis was to implement a blockchain explorer for Tron. I have used Blockbook for this and would like to contribute my work to this repository. 

# Summary of changes

1. Added Coin configuration for **Tron** and **Tron Nile Testnet**
2. `TronRPC` extends the `EthereumRPC` due to the fact, that **Java-Tron** client has JSON-RPC interface compatible with Ethereum
3. The interface is not fully compatible, which lead to specific implementations and certain limitations described bellow
4. Unit, JSON-RPC, API and FakeBlockchain tests were implemented.
5. Extended the **ZeroMQ** implementation to be modular due to the fact, that **Java-Tron** uses ZeroMQ for notifying of new blocks and transactions.

# Merge Checklist
 - [x] Coin configuration file added (tron.json)
 - [x] `tron` package implements JSON-RPC, parser, client, and types
 - [x] Unit and integration tests added and passing
 - [x] Documentation updated (docs/ports.md, docs/config.md)
 - [ ] Code review by at least one maintainer

# Known limitations
- Java-Tron JSON-RPC does not support equivalent method to `eth_sendRawTransaction`. This could be solved by implementing HTTP Client and calling the `/wallet/broadcasthex` endpoint in HTTP API.
- TRC-10 tokens are specific function of Tron, recognizition of such transaction can be done only through the HTTP API.
- This solution relies on the current implementation of the `eth` package, but utilizes the current code and minimizes duplicates.

**The Tron Blockbook is currently running on [http://147.229.8.210:51661/](http://147.229.8.210:51661/).**

I will be happy to discuss my implementation and improve my current solution, so it can be integrated into the Blockbook repository.


MiroslavProchazka:
@etimofeeva Can you also take a look, please ? 🙏

MiroslavProchazka:
Hello @cranycrane, thank you a lot for your contribution! 
Can you please reach out to me at miroslav.prochazka@satoshilabs.com, so we can chat about the PR outside of GH? 

Thanks a lot! 
MP

overcat:
Hi @cranycrane, could you share the hardware requirements for running a mainnet node and a testnet node?

tests/tests.json
```diff
@@ -1,253 +1,259 @@
...
+  "omotenashicoin_testnet": {
+    "rpc": ["GetBlock", "GetBlockHash", "GetTransaction", "GetTransactionForMempool", "MempoolSync",
+      "EstimateSmartFee", "EstimateFee", "GetBestBlockHash", "GetBestBlockHeight"],
+    "sync": ["ConnectBlocksParallel", "ConnectBlocks"]
+  },
+  "trezarcoin": {
+    "rpc":  ["GetBlock", "GetBlockHash", "GetTransaction", "GetTransactionForMempool", "MempoolSync",
+      "EstimateFee", "GetBestBlockHash", "GetBestBlockHeight", "GetBlockHeader"],
+    "sync": ["ConnectBlocksParallel", "ConnectBlocks", "HandleFork"]
+  },
```
etimofeeva: since you're adding only tron and tron_testnet_nile test cases, could you fix the formatting so that only your new changes are displayed in PR, please?

bchain/coins/bsc/bscrpc.go
```diff
@@ -38,7 +38,7 @@ func NewBNBSmartChainRPC(config json.RawMessage, pushHandler func(bchain.Notific
 	s := &BNBSmartChainRPC{
 		EthereumRPC: c.(*eth.EthereumRPC),
 	}
-	s.Parser.EnsSuffix = ".bnb"
+	s.Parser.(*eth.EthereumParser).EnsSuffix = ".bnb"
```
etimofeeva: is this change intentional? I think this could be refactored so that the change here isn't needed: either through constructor or interface.

bchain/coins/tron/tronparser.go
```diff
@@ -0,0 +1,191 @@
...
+func (p *TronParser) FromTronAddressToHex(addr string) string {
+	desc, err := p.GetAddrDescFromAddress(addr)
+	if err != nil {
+		return addr
+	}
+	return "0x" + hex.EncodeToString(desc)
+}
+
+func (p *TronParser) IsTronAddress(desc bchain.AddressDescriptor) bool {
+	return len(desc) == TronTypeAddressDescriptorLen && desc[0] == 0x41
```
overcat: The `AddressDescriptor` has already been stripped of the 0x41 prefix in `GetAddrDescFromAddress`, so `desc[0] == 0x41` will never be true.

bchain/coins/tron/tronparser.go
```diff
@@ -0,0 +1,191 @@
...
+	r.Tx.To = p.FromTronAddressToHex(r.Tx.To)
+
+	for _, l := range r.Receipt.Logs {
+		l.Address = p.FromTronAddressToHex(l.Address)
+	}
+
+	return p.EthereumParser.PackTx(tx, height, blockTime)
+}
+
+func SanitizeHexUint64String(s string) string {
```
overcat, line 320: It would be even better if we could add a document explaining why we need it.

bchain/coins/tron/evm.go
```diff
@@ -0,0 +1,278 @@
...
+	rpcClient *TronRPCClient
+}
+
+// EstimateGas returns the current estimated gas cost for executing a transaction
+func (c *TronClient) EstimateGas(ctx context.Context, msg interface{}) (uint64, error) {
+	return c.Client.EstimateGas(ctx, msg.(ethereum.CallMsg))
+}
+
+// BalanceAt returns the balance for the given account at a specific block, or latest known block if no block number is provided
+// Only parameter 'latest' is available
```
overcat:
```suggestion
// IMPORTANT: Tron RPC only supports 'latest' block parameter. The blockNumber parameter is ignored.
// This is a Tron limitation - historical balance queries are not supported.
```

bchain/coins/tron/evm.go
```diff
@@ -0,0 +1,278 @@
...
+		rawData, err = fixStateRoot(rawData)
+		if err != nil {
+			return err
+		}
+	}
+
+	return json.Unmarshal(rawData, result)
+}
+
+func fixStateRoot(data []byte) ([]byte, error) {
```
overcat:
```suggestion
// fixStateRoot works around Tron JSON-RPC returning stateRoot in a format incompatible with go-ethereum
// Issue: Tron returns stateRoot as "0x" (empty) or with incorrect length, which causes go-ethereum
// deserialization to fail since it expects a valid 32-byte hash (66 chars: "0x" + 64 hex digits)
//
// This is likely because Tron uses a different state storage mechanism than Ethereum's MPT (Merkle Patricia Tree),
// but still tries to maintain API compatibility. The stateRoot field may not have the same meaning in Tron.
//
// Workaround: Replace invalid stateRoot with a zero hash to allow successful parsing by go-ethereum library
// Reference: https://github.com/tronprotocol/java-tron/issues/5518
func fixStateRoot(data []byte) ([]byte, error) {
```

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,368 @@
...
+	if err != nil {
+		return nil, errors.Annotatef(err, "txid %v", txid)
+	}
+
+	return &receipt, nil
+}
+
+// Tron does not have any method for getting mempool transactions (does not support parameter 'pending' in eth_getBlockByNumber)
+// https://developers.tron.network/reference/eth_getblockbynumber
+func (b *TronRPC) GetMempoolTransactions() ([]string, error) {
```
overcat: Would it be better to return an empty list here instead of nil?

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,368 @@
...
+
+func (b *TronRPC) getTransactionReceipt(txid string) (*bchain.RpcReceipt, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), b.Timeout)
+	defer cancel()
+
+	hash := ethcommon.HexToHash(txid)
+	var receipt bchain.RpcReceipt
+	err := b.RPC.CallContext(ctx, &receipt, "eth_getTransactionReceipt", hash)
+	if err != nil {
+		return nil, errors.Annotatef(err, "txid %v", txid)
```
overcat:
```suggestion
		return nil, errors.Annotatef(err, "failed to get transaction receipt for txid %v", txid)
```

bchain/coins/tron/evm.go
```diff
@@ -0,0 +1,278 @@
...
+// Tron RPC returns genesis block
+func (c *TronClient) NetworkID(ctx context.Context) (*big.Int, error) {
+	var ver string
+
+	if err := c.rpcClient.CallContext(ctx, &ver, "net_version"); err != nil {
+		return nil, err
+	}
+
+	switch ver {
+	case "0x2b6653dc":
```
overcat: It might be better to use const.
```	
const (
	    MainnetGenesisHash = "0x2b6653dc"
	    NileTestnetGenesisHash = "0xcd8690dc"
)
```

bchain/coins/tron/tronparser.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	if len(address) == TronAddressLen {
+		decoded := base58.Decode(address)
+		if len(decoded) != 25 || decoded[0] != 0x41 {
+			return nil, errors.New("invalid Tron base58 address")
+		}
+		return decoded[1:21], nil
+	} else if len(address) != TronTypeAddressDescriptorLen*2 {
+		return nil, bchain.ErrAddressMissing
```
overcat, line 72: Would it be better to print out the incorrect address for debugging?

bchain/coins/tron/tronparser.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return address
+	}
+	return ToTronAddressFromDesc(b)
+}
+
+func (p *TronParser) FromTronAddressToHex(addr string) string {
+	desc, err := p.GetAddrDescFromAddress(addr)
+	if err != nil {
+		return addr
```
overcat: Consider logging the error or returning an error to caller?

bchain/coins/bsc/bscrpc.go
cranycrane: refactored through `EthereumLikeParser` interface

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,368 @@
...
+		return nil, nil
+	}
+	contract.Contract = ToTronAddressFromAddress(contract.Contract)
+	glog.Infof("Getting contract info for: %s", contract.Contract)
+	return contract, nil
+}
+
+// SendRawTransaction is not supported by Tron JSON-RPC
+func (b *TronRPC) SendRawTransaction(hex string) (string, error) {
+	return "", nil
```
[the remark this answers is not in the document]
cranycrane: resolved.

bchain/coins/tron/tronparser.go
cranycrane: resolved.

bchain/coins/tron/evm.go
cranycrane: resolved.

bchain/coins/tron/evm.go
cranycrane: resolved.

bchain/coins/tron/evm.go
cranycrane: resolved.

bchain/coins/tron/tronparser.go
cranycrane: added logging. resolved.

bchain/coins/tron/tronparser.go
cranycrane: added returning an error to caller and checking err when calling. resolved

bchain/coins/tron/tronrpc.go
cranycrane: resolved.

bchain/coins/tron/tronrpc.go
cranycrane: yes. thank you. resolved.

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+// EthereumTypeGetRawTransaction is not supported by Tron JSON-RPC
+func (b *TronRPC) EthereumTypeGetRawTransaction(txid string) (string, error) {
+	return "", nil
```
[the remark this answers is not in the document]
cranycrane: resolved.

tests/tests.json
cranycrane: resolved.

bchain/coins/tron/tronparser.go
```diff
@@ -0,0 +1,248 @@
...
+	r.Tx.From, err = p.FromTronAddressToHex(r.Tx.From)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert 'from' address: %w", err)
+	}
+
+	r.Tx.To, err = p.FromTronAddressToHex(r.Tx.To)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert 'to' address: %w", err)
+	}
+
```
pragmaxim, line 256: at line 186, I think that TronParser.PackTx errors when `to` is empty/null (contract creation), causing pack/DB write failures for those txs; guard for empty value or treat it as "0x" before conversion perhaps?

bchain/coins/tron/evm.go
```diff
@@ -0,0 +1,293 @@
...
+	var result hexutil.Big
+	err := c.rpcClient.CallContext(ctx, &result, "eth_getBalance", common.BytesToAddress(addrDesc), "latest")
+	return (*big.Int)(&result), err
+}
+
+// NonceAt is not supported by Tron RPC
+func (c *TronClient) NonceAt(ctx context.Context, addrDesc bchain.AddressDescriptor, blockNumber *big.Int) (uint64, error) {
+	return 1, nil
+}
+
```
pragmaxim, line 47: if NonceAt is not supported, shouldn't it return `0` instead of `1` ?

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -0,0 +1,205 @@
...
+			return bchain.CREATE,
+				ToTronAddressFromAddress(itx.TransferToAddress),
+				nil
+
+		case bchain.SELFDESTRUCT:
+			topType = bchain.SELFDESTRUCT
+
+		default:
+			panic("unhandled eth internal transaction type")
+		}
```
pragmaxim, line 254: if Tron adds new `note` values, block processing can crash, maybe default to `CALL` or is it possible to not panic here ?

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,379 @@
...
+	var newContracts []bchain.ContractInfo
+
+	for i := range block.Txs {
+		tx := &block.Txs[i]
+		csd, ok := tx.CoinSpecificData.(bchain.EthereumSpecificData)
+		if !ok || csd.Tx == nil {
+			continue
+		}
+
+		if csd.Tx.To == "" && csd.Tx.GasLimit != "0x0" {
```
pragmaxim: TronRPC.GetBlock does a per‑tx eth_getTransactionReceipt call for contract-creation candidates; on blocks with multiple creations, this adds N extra round trips and can bottleneck sync. Consider batching or sourcing contract address from existing block/txinfo data.

In a block with many contract creations, this becomes N extra RPC round trips, which can significantly slow sync.

pragmaxim (changes requested):
Hi @cranycrane, I noticed of a few places that would need either explanation or improvement.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func NewTronInternalDataProvider(http TronHTTP, timeout time.Duration) *TronInternalDataProvider {
+	return &TronInternalDataProvider{
+		http:    http,
+		timeout: timeout,
+	}
+}
+
+func (p *TronInternalDataProvider) GetInternalDataForBlock(
```
pragmaxim, line 52: If `processInternalTransactions` is enabled, there is an extra HTTP call per block to `/wallet/gettransactioninfobyblocknum`, plus JSON decode ... possible bottleneck.

bchain/coins/tron/evm.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	if err != nil {
+		return nil, errors.Annotatef(err, "hash %v, height %v", hash, height)
+	} else if len(raw) == 0 || (len(raw) == 4 && string(raw) == "null") {
+		return nil, bchain.ErrBlockNotFound
+	}
+	return raw, nil
+}
+
+func (c *TronRPCClient) CallContext(ctx context.Context, result interface{}, method string, args ...interface{}) error {
```
pragmaxim, line 230: Tron’s `fixStateRoot` does a full JSON unmarshal+marshal for every block RPC (hash/number). This happens before the block is unmarshaled again in GetBlock, so large blocks get decoded multiple times.

pragmaxim (changes requested):
I found a few more chances for possible throughput improvement

bchain/coins/tron/tronparser.go
cranycrane: you are right, fixed.

bchain/coins/tron/tronInternalDataProvider.go
cranycrane: left default + logging warning so it is kept in mind , that something has changed, when new `note` value appears

bchain/coins/tron/tronrpc.go
cranycrane: thanks for pointing this. I have implemented the tron internal transaction processing recently, therefore the detection of contract creation/destruction is handled there. TronRPC.GetBlock is not needed anymore and I have removed it.

bchain/coins/tron/evm.go
cranycrane: fixed.

bchain/coins/tron/tronInternalDataProvider.go
cranycrane: if we want internal transaction to be processed, there needs to be this HTTP call to get the data. It is the same in `ethrpc.go` that calls `debug_traceBlockByHash` when processing internal transactions.

bchain/coins/tron/evm.go
cranycrane: thank you. I have refactored the method so it does not fix the root when the result interface is `json.RawMessage`, because it is only needed inside the `ethclient` when unmarshaling into the `types.Header` struct. Also I have optimized the `fixStateRoot` so it only replaces the bytes without the need to unmarshal and marshal again.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -0,0 +1,219 @@
...
+
+		topType, contractAddr, err := detectTopType(info.InternalTransactions)
+		d.Type = topType
+		if err != nil {
+			return data, contracts, err
+		}
+
+		if contractAddr != "" {
+			d.Contract = contractAddr
+		}
```
pragmaxim: This is probably just marginal case, but contract creation detected does not update `d.Type` or `d.Contract` after `topType` is overridden, so internal data can report CALL with empty contract even while `contracts` is populated. This makes internal tx data inconsistent for create-only txs (no internal create note).

Suggested fix :
```go
  topType, contractAddr, err := detectTopType(info.InternalTransactions)
  if err != nil {
      return data, contracts, err
  }

  if topType == bchain.CALL && info.ContractAddress != "" {
      topType = bchain.CREATE
      contractAddr = ToTronAddressFromAddress(info.ContractAddress)
  }

  d.Type = topType
  if contractAddr != "" {
      d.Contract = contractAddr
  }

  if topType == bchain.CREATE && contractAddr != "" {
      contracts = append(contracts, bchain.ContractInfo{
          Contract:       contractAddr,
          CreatedInBlock: blockHeight,
          Standard:       bchain.UnhandledTokenStandard,
      })
  } else if topType == bchain.SELFDESTRUCT {
      contracts = append(contracts, bchain.ContractInfo{
          Contract:          contractAddr,
          DestructedInBlock: blockHeight,
      })
  }
```

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,311 @@
...
+			ScriptPubKey: bchain.ScriptPubKey{
+				Addresses: []string{ToTronAddressFromAddress(csd.Receipt.ContractAddress)}},
+		}}
+
+		csd.InternalData = &bchain.EthereumInternalData{
+			Type:     bchain.CREATE,
+			Contract: ToTronAddressFromAddress(csd.Receipt.ContractAddress),
+		}
+		tx.CoinSpecificData = csd
+	}
```
pragmaxim: `GetTransaction` assumes `csd.Receipt` is non-nil. For mempool contract-creation transactions (no `to` address), `csd.Receipt` is nil and `csd.Receipt.ContractAddress` will panic.

  Suggested fix:
  ```go
  if len(tx.Vout) > 0 &&
      tx.Vout[0].ScriptPubKey.Addresses == nil &&
      csd.Receipt != nil &&
      csd.Receipt.ContractAddress != "" {
      tx.Vout = []bchain.Vout{{
          ValueSat: tx.Vout[0].ValueSat,
          N:        0,
          ScriptPubKey: bchain.ScriptPubKey{
              Addresses: []string{ToTronAddressFromAddress(csd.Receipt.ContractAddress)},
          },
      }}

      csd.InternalData = &bchain.EthereumInternalData{
          Type:     bchain.CREATE,
          Contract: ToTronAddressFromAddress(csd.Receipt.ContractAddress),
      }
      tx.CoinSpecificData = csd
  }
  ```

bchain/coins/tron/tronparser.go
```diff
@@ -0,0 +1,250 @@
...
+		address = address[2:]
+	}
+
+	if len(address) == TronAddressLen {
+		decoded := base58.Decode(address)
+		if len(decoded) != 25 || decoded[0] != 0x41 {
+			return nil, errors.New("invalid Tron base58 address")
+		}
+		return decoded[1:21], nil
+	} else if len(address) != TronTypeAddressDescriptorLen*2 {
```
pragmaxim, line 70: Base58 address parsing does not validate checksum (uses `base58.Decode` with length/prefix checks only). Invalid addresses with the right length/prefix can be accepted and treated as valid.

  Suggested fix:
  ```go
  decoded := base58.Decode(address)
  if len(decoded) != 25 || decoded[0] != 0x41 {
      return nil, errors.New("invalid Tron base58 address")
  }
  payload := decoded[:21]
  checksum := decoded[21:]
  first := sha256.Sum256(payload)
  second := sha256.Sum256(first[:])
  if !bytes.Equal(checksum, second[:4]) {
      return nil, errors.New("invalid Tron base58 checksum")
  }
  return payload[1:], nil
  ```

bchain/coins/tron/tronInternalDataProvider.go
cranycrane: fixed

bchain/coins/tron/tronrpc.go
cranycrane: thanks. fixed

bchain/coins/tron/tronparser.go
cranycrane: fixed

pragmaxim (approved):
Perfect, LGTM!

bchain/coins/tron/tronparser.go
```diff
@@ -0,0 +1,348 @@
...
+func ToTronAddressFromAddress(address string) string {
+	if has0xPrefix(address) {
+		address = address[2:]
+	}
+	b, err := hex.DecodeString(address)
+	if err != nil {
+		return address
+	}
+	return ToTronAddressFromDesc(b)
+}
```
pragmaxim, line 125: This basically turns empty input string into a synthetic Tron Base58Check address `8M5JoSg` ... 

This propagates into multiple call sites that pass optional fields without checking for emptiness, for example `delegateTo` and `tx.From` in https://github.com/trezor/blockbook/pull/1273/changes?mode=single#diff-ab8e0c1865ac6356c3faaad5e7d327cad0dd583e78debe1ec51717c3afca7501R157, plus internal-tx address handling in https://github.com/trezor/blockbook/pull/1273/changes?mode=single#diff-0b7f2503844081fafaad9abbc99b69073d8b0757fc8b9585353e1372a209f9d4R157 and others.

The result is silent data corruption: a missing backend field turns into a seemingly valid Tron address which then gets exposed through API/UI or indexed as a contract/internal-transfer participant. This should return `""` (or error) for empty input before any base58 conversion.

How this happens: The backend response structs use plain string fields, not *string. In Go, if JSON omits such a field, it becomes "". That is true for `owner_address`, `receiver_address`, `to_address`, `contract_address` in https://github.com/trezor/blockbook/pull/1273/changes?mode=single#diff-0b7f2503844081fafaad9abbc99b69073d8b0757fc8b9585353e1372a209f9d4R24

Those fields are then passed into `ToTronAddressFromAddress` without an empty check in several places:
```
bchain/coins/tron/txextra.go:95
bchain/coins/tron/txextra.go:157
bchain/coins/tron/tronInternalDataProvider.go:187
bchain/coins/tron/tronInternalDataProvider.go:188
bchain/coins/tron/tronInternalDataProvider.go:246
bchain/coins/tron/tronInternalDataProvider.go:250
```

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,935 @@
...
+		if txByID == nil { // todo possibly can be deleted
+			b.ObserveChainDataFallback("tron_getblock", "missing_tx_by_id_map")
+			glog.V(1).Infof("Tron GetBlock fallback to gettransactionbyid for tx %s in block %d", tx.Hash, bbh.Height)
+			txByID, err = b.getTransactionByID(tx.Hash, isSolidified)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		txInfo := txInfosByID[strip0xPrefix(tx.Hash)]
```
pragmaxim, line 681: `GetBlock` has a fallback for missing `gettransactionbyid`, but not for missing `gettransactioninfobyblocknum`, so partial HTTP responses can still take the whole block path down.

`txInfo` is used directly and then passed into `buildTxFromHTTPData` and then `buildTxFromHTTPData` immediately calls `tronBuildEthereumSpecificData(txByID, txInfo)`. 

So it either crash or in case of a silent/swallowed error, we could loose a lot of transactions without knowing it? 

The best fix is to mirror the existing txByID fallback and fetch the missing tx info per tx:
```
txInfo := txInfosByID[strip0xPrefix(tx.Hash)]
if txInfo == nil {
    b.ObserveChainDataFallback("tron_getblock", "missing_tx_info_by_block")
    glog.V(1).Infof("Tron GetBlock fallback to gettransactioninfobyid for tx %s in block %d", tx.Hash, bbh.Height)

    txInfo, err = b.getTransactionInfoByID(tx.Hash, isSolidified)
    if err != nil && isTronTxNotFound(err) {
        txInfo, _, err = b.getTransactionInfoByIDWithFallback(tx.Hash)
    }
    if err != nil {
        return nil, err
    }
}

if txInfo == nil {
    return nil, errors.Errorf("missing txInfo for txid %v", txByID.TxID)
}
```

bchain/coins/tron/tronrpc.go
```diff
@@ -0,0 +1,935 @@
...
+		}
+		txByID, err := b.getTransactionByID(txid, false)
+		if err != nil {
+			return nil, err
+		}
+
+		blockTime, blockNumber, hasBlockNumber := tronTxMeta(txInfo)
+		confirmations := b.computeConfirmationsFromBlockNumber(txid, blockNumber, hasBlockNumber)
+		return b.buildTxFromHTTPData(txByID, txInfo, blockTime, confirmations, nil, false)
+	}
```
pragmaxim: A recently mined tx can still be returned as pending if it is still present in the local mempool cache.

`GetTransaction` short-circuits on local mempool membership (lines `785-797`). That branch fetches fresh backend data, computes `blockNumber`/confirmations, and then still calls `buildTxFromHTTPData(..., false)`. Inside `buildTxFromHTTPData`, `false` forces `csd.Receipt = nil`, which makes the tx pending-style even if the backend is already returning block metadata.

Because mempool eviction is asynchronous, there is a stale window where API/Suite can observe a mined tx as still pending. The branch should derive pending/confirmed from the fetched backend response, not from the presence of an old local mempool entry.

Basically the fix is to remove this whole branch, since it does not add new capability. It only changes the decision source for “pending vs confirmed” from backend state to local mempool state, and that is exactly what causes the bug.

bchain/coins/tron/tronparser.go
cranycrane: resolved by returing "" when empty string provided (+ added tests)

bchain/coins/tron/tronrpc.go
cranycrane: deleted the mempool check branch as recommended. 

But just a note so we are on the same page - tron transaction can be included in a block and be "unconfirmed"/pending at the same time. The transaction is confirmed after 19 next blocks are created (just check on tronscan some new block and wait about 1 minute when it changes its status from "unconfirmed"->"confirmed")

bchain/coins/tron/tronrpc.go
cranycrane: fixed - added fallback for missing `txInfo`

PR #1289, opened 2025-07-25
ens resolver 

etimofeeva:
wip of ens resolver 

workflow: 
types.go → ethparser.go → worker.go


bchain/coins/eth/ethrpc.go
```diff
@@ -1179,3 +1197,159 @@ func (b *EthereumRPC) EthereumTypeGetNonce(addrDesc bchain.AddressDescriptor) (u
...
+	name = strings.ToLower(strings.TrimSpace(name))
+
+	// Only check expiration for .eth domains
+	if !strings.HasSuffix(name, ".eth") {
+		glog.Infof("CheckENSExpiration: %s is not a .eth domain, skipping expiration check", name)
+		return false, nil
+	}
+
+	// Extract the label (part before .eth)
+	label := strings.TrimSuffix(name, ".eth")
```
pragmaxim, line 1664: Imho `CheckENSExpiration` uses the full label before `.eth` (e.g., `sub.vitalik`), but the Base Registrar only tracks second-level `.eth` names. Valid subdomains will therefore be treated as expired and rejected.

server/public.go
```diff
@@ -1025,6 +1025,28 @@ func (s *PublicServer) explorerSearch(w http.ResponseWriter, r *http.Request) (t
...
+			}); ok {
+				expired, err := ensResolver.CheckENSExpiration(q)
+				if err == nil && !expired {
+					ensRes, err := ensResolver.ResolveENS(q)
+					if err == nil && ensRes.Address != "" {
+						http.Redirect(w, r, joinURL("/address/", ensRes.Address), http.StatusFound)
+						return noTpl, nil, nil
+					}
+				}
+				ensRes, err := ensResolver.ResolveENS(q)
```
pragmaxim: I think that `explorerSearch` calls `ResolveENS` twice and runs the second call even when expiration fails, doubling RPC load and allowing expired names to redirect. That second call ignores the expiration result, so an expired name can still redirect if ResolveENS returns an address. It also doubles the RPC load on successful, non‑expired lookups.

Issue #1310, opened 2025-08-15
EVM: fallback to main provider when alternative provider does not work


tomasklim:
In case alternative provider does not respond or responds with errors. Broadcast tx to main provider.

Issue #1297, opened 2025-07-31
Allow sendTransaction to also accept json with param deactivating alternative RPC provider


tomasklim:
`sendTransaction` endpoint now accepts string of hex data (EVM)

I want it to accept both `string` and `json`.

If `string` is accepted, let's use alternative RPC provider, if defined in config

If `json` is accepted, let's use alternative RPC provider, unless it is disabled by param. (Feel free to call the attributes better)

```
{
  data: "010000000001019d64f0c72a0d206001decbffaa7",
  defaultProvider: true
}
```

PR #1313, opened 2025-08-19
Allow sendTransaction to also accept json with param deactivating alternative RPC provider

etimofeeva:
solves #1297 and #1310

### DIFF
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1116,12 +1116,15 @@ func (b *EthereumRPC) EthereumTypeGetEip1559Fees() (*bchain.Eip1559Fees, error) 
 }
 
 // SendRawTransaction sends raw transaction
-func (b *EthereumRPC) SendRawTransaction(hex string) (string, error) {
+func (b *EthereumRPC) SendRawTransaction(hex string, disableAlternativeRPC bool) (string, error) {
 	var txid string
 	var retErr error
 
-	if b.alternativeSendTxProvider != nil {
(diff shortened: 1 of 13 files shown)


bchain/coins/eth/ethrpc.go
```diff
@@ -1116,12 +1116,15 @@ func (b *EthereumRPC) EthereumTypeGetEip1559Fees() (*bchain.Eip1559Fees, error)
 }
 
 // SendRawTransaction sends raw transaction
-func (b *EthereumRPC) SendRawTransaction(hex string) (string, error) {
+func (b *EthereumRPC) SendRawTransaction(hex string, alternativeRPC bool) (string, error) {
 	var txid string
 	var retErr error
 
-	if b.alternativeSendTxProvider != nil {
+	if alternativeRPC != false && b.alternativeSendTxProvider != nil {
```
martinboehm, line 1123: Probably

martinboehm:
You made the flag opt-in, I think it should be opt-out flag. I.e. false (not specified) value should use alternative provider, true value will disable alternative provider.

server/public.go
```diff
@@ -1509,7 +1509,7 @@ func (s *PublicServer) apiSendTx(r *http.Request, apiVersion int) (interface{},
 		}
 	}
 	if len(hex) > 0 {
-		res.Result, err = s.chain.SendRawTransaction(hex)
+		res.Result, err = s.chain.SendRawTransaction(hex, true)
```
martinboehm: Is this intentional? We use as default false elsewhere...

etimofeeva: no it wasn't and you're right, I've corrected it

Issue #1296, opened 2025-07-31
EVM: Use alternative RPC provider url for simulation and nonce


tomasklim:
Request `eth_getTransactionCount` should go to alternative RPC provider (probably part of `getAccountInfo`)
Request `eth_estimateGas` should go to alternative RPC provider (probably `estimateFee` in BlockBook slang)
Request `eth_call` should go to alternative RPC provider (probably `rpcCall` in BlockBook slang)

As txs are broadcasted to alternative RPC provider, it should use it also for simulation and nonce.

I think there is no need to duplicate URL in config. Those methods should always point to alternative RPC provider if defined

PR #1316, opened 2025-08-21
Use alternative RPC provider url for simulation and nonce

etimofeeva:
Solves #1296

### DIFF
--- a/bchain/coins/eth/contract.go
+++ b/bchain/coins/eth/contract.go
@@ -275,9 +275,6 @@ func contractGetTransfersFromTx(tx *bchain.RpcTransaction) (bchain.TokenTransfer
 
 // EthereumTypeRpcCall calls eth_call with given data and to address
 func (b *EthereumRPC) EthereumTypeRpcCall(data, to, from string) (string, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), b.Timeout)
-	defer cancel()
-	var r string
 	args := map[string]interface{}{
 		"data": data,
 		"to":   to,
@@ -285,6 +282,22 @@ func (b *EthereumRPC) EthereumTypeRpcCall(data, to, from string) (string, error)
 	if from != "" {
 		args["from"] = from
 	}
+
+	if b.alternativeSendTxProvider != nil {
+		result, err := b.alternativeSendTxProvider.callHttpStringResult(
+			b.alternativeSendTxProvider.urls[0],
+			"eth_call",
+			args,
+			"latest",
+		)
+		if err == nil {
+			return result, nil
+		}
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), b.Timeout)
+	defer cancel()
+	var r string
 	err := b.RPC.CallContext(ctx, &r, "eth_call", args, "latest")
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


bchain/coins/eth/contract.go
```diff
@@ -275,6 +275,25 @@ func contractGetTransfersFromTx(tx *bchain.RpcTransaction) (bchain.TokenTransfer
 
 // EthereumTypeRpcCall calls eth_call with given data and to address
 func (b *EthereumRPC) EthereumTypeRpcCall(data, to, from string) (string, error) {
+	if b.alternativeSendTxProvider != nil {
+		args := map[string]interface{}{
```
martinboehm: Do not repeat construction of `args` here and on the line 300, do it once before the `if b.alternativeSendTxProvider != nil`

bchain/coins/eth/ethrpc.go
```diff
@@ -985,6 +985,21 @@ func (b *EthereumRPC) GetMempoolTransactions() ([]string, error) {
 
 // EstimateFee returns fee estimation
 func (b *EthereumRPC) EstimateFee(blocks int) (big.Int, error) {
+	if b.alternativeSendTxProvider != nil {
```
martinboehm, line 993: I would move the `alternativeSendTxProvider` part of code to `EstimateSmartFee` method. Both are publicly accessible and if somebody calls directly `EstimateSmartFee`, they will bypass the condition.

bchain/coins/eth/ethrpc.go
```diff
@@ -1170,9 +1196,33 @@ func (b *EthereumRPC) EthereumTypeGetBalance(addrDesc bchain.AddressDescriptor)
...
-	return b.Client.NonceAt(ctx, addrDesc, nil)
+
+	if b.alternativeSendTxProvider != nil {
+		result, err := b.alternativeSendTxProvider.callHttpStringResult(
+			b.alternativeSendTxProvider.urls[0],
+			"eth_getTransactionCount",
+			addrDesc,
+			"pending",
+		)
+		if err == nil {
```
martinboehm: The decoding of the nonce can be done only once, after the data are returned from one or the other backend.

bchain/coins/eth/contract.go
etimofeeva: done

bchain/coins/eth/ethrpc.go
etimofeeva: done

bchain/coins/eth/ethrpc.go
etimofeeva: resolved

bchain/coins/eth/ethrpc.go
```diff
@@ -1170,9 +1195,36 @@ func (b *EthereumRPC) EthereumTypeGetBalance(addrDesc bchain.AddressDescriptor)
...
+	var err error
+
+	if b.alternativeSendTxProvider != nil {
+		result, err = b.alternativeSendTxProvider.callHttpStringResult(
+			b.alternativeSendTxProvider.urls[0],
+			"eth_getTransactionCount",
+			addrDesc,
+			"pending",
+		)
+		if err != nil {
```
martinboehm, line 1209: I would at least log error here. If there is some issue, we will not know at all that it is not using the alternative provider. Also, it would be better to avoid duplicate  calls of `b.callRpcStringResult("eth_getTransactionCount", addrDesc, "pending")`. Maybe the code could be structured that you first try to use the alternative provider (if enabled) and then if there is a `result`, the non alternative way is skipped, otherwise it is used.

etimofeeva: resolved, thank you for the review

PR #1319, opened 2025-08-26
refactor: use the built-in max/min to simplify the code

wmypku:
Inspired by https://github.com/trezor/blockbook/pull/1222 and replace all.

Use the built-in max/min from the standard library in Go 1.21 to simplify the code. https://pkg.go.dev/builtin@go1.21.0#max


wmypku:
friendly ping~

etimofeeva (approved):

PR #1325, opened 2025-09-04
Resolve Arbitrum, Base, and Zcash build errors

emersonian:
Builds for the following coins are failing for me on Debian 13 running Docker 28.4.0: (built using "make all-(coin)")

- Arbitrum (+ archive)
- Arbitrum Nova (+ archive)
- Base (+ archive)
- Base Op-Node (+ archive)
- Zcash

Two errors are showing upon build attempt: inability to connect to the Docker socket, and an error "basename: missing operand".

Upon inspection of the coin configs, I noticed that the common denominator between these coins is that they use "docker cp" within their extract_command.

This patch fixes their builds by mounting the Docker socket within builds that use the deb-% target, enabling docker-in-docker. It also resolves the basename error, caused by BinaryURL being empty in some cases.


emersonian:
About to put in a PR for Zebra 3.0.0 - @martinboehm can you check your email real quick to see if you saw my message? Also having some ETH sync issues and have an idea, Thank you.

PR #1337, opened 2025-09-29
prysm (+testnets) 6.0.4 -> 6.1.0

f7b:


martinboehm (changes requested):
No more Holesky? As long as it is not removed, it should be updated.

PR #1370, opened 2025-11-23
Enhance CoinGecko Integration with Plan-Aware API Handling

RampantDespair:
This PR adds support for a **`plan`** parameter to the CoinGecko downloader, enabling correct API endpoint selection and header usage for **free**, **demo**, and **pro** plans.

### Key Changes

* Added `plan` field to the `Coingecko` struct.
* Updated `NewCoinGeckoDownloader` to accept a `plan` argument.
* Added plan-aware logic for:

  * Selecting the correct API base URL (`pro` vs. `free`).
  * Setting the appropriate API key header (`x-cg-pro-api-key` vs. `x-cg-demo-api-key`).
* Extended `fiat_rates.go` config params to accept and forward `plan`.
* Updated all internal request calls to pass `cg.plan`.

### Reference

CoinGecko API header requirements depend on plan type:
[https://docs.coingecko.com/docs/setting-up-your-api-key#2-making-api-request](https://docs.coingecko.com/docs/setting-up-your-api-key#2-making-api-request)

### Summary

This update ensures that the downloader behaves correctly depending on whether a Pro or Demo/Free API key is used, aligning with CoinGecko’s documented API requirements.

### DIFF
--- a/fiat/coingecko.go
+++ b/fiat/coingecko.go
@@ -80,7 +80,9 @@ func NewCoinGeckoDownloader(db *db.RocksDB, network string, url string, coin str
 			freeURL = "https://api.coingecko.com/api/v3"
 		)
 
-		if apiKey != "" && (plan == "pro" || plan == "") {
+		plan = strings.ToLower(strings.TrimSpace(plan))
+
+		if apiKey != "" && plan != "free" {
 			url = proURL
 		} else {
 			url = freeURL
@@ -103,6 +105,7 @@ func NewCoinGeckoDownloader(db *db.RocksDB, network string, url string, coin str
 		db:              db,
 		throttlingDelay: time.Duration(throttlingDelayMs) * time.Millisecond,
 		metrics:         metrics,
+		plan:            plan,
 	}
 }
 
--- a/fiat/coingecko.go
+++ b/fiat/coingecko.go
@@ -38,6 +38,7 @@ type Coingecko struct {
 	updatingCurrent     bool
 	updatingTokens      bool
 	metrics             *common.Metrics
+	plan			    string
 }
 
 // simpleSupportedVSCurrencies https://api.coingecko.com/api/v3/simple/supported_vs_currencies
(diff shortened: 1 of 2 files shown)


RampantDespair:
Need any additional information or amendments to my PR?

pragmaxim (changes requested):
Hi, thank you for contributing, just a question, are you sure it works with the Pro plan? 

Imho `plan` is never stored on `Coingecko`, so `cg.plan` stays empty and every request with an API key uses the demo header. This breaks Pro keys and creates a URL/header  

Suggested fix:
  ```go
  plan = strings.ToLower(strings.TrimSpace(plan))
  if plan == "" && apiKey != "" {
      plan = "pro"
  }
  ...
  return &Coingecko{
      ...
      plan: plan,
  }
  ```

Also keeping it lower case is safer : 
```go
plan = strings.ToLower(strings.TrimSpace(plan))
```

RampantDespair:
@pragmaxim 

I normalize the plan value:

```go
plan = strings.ToLower(strings.TrimSpace(plan))
```

and then select the API endpoint as:

```go
if apiKey != "" && plan != "free" {
    url = proURL
} else {
    url = freeURL
}
```

With this logic:

* If a **Pro API key** is configured and `plan` is **unset or empty**, `plan != "free"` evaluates to `true`, so:

  * the **Pro endpoint** is used
  * the **Pro header** (`x-cg-pro-api-key`) is sent

* The behavior only changes when `plan` is **explicitly set to `"free"`**, in which case:

  * the public API endpoint is used
  * the demo/free header is sent

So Pro users are **not broken by default**. Existing setups behave exactly as before unless the user deliberately opts into the free plan. Lower-casing and trimming simply make the config more robust.

If you prefer making the default explicit for readability (e.g. mapping empty → `pro`), I can add that, but functionally the current logic already behaves correctly.

pragmaxim (approved):
LGTM

RampantDespair:
> We are accumulating some PRs into a release, so it is around the corner ... still testing that on all supported chains

@pragmaxim 
Is there some sort of estimate as to when this new release will materialize?

PR #1376, opened 2025-12-30
feat: added avalanche priority fees and corrected existing ones

etimofeeva:
Needed to add Infura priority fees for Avalanche and adjust existing ones due to the credit amount. 

Used this example for fee calculation: 
https://github.com/trezor/trezor-suite/issues/17781


pragmaxim (commented):
If I understand this correctly, Avalanche is now configured to use Infura's gas API. If Infura does not support chain ID 43114, the alternative fee provider will never return data. Because `EthereumTypeGetEip1559Fees` short-circuits to the alternative provider when configured, this results in `nil` EIP-1559 fees instead of falling back to the on-chain estimator.
   - https://github.com/trezor/blockbook/blob/infura-priority-fees/configs/coins/avalanche.json#L56
   - https://github.com/trezor/blockbook/blob/infura-priority-fees/configs/coins/avalanche_archive.json#L58
   - https://github.com/trezor/blockbook/blob/infura-priority-fees/bchain/coins/eth/ethrpc.go#L1067

pragmaxim (approved):
LGTM!

pragmaxim:
@copilot resolve the merge conflicts in this pull request

PR #1395, opened 2026-01-18
Short-circuit address parsing on hot path during block processing 

pragmaxim:
### Summary Of Changes
- Speed up padded address parsing in `addressFromPaddedHex` by decoding the last 20 bytes directly when the input is a 32-byte padded topic or calldata slot, and fall back to big.Int only for odd formats.
- Add unit tests that cover padded, unpadded, and invalid inputs for `addressFromPaddedHex`.

### addressFromPaddedHex: What Was Wrong And How It Was Fixed
- Problem: The original implementation always parsed the full hex string into a big.Int. For topics and calldata, addresses are 32-byte padded, so most of that work was wasted and done on every transfer/log.
- Fix: Strip optional `0x`, fast-path decode the last 40 hex characters (20 bytes) via `hex.DecodeString` and `BytesToAddress`, and fall back to the original big.Int parse only when the fast path does not apply or fails.

### DIFF
--- a/bchain/coins/eth/contract.go
+++ b/bchain/coins/eth/contract.go
@@ -2,6 +2,7 @@ package eth
 
 import (
 	"context"
+	"encoding/hex"
 	"math/big"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cranycrane:
follow-up at https://github.com/trezor/blockbook/pull/1779

PR #1389, opened 2026-01-13
Zcash: 3.0.0 -> 3.1.0 zebra

etimofeeva:
Upgraded zebra to 3.1.0 version.
Updated `configs/coins/zcash.json`:
- Changed `exec_command_template` from zcash.conf to zcash.toml
- Changed `server_config_file` from zcash.conf to zcash.toml.

PR #1402, opened 2026-01-23
Zcash: Upgrade to zebra v4.1.0

emersonian:
Zcash is about to deprecate zcashd, replacing it with their Rust rewrite, "zebra". The reason for so many recent version updates is that their team is finalizing the deprecation process, making sure that zebra is compatible.

Zebra v4.0.0 fixed the .toml config file name issue referenced in [#1389](https://github.com/trezor/blockbook/pull/1389), now config files [can have any file name](https://github.com/ZcashFoundation/zebra/pull/10222).


emersonian:
Updated to zebra v4.1.0, we should get this merged within the next week if possible to give people time to upgrade, see the above comment. Thank you. @pragmaxim

pragmaxim (approved):
LGTM

PR #1414, opened 2026-02-02
Add GitHub Actions CI workflow for unit tests

etimofeeva:


pragmaxim:
@etimofeeva you need to rebase it on the branch fix-closing-rocksdb-on-shutdown-signal which has connectivity tests done the same way as unit test or integration tests, you simply change the word `integration` for `connectivity`

See https://github.com/trezor/blockbook/blob/fix-closing-rocksdb-on-shutdown-signal/docs/testing.md

there is : 


    make test – run unit tests only (note that make deb* and make all* commands always run also test target)
    make test-connectivity – run connectivity checks only
    make test-integration – run integration tests only
    make test-all – run all tests above

PR #1458, opened 2026-04-08
Remove string.ToLower normalization for token fiat rates

### LINKED ISSUE #1457: Tron token fiat rates fail due to lowercased addresses
TRC20 fiat rates appear to be broken for Tron tokens because Blockbook treats token identifiers as case-insensitive, but Tron contract addresses are Base58 and case-sensitive.

##  Observed behavior
For a Tron token such as USDT: `TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t`

Coingecko returnsthe canonical mixed-case Tron address. However, Blockbook lowercases the token parameter before fiat lookup, so the lookup uses: `tr7nhqjekqxgtci8q8zy4pl8otszgjlj6t`

That does not match the stored token key, so fiat rates are reported as missing.

## Evidence
Coingecko simple/token_price/tron also returns the canonical mixed-case key even when queried with a lowercased address:

```
curl -s "https://api.coingecko.com/api/v3/simple/token_price/tron?contract_addresses=tr7nhqjekqxgtci8q8zy4pl8otszgjlj6t&vs_currencies=trx,usd,eur" | jq
  {
    "TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t": {
      "usd": 0.999944,
      "eur": 0.855742
    }
  }
```

Suspected cause in code

- Token rates are stored under the Coingecko-returned token key without normalization (`coingecko.go`)
- Fiat API lowercases token query params in `public.go`
- `websocket.go` does the same:
- Token lookup lowercases the requested token again (`currencyrateticker.go`)

## Why this is a bug
This logic is fine for hex EVM addresses, but not for Tron Base58 addresses. Tron token contract addresses are case-sensitive and should not be lowercased.

## Expected behavior

Blockbook should use chain-aware token normalization for fiat lookups:

- Ethereum-like hex addresses: case-insensitive canonicalization is fine.
- Tron Base58 addresses: preserve canonical case and do not lowercase them.

## Acceptance criteria

- /api/v2/tickers?...&token=TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t returns fiat rates for Tron USDT
- WebSocket fiat token lookup works for Tron tokens
- Existing Ethereum token fiat lookups continue to work
- Add Tron-specific fiat-rate tests to prevent regression

cranycrane:
Closes #1457 

Add Tron fiat-rate regression tests and fix token lookup so Base58 Tron contract addresses are not lowercased, while hex EVM tokens still support lowercase normalization.


etimofeeva (approved):

pragmaxim:
@cranycrane can you please check the tests are fast? because they seem to be complex, loading Database ...

I had a routine  that for each PR  I was quite sure that the newly added tests are proportional or adequate to the rest, these new tests here should take no more than a second, otherwise it will keep bloating and soon we would have Ci/CD running for an hour

pragmaxim (approved):
LGTM!

PR #1476, opened 2026-04-21
Improving pagination defaults for GetBlock in websockets

pragmaxim:
We are getting security alerts due to strange pagination default for 1M transactions.

Even though the real limit is given by # transactions in a block, it should be probably done properly and mostly united with Rest interface. Rest and WS are currently different.


etimofeeva (approved):

cranycrane (approved):
LGTM

PR #1485, opened 2026-04-29
1484 add metric to alternative fee providers

### LINKED ISSUE #1484: Add metric to alternative fee providers
The motivation is such errors as 404 or others do not stay unseen and can be investigated when they happen.

cranycrane:


pragmaxim (approved):
LGTM

etimofeeva (approved):

PR #1488, opened 2026-04-30
Improve monitoring of Consensus node

pragmaxim:
Blockbook unfortunatelly has dependency on Consensus node just for obtaining INFO whether it is LIVE.
This is good to know however it introduces huge DevOps maintenance. This PR turns it into a sweet spot : 
 - we get information for local Consensus node (production)
 - we ignore it if it is not running locally and set `"unreachable-locally"` on Dev

This reduces maintenance overhead


cranycrane (approved):
LGTM

PR #1489, opened 2026-05-01
Retry unfinalized RPC responses

pragmaxim:
Map Avalanche "cannot query unfinalized data" RPC errors to ErrBlockNotFound instead of swallowing them. This prevents failed trace calls from being treated as successful empty results, which caused trace length mismatches during sync.


cranycrane (approved):
LGTM

PR #1490, opened 2026-05-01
fix(stuck-syncing): bound retry loops that could stall chain-tip resync

pragmaxim:
- Parallel/bulk coordinators: persistent GetBlockHash failures spun forever; now escalate to errResync once the tip has fallen below the height.
- getBlockWorker: mid-range non-retryable errors looped until the hash queue closed; now bounded at RecheckThreshold, then restart on reorg or surface the error to the coordinator.


pragmaxim:
Closing in favor of #1491

PR #1491, opened 2026-05-02
Bound retry loops and detect reorgs to prevent sync stalls

pragmaxim:
Refactored the sync coordinator and workers to handle persistent backend failures and chain reorgs more robustly. Previously, the sync process could get stuck in infinite retry loops if the chain tip moved or a block became temporarily unavailable.

- Abstract sync logic into helpers (getBlockHashForSync, waitForBlockWorkers) to reduce duplication between parallel and bulk sync modes.
- Implement deadlock prevention by using non-blocking selects on channel sends to workers.
- Add reorg detection that triggers a sync restart (errResync) when the requested height is found to be past the current chain tip.
- Enhance signal handling to ensure clean exits during worker synchronization.

**This solution won over #1490**

### DIFF
--- a/bchain/coins/eth/ethrpc_blockhash_test.go
+++ b/bchain/coins/eth/ethrpc_blockhash_test.go
@@ -1,0 +1,92 @@
+package eth
+
+import (
+	"context"
+	stdErrors "errors"
+	"math/big"
+	"testing"
+	"time"
+
+	ethereum "github.com/ethereum/go-ethereum"
+	"github.com/trezor/blockbook/bchain"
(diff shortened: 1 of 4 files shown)


cranycrane (approved):
LGTM

pragmaxim:
I am closing this because we are indefinitely trying to fix quicknode's loadbalancing behavior

pragmaxim:
@copilot Can you please improve the PR description based on the commits in this branch? So it reflects the changes necessary for fixing the reorg problem ?

PR #1495, opened 2026-05-04
change "platformvscurrency" from "eth" to "usd"

### LINKED ISSUE #1494: Use "usd" instead of "eth" in "platformVsCurrency"
https://satoshilabs.slack.com/archives/C08BYBXB2LC/p1777532764646319

cranycrane:
This change is **breaking** for all deployed eth-backends and it is needed to run the `reset_eth_token_rates.sh` after the change to wipe the `eth` token rates in the RocksDB.


pragmaxim (approved):
LGTM

PR #1499, opened 2026-05-05
perf(ws): skip mempool tx loading for AccountDetailsBasic

pragmaxim:
## Summary

For `details=basic` (the default level for WebSocket `getAccountInfo`), `GetAddress` previously loaded the full body of every mempool transaction touching the address only to compute `unconfirmedBalance` / `unconfirmedSending` / `unconfirmedReceiving` — aggregated mempool-amount fields that `details=basic` consumers do not read. This change drops the per-tx fetch and reports `unconfirmedTxs` from the cheap mempool-index length only. The aggregated amount fields are omitted from the response (`omitempty` added to `UnconfirmedBalanceSat`; the sibling `unconfirmedSending` / `unconfirmedReceiving` already had it). Legacy v1 REST keeps emitting `"unconfirmedBalance": "0"` via a nil-safe fallback in `AddressToV1`.

## Performance savings

The hot path before this change, in `api/worker.go` `GetAddress`:

```go
for _, txid := range txm {                                    // |txm| = mempool txs touching this address
    tx, err := w.getTransaction(txid, false, true, addresses) // full tx fetch + decode
    // ... compute uBalReceiving / uBalSending per address
}
```

That is `O(N)` `getTransaction` calls per request, where `N` is the count of mempool entries indexed under the address descriptor. After this change, `details=basic` performs **one** `getAddressTxids(addrDesc, true, …)` mempool-index lookup and returns its length as `unconfirmedTxs` — `O(N) → O(1)` on the mempool-aggregation step.

This matters in two places where `details=basic` is the actual hot path:

1. **WebSocket `getAccountInfo`** — `details` defaults to `basic` when the client omits it (`server/websocket.go:1014`). Every WS account-info call without an explicit detail level benefits.
2. **Trezor Suite's account-refresh thunk** — the first-phase outdated check at `accountsThunks.ts:104` always uses `details: 'basic'` and only reads the count via `isAccountOutdated`. Every Suite refresh tick against an active account gets cheaper.

Higher detail levels (`tokens`, `tokenBalances`, `txids`, `txslight`, `txs`) are unchanged — they may need per-tx data for the response payload itself, so the aggregation is essentially free there.

## Compatibility

- **Trezor Suite** — audited every `details: 'basic'` caller in suite (refresh thunk, send-form recipient validation, MCP nonce/sequence auto-fill, Solana ATA check, Solana send): none reads `unconfirmedBalance`. The transformer in `blockchain-link-utils/blockbook.ts:407–411` already handles the field being absent (`isNaN()` short-circuit yields `availableBalance = balance`). No suite change required.
   -  Two trigger points in Suite that fire fetchAndUpdateAccountThunk (which starts with details: 'basic'):
  1. `accountsMiddleware.ts:14–25` — when an account becomes 'loaded' (user navigates to it in the sidebar), if its cached timestamp is older than 10 s, fire the basic refresh thunk.
  2. `blockchainThunks.ts:251–255` (syncAccountsThunk) — periodically, for all visible accounts in the network: parallel-dispatch fetchAndUpdateAccountThunk per account. This runs immediately when blockchain state initializes after app boot,  then on a per-network interval thereafter.
- **REST v1** — `AddressToV1` emits `"unconfirmedBalance": "0"` for the nil case, so the v1 wire format is unchanged.
- **REST v2 with `?details=basic`** — `unconfirmedBalance` / `unconfirmedSending` / `unconfirmedReceiving` are omitted from the JSON; `unconfirmedTxs` is the raw mempool index size (may transiently include entries that have just been confirmed but not yet evicted from the mempool; today's loop filters those by re-checking `tx.Confirmations == 0`, so the count is approximate by 0–few entries).
- **WebSocket `getAccountInfo`** — same as REST v2 `?details=basic`, and this is the WS default. Documented under "Notes for `getAccountInfo`" in `docs/api.md`.
- **xpub** -   Only one production caller hits xpub-basic: accountsThunks.ts:104 (the refresh thunk):
```
  details: account.networkType === 'solana' ? 'txids' : 'basic',
```

## Test plan

- [x] `go build ./api/... ./server/...` clean
- [x] `go vet ./api/... ./server/...` clean
- [x] `go test ./...` clean (api, btc, eth, tron, fourbyte, common)
- [x] Manual: hit `/api/v2/address/<addr>?details=basic` against a node and confirm `unconfirmedBalance` is absent from the JSON
- [x] Manual: hit `/api/v2/address/<addr>?details=txids` against the same node and confirm `unconfirmedBalance` is still present
- [x] Manual: WS `getAccountInfo` without `details` set — confirm `unconfirmedBalance` is absent and `unconfirmedTxs` is populated
- [x] Manual: hit legacy `/api/v1/address/<addr>?details=basic` and confirm `unconfirmedBalance: "0"` is still emitted (string)

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -423,7 +423,7 @@ type Address struct {
 	BalanceSat            *Amount             `json:"balance" ts_doc:"Current confirmed balance (in satoshi or base units)."`
 	TotalReceivedSat      *Amount             `json:"totalReceived,omitempty" ts_doc:"Total amount ever received by this address."`
 	TotalSentSat          *Amount             `json:"totalSent,omitempty" ts_doc:"Total amount ever sent by this address."`
-	UnconfirmedBalanceSat *Amount             `json:"unconfirmedBalance" ts_doc:"Unconfirmed balance for this address."`
+	UnconfirmedBalanceSat *Amount             `json:"unconfirmedBalance,omitempty" ts_doc:"Unconfirmed balance for this address. Omitted for AccountDetailsBasic, where mempool transactions are not aggregated."`
 	UnconfirmedTxs        int                 `json:"unconfirmedTxs" ts_doc:"Number of unconfirmed transactions for this address."`
 	UnconfirmedSending    *Amount             `json:"unconfirmedSending,omitempty" ts_doc:"Unconfirmed outgoing balance for this address."`
 	UnconfirmedReceiving  *Amount             `json:"unconfirmedReceiving,omitempty" ts_doc:"Unconfirmed incoming balance for this address."`
--- a/api/typesv1.go
+++ b/api/typesv1.go
@@ -178,6 +178,12 @@ func (w *Worker) transactionsToV1(txs []*Tx) []*TxV1 {
 // AddressToV1 converts Address to AddressV1
 func (w *Worker) AddressToV1(a *Address) *AddressV1 {
 	d := w.chainParser.AmountDecimals()
+	// v1 always serializes UnconfirmedBalance as a decimal string;
+	// when the v2 field is omitted (AccountDetailsBasic), preserve "0".
+	unconfirmedBalance := "0"
+	if a.UnconfirmedBalanceSat != nil {
+		unconfirmedBalance = a.UnconfirmedBalanceSat.DecimalString(d)
+	}
 	return &AddressV1{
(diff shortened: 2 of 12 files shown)


cranycrane (approved):
LGTM

PR #1500, opened 2026-05-05
Decrease mempool timeout for MEV-protected coins

pragmaxim:
Alternative/private relays (MEV protection) often expire pending transactions an they keep hanging in blockbook mempool for hours.

Extending infura fees period


cranycrane (approved):
LGTM

PR #1508, opened 2026-05-07
Deduplicate address descriptors before counting and fix matching function

pragmaxim:
the matching function currently matches any address subscriber, including mempool-only subscribers, and only filters `newBlockTxs` later during send: [server/websocket.go](/home/lisak/src/trezor/blockbook/server/websocket.go:1736), [server/websocket.go](/home/lisak/src/trezor/blockbook/server/websocket.go:1712). That means one `newBlockTxs=true` client can cause confirmed-block conversion work for addresses subscribed by normal Suite clients.

1. **Fix matching so confirmed-block processing only matches `publishNewBlockTxs` subscribers.**
   Add a `newBlockTxsOnly` mode or separate helper/map, so `publishNewBlockTxsByAddr` does not treat mempool-only subscribers as matches. This reduces expensive tx conversion and Ethereum receipt RPCs without changing public behavior.

2. **Deduplicate address descriptors before counting.**
   Right now duplicate addresses in one `newBlockTxs=true` subscribe can inflate `newBlockTxsSubscriptionCount`, and unsubscribe may only decrement once. That can leave the count permanently above zero and keep per-block scanning enabled. Dedupe in `unmarshalAddresses` or before `subscribeAddresses` mutates state.

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -1367,15 +1367,31 @@ func (s *WebsocketServer) unmarshalAddresses(params []byte) ([]string, bool, err
 	if len(r.Addresses) > limit {
 		return nil, false, api.NewAPIError("addresses max "+strconv.Itoa(limit), true)
 	}
-	rv := make([]string, len(r.Addresses))
-	for i, a := range r.Addresses {
+	rv := make([]string, 0, len(r.Addresses))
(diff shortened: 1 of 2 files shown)


cranycrane (approved):
LGTM

PR #1512, opened 2026-05-08
Chain tip syncing on websocket connection

pragmaxim:
Since quicknode is loadbalancing requests, we need to sync the chain-tip on a single WS connection that we use to receive the `newHeads` event, consequently all the following rpc calls :
```
  eth_getBlockByNumber("latest", false)
  eth_getBlockByNumber(height, false)
  eth_getBlockByHash(hash, true)
  eth_getLogs({ fromBlock: height, toBlock: height })
```
That are currently `https` would happen on that particular WS connection...

That would solve a very unpleasant situation when we receive `newHeads` about new block from Node A and fetching the block returns 404 because the following requests were received by Node B which is behind.

All the remaining rpc calls like should remain HTTP as they are now and therefore be loadbalanced by quicknode's LB : 
 - Historical sync / bulk sync / catch-up
 - Trezor Suite client WebSocket/Rest requests (there may be some subscriptions, but unrelated to syncing)

### DIFF
--- a/bchain/coins/avalanche/evm.go
+++ b/bchain/coins/avalanche/evm.go
@@ -48,36 +48,8 @@ type AvalancheRPCClient struct {
 	*rpc.Client
 }
 
-// AvalancheDualRPCClient routes calls and subscriptions to separate RPC clients.
-type AvalancheDualRPCClient struct {
-	CallClient *AvalancheRPCClient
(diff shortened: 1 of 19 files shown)


pragmaxim:
Closing in favor of https://github.com/trezor/blockbook/pull/1513

PR #1513, opened 2026-05-08
Split RPC traffic between HTTP pool and a WebSocket connection

pragmaxim:
## Summary

Routes Ethereum RPC traffic across two transports based on which call site issues it. The `newHeads` follow-up fetches now ride the same WebSocket connection that delivered the event, while bulk catch-up and public-API traffic continue to fan out across the HTTPS pool.

## Routing today

| Call site | Transport |
|-----------|-----------|
| `newHeads` subscription, `newPendingTransactions` subscription | **WS** (unchanged) |
| `refreshBestHeaderFromChain` — the `eth_getBlockByNumber("latest")` follow-up to a `newHeads` event | **WS** |
| `db/sync.go` sequential `connectBlocks` / `getBlockChain` (per-`newHeads` tip-sync, 1–3 blocks): `eth_getBlockByHash`, `eth_getBlockByNumber`, `eth_getLogs`, `debug_traceBlockByHash` | **WS** |
| `db/sync.go` `BulkConnectBlocks` (initial catch-up) | **HTTP pool** |
| `db/sync.go` `ParallelConnectBlocks` (EVM gap ≥ 4) | **HTTP pool** |
| `db/sync.go:resyncIndex` head-of-function calls (`GetBestBlockHash`, `GetBlockHash` for fork detection) | **HTTP pool** |
| Trezor Suite REST/WS public API (`server/`, `api/worker.go`) | **HTTP pool** |
| Mempool (`GetMempoolTransactions`, `eth_getTransactionByHash` for pending, `eth_pendingTransactions`) | **HTTP pool** |
| `getBestHeader` cache-miss / startup fallback | **HTTP pool** |
| `BatchCallContext` (ERC-20 fan-out in `contract.go`) | **HTTP pool** (always) |

The WS connection is one TCP/WSS stream pinned to a single backend by the load balancer, so all four target methods follow up against the same node that announced the block — eliminating the LB-drift case where one node says block N exists and another returns null. High-volume bulk RPC continues to fan out across the HTTPS pool.

L2 forks (Optimism, Base, Arbitrum, Polygon, BSC) inherit the new behaviour because they all use `eth.OpenRPC`. Non-EVM coins (Bitcoin, Tron, …) are unaffected.

## How

- New optional `bchain.SyncableBlockChain` interface; `EthereumRPC` implements it via a thin `ethSyncView` facade.
- `SyncWorker` now carries two `BlockChain` references: `chain` (HTTP) for bulk/parallel/orchestration, `tipChain` (the sync facade) for the sequential `connectBlocks` path.
- `DualRPCClient.CallContext` and `EthereumClient.<method>` dispatch on a context-value flag set by `WithSyncRoute(ctx)`.
- `BatchCallContext` is unchanged — batching stays on the HTTP client.
- Internal helpers in `ethrpc.go` accept the caller's parent ctx so the routing flag propagates from facade entry down to every underlying RPC.

## Test plan

- [x] `go build ./... && go vet ./...` clean
- [x] `go test -tags=unittest ./...` clean
- [x] New unit tests in `bchain/coins/eth/sync_route_test.go`:
  - context-value round-trip
  - `EthereumClient.pick` dispatch (including aliased clients)
  - `DualRPCClient.CallContext` routes to `SubClient` only with `WithSyncRoute(ctx)`
  - `DualRPCClient.BatchCallContext` always uses `CallClient`
  - `ethSyncView.GetBlock` propagates `WithSyncRoute` to the underlying RPC; bare `EthereumRPC.GetBlock` does not
- [ ] End-to-end soak against QuickNode (HTTPS + WSS): confirm tip-sync RPCs land on the WS connection, bulk catch-up traffic stays on HTTPS, public REST/WS responses unaffected
- [ ] WS reconnect drill: drop the WS connection and confirm `subscribe` reconnects and tip-sync recovers

### DIFF
--- a/bchain/coins/avalanche/avalancherpc.go
+++ b/bchain/coins/avalanche/avalancherpc.go
@@ -35,6 +35,11 @@ func dialRPC(rawURL string) (*rpc.Client, error) {
 }
 
 // OpenRPC opens RPC connections for Avalanche to separate HTTP and WS endpoints.
+// Mirrors eth.OpenRPC: when distinct URLs are configured, a separate WS
+// *rpc.Client is dialed and wrapped into both an AvalancheRPCClient (for
+// custom JSON-RPC dispatch) and an *ethclient.Client (for typed Avalanche
+// methods). When the same URL serves both, the WS-side fields alias the
+// HTTP-side so Close does not double-close.
 var OpenRPC = func(httpURL, wsURL string) (bchain.EVMRPCClient, bchain.EVMClient, error) {
 	callURL, subURL, err := eth.NormalizeRPCURLs(httpURL, wsURL)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


pragmaxim:
Closing since such flexibility is not really possible without introducing high complexity.

The only possible option is having both chain tip and historical syncing on WS and keep HTTPS for REST/WS public API and mempool

PR #1514, opened 2026-05-10
Agentic local testing

pragmaxim:
I made a set of scripts that simplify local testing without docker which makes it suitable for agents to test new stuff and avoid regressions.

I would keep AGENTS.md as minimalistic as possible, Testing is essential, but I see agents simply already use go formatting themselves, there is no need to introduce needless instructions.

What I find important to state is the unknown facts that could be very useful for performance, etc.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,35 @@
+# Agent instructions
+
+## Testing
+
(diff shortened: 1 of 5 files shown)


cranycrane (approved):
LGTM

PR #1520, opened 2026-05-21
Limit fiat timestamp requests

pragmaxim:
## Summary
- Add a shared 1000-timestamp maximum for fiat-rate timestamp lookups.
- Count REST timestamp values before splitting the query parameter.
- Add a global WebSocket active request limit alongside the per-connection pending request limit.

## Why
This keeps fiat-rate batch lookups and WebSocket work bounded under large caller-provided input while preserving existing small batch behavior.

## Validation
- contrib/tests/run-unit-tests.sh -run 'Test(GetFiatRatesForTimestamps_LimitsInput|WebsocketTrackWorkAppliesGlobalLimit)$'
- contrib/tests/run-unit-tests.sh -run 'Test_PublicServer_BitcoinType$'
- contrib/tests/run-unit-tests.sh

### DIFF
--- a/api/fiat_rates_api.go
+++ b/api/fiat_rates_api.go
@@ -10,6 +10,9 @@ import (
 	"github.com/trezor/blockbook/common"
 )
 
+// MaxFiatRatesTimestamps limits batch fiat-rate lookups to bounded request work.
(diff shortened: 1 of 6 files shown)


cranycrane (approved):
LGTM

PR #1521, opened 2026-05-21
Limit xpub descriptor expansion

pragmaxim:
## Summary
- Limit xpub descriptor change-list expansion before account scans.
- Bound scan width when wider change lists are combined with large gaps.
- Cap the in-memory xpub cache and evict the least recently accessed entries.

## Details
Broad change lists previously expanded directly into scan branches, and cache eviction was based only on age. This adds explicit limits so descriptor scans and cached entries stay predictable.

## Validation
- contrib/tests/run-unit-tests.sh -run 'Test(ParseXpubDescriptors|ValidateXpubScanLimits|TrimXpubCacheItemsLocked)$' -count=1
- contrib/tests/run-unit-tests.sh -count=1

### DIFF
--- a/api/xpub.go
+++ b/api/xpub.go
@@ -22,6 +22,8 @@ const txInput = 1
 const txOutput = 2
 
 const xpubCacheExpirationSeconds = 3600
+const xpubCacheMaxEntries = 128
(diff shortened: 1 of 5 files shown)


cranycrane (approved):
LGTM

PR #1525, opened 2026-05-27
Fix NFT self-transfer reorg rollback

pragmaxim:
Fix ERC721/1155 self-transfer ownership rollback after reorg.

Probability is very low. The bug needs a pretty specific overlap:
- An ERC721 Transfer event with from == to.
- That exact block later gets disconnected in a reorg.
- The address/contract entry survives disconnect, meaning the address already had other tx history for that contract.
- The rarest part is usually the combination of self-transfer + reorg of that same block. ERC721 self-transfers are allowed and do happen, but they are not the common transfer path. Reorgs are also usually shallow and tip-local on healthy chains.


cranycrane (approved):
LGTM

PR #1530, opened 2026-05-28
fix(sync): prevent sync stalls on load-balanced EVM RPC skew

pragmaxim:
## Summary
- Make EVM sync tolerate load-balanced backend tip skew when a later best-height probe is behind an already observed block hash.
- Prevent parallel and bulk sync from wedging while queueing block hashes after workers have already reported an abort/resync condition.
- Preserve real reorg handling by restarting only when a chain-state probe proves the same height has a different hash.

## Reasoning
Polygon can hit a QuickNode load-balanced RPC pool where consecutive calls land on nodes with different heights. Blockbook may first observe the next block hash at height `N`, then ask for best height and get `N-1` from a lagging backend. The previous code treated that single lagging response as a hard `remote best height error`, which made tip syncing retry instead of progressing.

```
sync.go:258] resync: error - remote best height 87570258 less than sync start height 87570259
blockbook.go:586] syncIndexLoop /go/src/github.com/trezor/blockbook/db/sync.go:259: resync: remote best height error, will retry...
```

**There was also a separate internal stall path: the parallel/bulk coordinator used a blocking `hch <- hashHeight{...}` send. If all workers were stuck retrying `block not found`, the hash queue could fill. At that point the coordinator no longer selected on `abortCh`, so a worker could report `errResync` or another terminal error and Blockbook would not observe it. That matches the symptom where sync appears stuck internally after height inconsistency without `debug_traceBlockByHash` errors.**

## Solution
- If `GetBestBlockHeight()` returns below `startHeight`, log a warning and fall back to sequential sync for that round instead of returning a hard error.
- Make queued hash sends abort-aware via `sendHashHeight`, so `ParallelConnectBlocks` and `BulkConnectBlocks` can return worker aborts even when the hash queue is full.
- Adjust missing-block rechecks for load-balanced EVM RPCs: a lagging best-height probe or a missing hash probe no longer proves a reorg by itself. Sync restarts immediately only when a probe can serve the height with a different hash.
- In sequential tip sync, continue retrying a missing block above an observed backend height when Blockbook already has a concrete hash for that block.

## Proofs
- `TestParallelConnectBlocksReturnsWorkerAbortWhenHashQueueFull` reproduces the queue-full/coordinator-stopped-reading-abort case and asserts the sync returns `errResync` instead of hanging.
- `TestGetBlockChainRetriesKnownHashAboveObservedBestHeight` covers the load-balanced tip skew case where the block hash is known but a later best-height probe is behind.
- `TestShouldRestartSyncOnMissingBlockIgnoresLaggingBestHeight` and `TestShouldRestartSyncOnMissingBlockIgnoresMissingHashProbe` cover the QuickNode-style lagging probe behavior.
- Existing changed-hash behavior is kept by `TestGetBlockChainMissingBlockChangedHashResyncs`.

## Tests
- `contrib/tests/run-unit-tests.sh -count=1`
- `contrib/tests/run-integration-tests.sh -run 'TestIntegration/.*/connectivity' -count=1`\n- `contrib/tests/run-integration-tests.sh -run 'TestIntegration/polygon=main/rpc/GetBlock' -count=1`

### DIFF
--- a/db/sync.go
+++ b/db/sync.go
@@ -119,6 +119,14 @@ func NewSyncWorkerWithConfig(db *RocksDB, chain bchain.BlockChain, syncWorkers, 
 	if cfg != nil {
 		effectiveCfg = *cfg
 	}
+	// MaxStallDuration is the load-bearing liveness cap (see its doc): the retry
+	// loops disable the cap when it's <= 0, which would let a chain-shortening
+	// reorg spin forever. Enforce the invariant structurally here so every caller
+	// (including tests passing a partial cfg) gets a safe value, not just the
+	// ApplyMissingBlockRetryOverride path.
+	if effectiveCfg.MissingBlockRetry.MaxStallDuration <= 0 {
+		effectiveCfg.MissingBlockRetry.MaxStallDuration = DefaultMissingBlockRetryConfig().MaxStallDuration
+	}
 	return &SyncWorker{
 		db:                db,
 		chain:             chain,
--- a/db/sync.go
+++ b/db/sync.go
@@ -43,10 +43,11 @@ type MissingBlockRetryConfig struct {
 	TipRecheckThreshold int
 	// RetryDelay keeps retry pressure low while still reacting quickly to transient backend gaps.
 	RetryDelay time.Duration
-	// MaxStallDuration caps the wall-clock time a single block fetch may spend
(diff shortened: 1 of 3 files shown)


cranycrane (approved):
LGTM

PR #1552, opened 2026-06-09
xpub(fix): XSS in xpub explorer descriptor handling

pragmaxim:
Blockbook accepts xpub descriptors on the /xpub/ explorer route. The Bitcoin descriptor parser matches only a valid prefix because xpubDesriptorRegex is anchored with ^ but has no terminal $, and ParseXpub uses FindStringSubmatch. The parser validates the extracted xpub, but descriptor.XpubDescriptor and the API response keep the original user-supplied string. The xpub template then renders that string through template.JSStr inside the QR-code JavaScript expression.

### DIFF
--- a/bchain/coins/btc/bitcoinlikeparser.go
+++ b/bchain/coins/btc/bitcoinlikeparser.go
@@ -442,7 +442,7 @@ var (
 
 func init() {
 	var err error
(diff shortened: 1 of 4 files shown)


cranycrane (approved):
LGTM

PR #1553, opened 2026-06-09
Security: CI/CD hardening 

pragmaxim:
Addresses the **high**, **medium**, and the repo-code-fixable **low** severity findings from the CI/CD + `contrib/` security review. One commit per finding, each independently validated on CI and via a real dev deploy. The only remaining items require GitHub-settings/infra changes (H2, M1) or are low-value non-CI ops hygiene (the rest of L2) — listed at the bottom.

## Committed fixes

| Commit | Finding | Summary |
|---|---|---|
| `7d7d2ec5` | **H3** | De-template `branch_or_tag`/coins in `deploy.yml` so GitHub expression interpolation can't reach the shell before `validate_branch_or_tag.py` runs (passes values via env vars). |
| `2ab49a8a` | **M4** | Pin `actions/checkout`, `setup-node`, `cache`, `upload-artifact` to full commit SHAs; pin `PyYAML`; add `dependabot.yml` (github-actions + pip). |
| `e508dee9` | **H1** | `permissions: contents: read`, same-repo fork-PR guard on `unit-tests`, `persist-credentials: false` on all checkouts in `testing.yml`. |
| `a038b9a7`, `460f49e6` | **M2** | Pin/verify backend artifact sources (omotenashicoin×2, tron×2, optimism-legacy → commit pins + sha256; divi, deeponion → sha256) and add a `backend-artifact-lint` extends the lint to platform-specific artifact overrides. |
| `de70bf16`, `37f4a462` | **M3** | Verify the Go tarball sha256 (per-arch) and assert the RocksDB v9.10.0 commit in `build/docker/bin/Dockerfile`; `37f4a462` adds sha256 verification of the Docker CLI tarball. |
| `8e02227d` | **L1** | `wait_for_sync.py`: replace the unconditional `ssl._create_unverified_context()` with opt-in verification — `SYNC_CA_FILE` pins an internal CA (preferred for internal hosts), `SYNC_TLS_INSECURE=0` uses the system trust store, default stays unverified for self-signed dev certs — and log the active TLS mode so a relaxed gate is visible in run logs. |
| `c0232206` | **L2** | `contrib/scripts/deploy-dev.sh`: add `set -euo pipefail`, fix the `${UPDATE_VENDOR:0}` → `${UPDATE_VENDOR:-0}` typo, quote expansions + carry args in an array, track per-coin status (no longer masked by a later success), and use POSIX `=` instead of the `==` bashism. (Operator helper; not invoked by CI.) |

All checksums/commit pins were computed by downloading the pinned artifacts. nuls/unobtanium stay on HTTP (both already sha256-verified; neither upstream serves HTTPS) — the new lint warns rather than fails on checksum-protected HTTP.


cranycrane (approved):
lgtm

PR #1560, opened 2026-06-10
fix(fiat-rates): make CoinGecko fetching resilient to 429

### LINKED ISSUE #1559: fix: fetching historical rates should be resillient
- The fetching of historical rates for base crypto/tokens should be resillient to 429 and implement a backoff logic

cranycrane:
### Summary

Reworks CoinGecko fiat-rate fetching to be resilient and self-healing. Two problems motivated this: the Free tier is heavily rate-limited, and a blockbook instance can start with gaps in its historical rates (a long trailing gap after downtime, or interior holes). Previously the Trezor CDN was used only for from-scratch bootstrap, so any later gap was refetched through the Free tier — slowly, and competing with the chain-tip updates.

The branch (a) makes throttled fetching wait-and-retry instead of giving up, (b) routes all high-volume historical backfill to the no-rate-limit CDN while leaving only the chain tip (≤1 day) on the Free tier, and (c) adds a blocking, at-startup self-healing pass that detects and repairs missing daily rates before the periodic downloader loops start.

### Free-plan throttling resilience

- Detect throttling by HTTP 429 / `Retry-After` and Cloudflare 1015 HTML ban bodies.
- Pace free/keyless requests at 10 req/min; keep bootstrap CDN requests on the faster local interval.
- Replace capped retries with retry-until-success backoff honoring `Retry-After`, behind a shared throttle window across all downloader requests.
- Prioritize current/high-granularity requests over historical/token backfill while throttled.
- Run current-rate and historical-rate fetching in separate goroutines.
- Move list/token caches onto the downloader instance behind a mutex; make token-update state race-safe.

### CDN routing (Free tier only for the chain tip)

- Resolve a per-series fetch range from the gap: `1 day` → tip (Free tier), `2..365 days` → backfill, `> 365` → capped at 365.
- Route tip fetches to the Free/Pro endpoint and everything larger to the CDN, so bulk backfill never touches the rate-limited path.

### Startup self-healing reconciliation

A blocking pass that runs once RocksDB is initialized, before the periodic loops, so the DB is consistent and there is no Free-tier contention.

- **Stage 1 (cheap, key-only scan):** compute which whole days are missing within the 365-day window. Zero missing → returns immediately with no network requests (the healthy common case). `≥ 90` missing → reported as `gap_too_large` and skipped (a months-long gap signals a bug/misconfig, not routine catch-up).
- **Stage 2 (bounded fetch):** fetch only the missing-day span from the CDN and fill only those days (interior holes and trailing gaps), leaving existing records untouched. Write set is bounded by the missing-day count regardless of series count.
- Repairs the base coin across every vs-currency plus every token.
- Crash-safe and idempotent: partial progress is persisted, so an interrupted run resumes from the remaining gap on the next startup.
- A panic in reconciliation can never brick startup (recovered and logged).

### Cancellation and wall-clock budget

- The pass is bounded by a `context.Context` carrying both the shutdown signal (`chanOsSignal`) and a wall-clock budget, threaded through the whole fetch path.
- A `SIGTERM` mid-repair aborts promptly — even while a single request is stuck retrying a 429.
- A persistently slow/throttling CDN can no longer stall startup: once the budget elapses the pass aborts, persists what it fetched, and lets blockbook come up.
- Steady-state loops pass `context.Background()` and keep their retry-forever behavior unchanged.

### Configuration

- `reconcileHistoricalAtStartup` (bool, default enabled) — toggle the startup pass per coin.
- `reconcileHistoricalMaxMinutes` (int, default 5) — wall-clock budget for the pass.

### Observability

- New counters: `fiat_rates_fetched_units_total` and `fiat_rates_fetched_tokens_total` (by phase: tip / backfill / bootstrap / reconcile), and `fiat_rates_unable_total` (by phase and reason: `gap_too_large`, `fetch_failed`, `provider_banned`, `no_base_ticker`, `budget_exhausted`).
- New Grafana "Fiat Rates" dashboard section: chain-tip fetches, backfill fetches, reconciliation, and unable-to-fetch by reason.

### Tests

- Expanded `fiat/coingecko_test.go`: 429/Retry-After detection, request pacing, shared throttle window, priority yielding, retry-until-success historical/token updates, and bootstrap CDN pacing.
- CDN routing: range resolution and per-range source-URL selection.
- Reconciliation: interior-hole repair (base + token, present days untouched), healthy DB / young DB make no requests, `gap_too_large` skips fetching, no-CDN no-op, cancelled-context aborts before any fetch, and budget timeout aborts mid-backfill without error.

### DIFF
--- a/blockbook.go
+++ b/blockbook.go
@@ -763,8 +763,21 @@ func computeFeeStats(stopCompute chan os.Signal, blockFrom, blockTo int, db *db.
 	return err
 }
 
+// runStartupSelfHealing runs blocking, at-startup self-healing tasks once RocksDB is
+// initialized. It blocks until done so the rest of startup proceeds against a consistent DB.
+// Future self-healing tasks can be added here; today it reconciles missing historical fiat
+// rates before the periodic downloader loops start.
+func runStartupSelfHealing() {
+	if fiatRates != nil && fiatRates.Enabled {
+		fiatRates.ReconcileHistoricalRatesAtStartup()
+	}
+}
+
 func initDownloaders(db *db.RocksDB, chain bchain.BlockChain, config *common.Config) {
 	if fiatRates.Enabled {
+		// Block on startup self-healing before the periodic loops begin, so historical gaps
+		// are repaired via the CDN without contending with the Free-tier tip loop.
+		runStartupSelfHealing()
 		go fiatRates.RunDownloader()
 	}
 
--- a/blockbook.go
+++ b/blockbook.go
@@ -769,7 +769,9 @@ func computeFeeStats(stopCompute chan os.Signal, blockFrom, blockTo int, db *db.
 // rates before the periodic downloader loops start.
 func runStartupSelfHealing() {
 	if fiatRates != nil && fiatRates.Enabled {
-		fiatRates.ReconcileHistoricalRatesAtStartup()
+		// chanOsSignal is closed on shutdown, so a SIGTERM during a long reconciliation
+		// aborts it promptly instead of blocking startup-shutdown until the backfill ends.
+		fiatRates.ReconcileHistoricalRatesAtStartup(chanOsSignal)
 	}
 }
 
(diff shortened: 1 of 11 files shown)


pragmaxim (approved):
LGTM

PR #1562, opened 2026-06-10
Reconcile alternative provider mempool cache +  Expose confirmed nonce 

pragmaxim:
@cranycrane this is a follow up on the https://github.com/trezor/blockbook/pull/1500 and there is a background story https://gist.github.com/pragmaxim/7bcd54ac8bc7f5b72b8264c29e8bee33

This PR has two parts: (1) reconciliation of the alternative-provider EVM send-tx mempool cache, and (2) exposing the confirmed account nonce on the address API.

### Part 1 — Alternative-provider mempool cache reconciliation

Reconciles the alternative-provider EVM send-tx mempool cache, which Blockbook keeps itself because private / Blink-style relays do not expose a mempool list to sync against. No API/wire change in this part.

#### Liveness reconciliation (periodic recheck with `eth_getTransactionByHash`)
- remove local pending entries when all alternative providers return empty, or when a provider reports the tx as mined
- keep txs on provider / network errors and when any configured provider still knows the tx
- skip a freshly submitted tx for one check period (a load-balanced provider node may transiently not know it yet)
- document why tx removal by txid was not chosen and how this follows PR #1500

#### Nonce-superseded healing (deterministic cleanup of permanently-stuck txs)
- for a tx the provider still reports as *pending*, fetch the confirmed account nonce via `eth_getTransactionCount(from, "latest")` and evict it when its nonce is strictly below that count — the nonce is already spent on-chain, so the tx can never be mined (e.g. a replacement submitted outside Blockbook, or a wallet that picked a nonce below the account's current nonce)
- scoped to consumed nonces only: `nonce == count` is the next mineable tx and is kept; `nonce > count` is a gap that may still be filled, so it is kept and left to the timeout (no gap eviction — removing a legitimately pending tx would be worse than the bug being fixed)
- evicted via the same path as the mined / provider-missing removals (both Blockbook mempool and cache)

Safety of the nonce check:
- the confirmed nonce is the **lowest** value across all configured providers, so a lagging or misbehaving provider cannot evict a still-mineable tx — eviction requires unanimous agreement that the nonce is spent
- on any lookup failure the tx is kept; the timeout remains the safety net
- lookups are memoized per sender, so each sender is queried at most once per reconcile cycle
- `"latest"` carries the usual chain-tip reorg caveat (documented in code); the exposure matches the existing mined-tx removal and is bounded — eviction only drops the cache entry, and a still-valid tx is re-indexed when actually mined

#### Observability
- every reconciliation decision is counted under `blockbook_eth_alternative_mempool_reconciliation_events_total{action=...}`, with the new `nonce_superseded` action alongside `skipped_fresh`, `kept`, `provider_error`, `provider_missing`, `mined` and `timeout`
- surfaced in the Grafana "Alternative mempool reconciliation" panel

#### Endpoint change
The address response gains a new optional field `confirmedNonce` for Ethereum-like (EVM + Tron) addresses, **gated behind an opt-in flag**:
- REST: `GET /api/v2/address/<address>?confirmedNonce=true`
- WebSocket: `getAccountInfo` request gains a `confirmedNonce: true` boolean

The two nonce values:
- `nonce` (unchanged, always returned) — the **pending** nonce from `eth_getTransactionCount(addr, "pending")`; counts mempool txs, i.e. the next nonce the account will use
- `confirmedNonce` (new, only when opted in) — the **confirmed** nonce from `eth_getTransactionCount(addr, "latest")`; reflects only mined txs. Equals `nonce` when the account has nothing pending. **Omitted entirely when the flag is not set.**

#### Why gated
The confirmed nonce requires a **second** `eth_getTransactionCount("latest")` backend call, which most callers do not need. Gating keeps the default cost unchanged (one `pending` call, exactly as before) and only pays for the extra call when a client explicitly asks. Consumers that need the last confirmed / minted nonce (e.g. trezor-suite) previously had to infer it from the most recent confirmed tx — fragile with gaps, replacements, or externally-submitted txs.

#### Implementation
- new query param / WS field map to `AddressFilter.WithConfirmedNonce`, which drives a `withConfirmed bool` on `BlockChain.EthereumTypeGetNonce` → `EthereumTypeGetNonces(addrDesc, withConfirmed)`
- **flag off (default):** only the `pending` nonce is fetched (single call), confirmed value is unused and the field is omitted — identical to pre-feature behavior, no second RPC call
- **flag on:** `pending` + `latest` are fetched in a single JSON-RPC **batch** round-trip (so the confirmed value adds no extra latency beyond the one batched request); falls back to two sequential calls when the RPC client does not support batching. The alternative-provider path batches the same way and falls back to the primary RPC on failure
- **confirmed nonce is best-effort:** `EthereumTypeGetNonces` returns `(pending, confirmed, confirmedOK, err)`. The pending nonce is required (error is fatal); a failure of *only* the `latest` lookup is logged and surfaced as `confirmedOK=false`, so the worker omits `confirmedNonce` instead of failing the whole account response. This means an opted-in client cannot have a transient `latest` failure take down balance/txs/tokens
- Tron exposes only the latest nonce via a single `NonceAt` call, so it returns the same value for both and ignores the flag

#### Compatibility
- additive and optional (`omitempty`) — no breaking change; existing `nonce` semantics are untouched, and the default response is byte-for-byte unchanged
- `openapi.yaml` (the `confirmedNonce` query parameter + the `WsAccountInfoReq.confirmedNonce` property + the response field) and the generated `blockbook-api.ts` are updated in lockstep and verified by the OpenAPI parity typecheck. `blockbook-api.ts` was hand-edited rather than regenerated, because the typescriptify generator is currently not reproducible — tracked in #1566
- trezor-suite tolerates the extra field (no runtime schema rejection); it must sync its generated types and pass the new flag to consume `confirmedNonce`, but no coordinated release is required — the two can ship independently in any order

### Tests
- `contrib/tests/run-unit-tests.sh`
- `bchain/coins/eth` unit tests cover dropped / fresh / known / mined / provider-error reconciliation and the nonce paths: superseded removal, current-nonce kept, future/gap-nonce kept, lookup-failure kept, unparsable nonces kept, lowest-confirmed-nonce-across-providers, and per-sender memoization (all green under `-race`)
- `server` ethereum-type and Tron address tests cover both the gated-on path (REST `?confirmedNonce=true` and the WebSocket `confirmedNonce` flag → field present) and the gated-off default (field absent)
- `bchain/coins/eth` unit tests for `EthereumTypeGetNonces`: gated-off fetches pending only (no `latest` call), gated-on batched success, confirmed-failure best-effort (pending returned, no error), pending-failure and batch-transport-failure fatal, and the sequential fallback incl. its best-effort path
- OpenAPI **e2e** over both transports (enabled for EVM coins in `tests/tests.json`): `GetAddressConfirmedNonceEVM` (REST) and `WsGetAccountInfoConfirmedNonceEVM` (WebSocket `getAccountInfo`) each assert the opt-in gate (no `confirmedNonce` without the flag) and, when opted in, validate the field and `confirmed <= pending`, skipping gracefully on backends where the feature isn't deployed yet
- OpenAPI: `lint:spec` + parity typecheck (`tests/openapi`) confirm `blockbook-api.ts` ↔ `openapi.yaml` stay in sync


cranycrane (commented):
Starting review...

cranycrane (commented):
## 📋 Review Summary

The PR implements a reconciliation mechanism for alternative-provider mempool transactions. It periodically checks the state of cached transactions via `eth_getTransactionByHash` and removes them if they are mined, dropped by all providers, or timed out.

## 🔍 General Feedback

- Solid unit test coverage for the new reconciliation logic.
- Proper use of `sync.Once` and `sync.Mutex` for concurrency safety.
- **Critical:** The documentation file `docs/tx-removal-decision.md` linked in the README is missing from the PR.

cranycrane (commented):
## 📋 Review Summary

The PR implements a reconciliation mechanism for alternative-provider mempool transactions. It periodically checks the state of cached transactions via `eth_getTransactionByHash` and removes them if they are mined, dropped by all providers, or timed out.

## 🔍 General Feedback

- Solid unit test coverage for the new reconciliation logic.
- Proper use of `sync.Once` and `sync.Mutex` for concurrency safety.
- **Critical:** The documentation file `docs/tx-removal-decision.md` linked in the README is missing from the PR.

## 📝 Inline Observations

- **docs/README.md (line 14):** [MEDIUM] The file `docs/tx-removal-decision.md` is linked here but is missing from the repository/PR. This will result in a broken link.
- **bchain/coins/eth/alternativesendtx.go (line 231):** [LOW] Checking `tx == (bchain.RpcTransaction{})` to determine if the transaction was found might be brittle. It's safer to check if the hash is present:
  ```go
  if tx.Hash == "" {
  ```

cranycrane (approved):
LGTM

PR #1563, opened 2026-06-10
feat: rate-limit WebSocket and REST API traffic

pragmaxim:
## What

Adds application-level abuse protection for public WebSocket and REST API traffic, with shared and spoof-safe client-IP attribution.

- WebSocket: per-connection message-rate limit and temporary per-IP blocklist for clearly abusive traffic.
- REST API: middleware around public `/api` routes with per-client request-rate, burst, concurrency, and optional temporary block limits.
- Shared client-IP config: REST and WebSocket now use the same proxy and Cloudflare attribution rules via `<NET>_TRUSTED_PROXIES` and `<NET>_CLOUDFLARE_IPS`.
- Compatibility: existing `<NET>_WS_TRUSTED_PROXIES` and `<NET>_WS_CLOUDFLARE_IPS` are still accepted as legacy fallbacks when the shared variables are unset.
- Metrics/docs: added REST limiter metrics and documented the shared attribution variables plus REST policy variables.
- Balance history: bounded per-request work — a `balancehistory` query over a heavy address or xpub is rejected past a per-transaction cap instead of running an unbounded DB scan. The cap is **split by transport**: `<NET>_WS_BALANCE_HISTORY_MAX_TXS` (default `1000000`) for the WebSocket `getBalanceHistory` method and `<NET>_REST_BALANCE_HISTORY_MAX_TXS` (default `250000`) for the REST endpoint, with `<NET>_BALANCE_HISTORY_MAX_TXS` kept as a backward-compatible shared fallback. Trezor Suite uses WebSocket only and requests an account's full history for its balance graph (and never derives the displayed balance from it), so the WS surface gets generous headroom while the open REST surface stays tight. The xpub path range-limits and caps its txid load *before* the limit check, so a narrower `from`/`to` actually reduces the work and can avoid rejection (previously it loaded full history regardless).
- Balance-history observability: new `blockbook_balance_history_txs` histogram (transactions aggregated per request — the work the cap bounds) and `blockbook_balance_history_cap_exceeded_total` counter, both labelled by `transport` (ws/rest) and `path` (address/xpub), plus WebSocket and REST Grafana panels (per-request tx quantiles + cap-rejection rate) to size the caps from real traffic.

## Why

The existing WebSocket limiter work protected upgraded connections, but repeated public REST API calls could still drive backend RPC/RocksDB work with no application-level per-client throttle. REST needed the same trustworthy client identity as WebSockets, but not a shared limiter state machine: REST limits request starts and in-flight handlers, while WebSocket limits connections and per-connection message volume.

During review, the REST proxy envs were also duplicating WebSocket proxy config. Those settings describe deployment topology, so they are now shared; only the protocol-specific policy knobs remain separate.

The balance-history cap was originally a single value shared by both transports. Because Trezor Suite (paying customers) reaches Blockbook over WebSocket only and asks for an account's full history to draw its balance graph, a tight shared cap risked breaking a heavy wallet's graph while a generous shared cap would have weakened DoS protection on the open REST surface. Splitting the cap by transport lets each surface be sized for its own threat model; the new metrics exist so the WS default can be confirmed (or adjusted) against real traffic instead of a guess.

## Configuration

Shared client attribution:

| Variable | Default | Meaning |
|---|---:|---|
| `<NET>_TRUSTED_PROXIES` | unset | Extra trusted CIDRs whose `X-Real-Ip` may be used; loopback/private/link-local peers are always trusted |
| `<NET>_CLOUDFLARE_IPS` | `builtin` | Cloudflare/CDN peer ranges used to verify `CF-Connecting-*`; accepts `builtin`, custom CIDRs, or `off` |

WebSocket policy:

| Variable | Default | Meaning |
|---|---:|---|
| `<NET>_WS_MESSAGE_RATE_LIMIT` | `2500` | max messages per trailing window; `0` disables |
| `<NET>_WS_MESSAGE_RATE_WINDOW` | `10m` | trailing message-rate window |
| `<NET>_WS_IP_BLOCK_DURATION` | `12h` | block duration after message-rate abuse; `0` disables blocking |

REST policy:

| Variable | Default | Meaning |
|---|---:|---|
| `<NET>_REST_RATE_LIMIT` | `600` | request starts per window; `0` disables request-rate limiting |
| `<NET>_REST_RATE_WINDOW` | `1m` | token-bucket refill window |
| `<NET>_REST_BURST` | `120` | token-bucket burst |
| `<NET>_REST_MAX_CONCURRENT` | `24` | in-flight public REST API requests per client key; `0` disables |
| `<NET>_REST_STATE_TTL` | `10m` | idle limiter state retention |
| `<NET>_REST_BLOCK_DURATION` | `0` | optional temporary block after repeated breaches |

Balance-history request cap (per transport):

| Variable | Default | Meaning |
|---|---:|---|
| `<NET>_WS_BALANCE_HISTORY_MAX_TXS` | `1000000` | max transactions aggregated per WebSocket `getBalanceHistory` request (address or xpub); `0` disables |
| `<NET>_REST_BALANCE_HISTORY_MAX_TXS` | `250000` | max transactions aggregated per REST `balancehistory` request (address or xpub); `0` disables |
| `<NET>_BALANCE_HISTORY_MAX_TXS` | unset | backward-compatible shared fallback that sets the default for both of the above; a transport-specific variable overrides it for its surface |

## Safety

- `CF-Connecting-*` is trusted only from verified Cloudflare/CDN peers unless verification is explicitly disabled.
- `X-Real-Ip` is trusted only from loopback/private/link-local peers or configured trusted proxy CIDRs.
- IPv6 client keys aggregate to `/64`.
- Auto-blocking refuses loopback/private/link-local, trusted proxy, and Cloudflare/CDN infrastructure keys.
- REST rejections return `429` before request body reads or handler work.
- `balancehistory` bounds its DB work to the transport-specific cap before aggregating and rejects rather than truncating, so a returned current balance is never affected — only the historical graph is gated. The xpub path loads each derived address's txids within the requested height range, capped at the limit, instead of the full history. A public xpub cap error is surfaced directly on the WebSocket path instead of being masked by the address-parse fallback, matching REST.

## Validation

- `rtk go test ./server -run 'TestRestAPI|TestClientIP|TestResolveClientIP|TestRateLimitKey|TestIsBlockableKey|TestConnMessageRate|TestParseCloudflareProxies|TestParseTrustedProxies' -count=1`
- `rtk go test ./api ./common ./server -count=1` (balance-history cap split, metric YAML↔struct invariants)
- `rtk contrib/tests/run-unit-tests.sh`
- `rtk python3 contrib/scripts/render_grafana.py --check`
- `rtk contrib/tests/run-integration-tests.sh -run 'TestIntegration/.*/connectivity' -count=1`
- `rtk contrib/tests/run-integration-tests.sh -run 'TestIntegration/ethereum=main' -count=1`
- `rtk bash -lc 'source contrib/gh-vars.sh && bb_export_gh_vars && OPENAPI_COINS=ethereum contrib/tests/run-openapi-tests.sh'`


cranycrane (approved):
LGTM

PR #1564, opened 2026-06-11
tests: system check for meeting conditions for running tests locally

pragmaxim:
This improves local development experience as AGENTS.md is instructed to run tests locally for development but if the developer does not have environment in check, it should be clear about what is wrong.


cranycrane (approved):
LGTM

PR #1574, opened 2026-06-19
feat(deploy): runtime env via optional EnvironmentFile

pragmaxim:
Gives Blockbook's runtime env vars (`INFURA_API_KEY`, per-coin tunables) a per-host channel so they no longer live in hand-edited `/etc/systemd/system.conf`.

**What**
- Blockbook's systemd unit gains an optional `EnvironmentFile=-/etc/blockbook/blockbook.env`. The leading `-` keeps it optional: public/community installs and hosts still using `DefaultEnvironment` are unaffected, and a unit-level value overrides `DefaultEnvironment` per key — so prod can migrate with no flag-day.
- The dev deploy path (`deploy.yml`) materializes the file from the `BB_BLOCKBOOK_ENV` GitHub secret before restart. It's a no-op when the secret is unset, so it's safe to merge before the secret exists. Prod hosts render the same file via bbctl.
- Folds the build-time `BB_*` variable-prefix list — previously duplicated across the Makefile, the `export-env-vars` action, and `templates.go` — into one canonical `build/bb-build-var-prefixes.txt`, guarded by a Go test so drift fails in CI instead of silently at build time.
- failing fast now if secret is expected and the underlying env var is not present which was not the case now !!

**Verified**
- Template tests pass; generated unit carries the line before `ExecStart`.
- Makefile forwarding output is byte-for-byte identical to the old regex (GNU Make 4.3); `export-env-vars` normalization unchanged.


cranycrane (approved):
LGTM

PR #1578, opened 2026-06-21
fix(tron): add polling-only tip sync fallback when ZeroMQ is disabled

h45hc47:
## Background

A Tron node can only emit events through **one** channel at a time — it either publishes to the ZeroMQ native queue **or** writes via the MongoDB event plugin, not both. This is decided once at plugin-load time in the node itself:

```java
// framework/src/main/java/org/tron/common/logsfilter/EventPluginLoader.java
if (config.isUseNativeQueue()) {
    return launchNativeQueue(config); // ZMQ path
}
return launchEventPlugin(config);     // MongoDB plugin path
```

So when an operator runs the node with the MongoDB plugin instead of the native queue, ZeroMQ is simply not available to Blockbook, and there's no way to have both.

## Symptom

Before this change, a Tron deployment running the node with the MongoDB event plugin (i.e. no working ZeroMQ feed) had two bad outcomes depending on configuration:

- **`message_queue_binding` non-empty but pointing at a dead/non-working instance:** no ZeroMQ messages ever arrive, so the only thing that advances the index is the hardcoded `resyncIndexPeriodMs` fallback (935093 ms ≈ 15.6 min). The tip effectively only moves about once every ~15 minutes.
- **`message_queue_binding` empty:** the backend doesn't degrade gracefully — it crashes during mempool initialization when it tries to set up the ZeroMQ subscription against an empty binding.

Neither path gives a usable, near-real-time tip when the node is on the MongoDB plugin path.

## Fix

- Add a **polling-only mode** keyed on `b.mq == nil`. In `tipWatchdog`, poll at the chain's block cadence (`AverageBlockTimeMs`) instead of `threshold/3`, with a 1s floor so a misconfigured `AverageBlockTimeMs` can't hammer the RPC.
- In `tipWatchdogTick`, when `b.mq == nil`, bypass the `lastNotifyNs` staleness gate entirely and drive sync directly from the ticker via `refreshBestHeaderFromChain`, pushing `NewBlock`/`NewTx` on advance — so the tip no longer depends on the ~15 min `resyncIndexPeriodMs` fallback.
- Deliberately **do not** arm `lastNotifyNs` in polling-only mode — that field is reserved for real ZeroMQ deliveries, so a genuinely dead ZeroMQ feed still surfaces via `subscription_age_seconds` rather than being masked.
- `InitializeMempool` now logs a warning and returns early (skipping ZeroMQ subscription) when `message_queue_binding` is empty, instead of crashing on init against an empty binding.

## Scope

Only affects Tron deployments running the node with the MongoDB event plugin (no ZeroMQ binding). Deployments using the native ZeroMQ queue take the unchanged `threshold/3` path and the original staleness-gated tick.

## Tests

`bchain/coins/tron` watchdog tests added:

- `TestTronTipWatchdogTickPollingOnlyDoesNotArmLiveness` — poll must not arm liveness.
- `TestTronTipWatchdogTickPollingOnlyImmediate` — poll fires immediately without waiting on the stale threshold or an armed `lastNs`.


cranycrane (approved):
LGTM, thanks for the contribution!

PR #1579, opened 2026-06-21
fix(api): always serialize token decimals, default unresolved to coin convention

### LINKED ISSUE #1577: Token transfers: `decimals` is sometimes missing — make it required or document the default
### Summary

Token transfers returned by blockbook (in address / transaction responses) sometimes **omit the `decimals` field** for tokens that the backend doesn't fully know. This forces clients to guess how to format the transferred amount, and there is no documented contract for what `decimals` means when it is absent.

Decimals are hacked here https://github.com/trezor/blockbook/blob/cfa737479d1cf6099fb2c663dac437de0eb64f87/configs/contract-fix/ethereum.json

### Why it's a problem

In Trezor Suite, a missing `decimals` was defaulted to `0`, which makes the UI render raw integer amounts (e.g. a transfer of `1000000000000000000` shown as `1,000,000,000,000,000,000` instead of `1`). Per the ERC‑20 standard the `decimals()` method is **optional**, and the conventional default is **18** (see [OpenZeppelin ERC20 docs](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20#ERC20-decimals--)). So neither `0` nor "unknown" is a safe assumption on the client side.

### Request

Decide and document the contract for `decimals` on token transfers, ideally one of:

1. **Make `decimals` required** — always populate it for token transfers (preferred), e.g. fall back to `18` when the token contract doesn't expose `decimals()`, matching the ERC‑20 convention.
2. Or, if it can genuinely be unknown, **document that it may be absent** and specify the recommended client behavior, so all clients handle it consistently.

### Context

- Trezor Suite PR working around this: https://github.com/trezor/trezor-suite/pull/28697
- Affects EVM-like token transfers where `decimals` is not present in the blockbook response.

pragmaxim:
## Problem

Closes #1577.

Blockbook's token-transfer and token-balance responses declared `decimals`
with `,omitempty`, so a value of `0` was dropped from the JSON entirely.
Clients (e.g. Trezor Suite) could not tell a genuine 0-decimal token apart
from missing metadata and had to assume a default — which is unsafe (`0`
underflows the amount, `18` overflows it). Trezor Suite worked around this
client-side in trezor/trezor-suite#28697 by defaulting missing decimals to 18.

This implements the issue's preferred fix: **always populate `decimals`,
defaulting to the coin convention (18 for ERC-20) when the on-chain value is
unavailable.**

## Changes

- Remove `,omitempty` from `api.Token.Decimals` and `api.TokenTransfer.Decimals`
  so the field is always present (consistent with `ContractInfoResult.Decimals`,
  which already is).
- `getContractDescriptorInfo`: when a contract's metadata cannot be read
  on-chain (a transient RPC error, or a token that does not expose
  `name`/`symbol`/`decimals`), fall back to `chainParser.AmountDecimals()`
  instead of surfacing — and persisting — an ambiguous `0`. The fallback is
  guarded on `UnhandledTokenStandard` (the sync-time sentinel), so genuinely
  0-decimal tokens, which always carry a resolved (handled) standard, keep
  their `0`.
- Keep `blockbook-api.ts` and `openapi.yaml` in parity: `decimals` is now
  required on the `Token` and `TokenTransfer` schemas. The OpenAPI parity
  typecheck (`tests/openapi`) passes.
- Add a regression test asserting `decimals` is serialized even when `0`.

### Why the second bullet matters

Removing `,omitempty` alone would have been a subtle regression: in the
unresolved-metadata path the sync-time `Decimals` is `0`, which `omitempty`
currently hides. Serving it as an authoritative `"decimals":0` would make
strict clients render amounts off by 10^18. The fallback ensures `0` only ever
appears for tokens whose precision is genuinely 0.

## Notes for clients

`decimals` is now authoritative and always present, so consumers no longer need
to guess. Note that genuine 0-decimal tokens now correctly arrive as
`"decimals":0` instead of being coerced to a default — clients using
`value || 18`-style fallbacks should prefer `value ?? 18` (or just trust the
wire value) to avoid mis-rendering them.

## Testing

- `go test -tags unittest ./api/ ./server/ ./db/ ./bchain/coins/eth/ ./bchain/coins/tron/` — all pass.
- `cd tests/openapi && npm run lint:spec && npm run generate && npm run typecheck` — all pass.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -235,7 +235,7 @@ type Token struct {
 	Contract         string                   `json:"contract,omitempty" ts_doc:"Contract address on-chain."`
 	Transfers        int                      `json:"transfers" ts_doc:"Total number of token transfers for this address."`
 	Symbol           string                   `json:"symbol,omitempty" ts_doc:"Symbol for the token (e.g., 'ETH', 'USDT')."`
-	Decimals         int                      `json:"decimals,omitempty" ts_doc:"Number of decimals for this token."`
+	Decimals         int                      `json:"decimals" ts_doc:"Number of decimals for this token. Always present; defaults to the coin convention (18 for ERC-20) when the contract value is unavailable."`
 	BalanceSat       *Amount                  `json:"balance,omitempty" ts_doc:"Current token balance (in minimal base units)."`
 	BaseValue        float64                  `json:"baseValue,omitempty" ts_doc:"Value in the base currency (e.g. ETH for ERC20 tokens)."`
 	SecondaryValue   float64                  `json:"secondaryValue,omitempty" ts_doc:"Value in a secondary currency (e.g. fiat), if available."`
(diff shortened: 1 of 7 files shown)


cranycrane (approved):
LGTM

PR #1582, opened 2026-06-22
1581 add upper cap limits to chosen websocket endpoints to increase resilliency

### LINKED ISSUE #1581: Add upper-cap limits to chosen Websocket endpoints to increase resilliency

cranycrane:
## Summary

Adds upper-cap / bound-checking limits to several server endpoints to improve resilience against unbounded memory growth and abusive request patterns (closes #1581). Each cap is covered by unit tests.

## Changes

**WebSocket — cap in-flight `getMempoolFilters` responses**
Per-connection semaphore (`maxWebsocketMempoolFiltersResponses = 4`) bounds each request from acceptance until its response is written (or drained on close), capping the peak memory held in large filter responses; excess requests are rejected with a `mempool_filters_limit` error. Adds a `release`/`finalize` hook on `WsRes` so the slot is freed exactly once across write, drain-on-close, and overflow paths.

**Fiat rates — bound and normalize currency selectors**
Replaces `removeEmpty` with `normalizeFiatCurrencies` (trim, lowercase, dedupe, bound) applied once at the entry points (`GetCurrentFiatRates`, `GetFiatRatesForTimestamps`, `GetBalanceHistory`, `GetXpubBalanceHistory`) before any per-result maps are allocated; downstream helpers rely on the pre-normalized slice. New limits `MaxFiatRatesCurrencies = 128` (applied to the raw selector list length) and `MaxFiatRatesCurrencyCodeLength = 16` return a client-facing error when exceeded.

**EVM — harden ERC-1155 `TransferBatch` log parsing**
Rewrites offset/length handling with explicit bounds and overflow checks to avoid possible OOM / out-of-range slicing from malformed log data, validating total data length before allocating the result slice.

## Behavior changes

- A currency selector longer than 16 characters is now rejected with a public `APIError` (HTTP 400 on REST). The REST tickers endpoint never splits `?currency=` on commas — it passes the whole value as a single selector — so a comma-joined value is only rejected once it exceeds 16 chars. `?currency=usd,eur,gbp,jpy` (15 chars) therefore still returns `200` with `{"usd,eur,gbp,jpy": -1}` exactly as before; only a longer value such as `?currency=usd,eur,gbp,cad,jpy` (19 chars) now returns `400 "currency code too long, max 16"`.
- The currency-count cap (128) applies to the raw selector list length before deduplication, so the validation work itself is bounded by the cap; a longer list is rejected even if it would deduplicate to fewer entries. This cap only bites on the array-accepting WS methods (`getCurrentFiatRates`, `getFiatRatesForTimestamps`) and the balance-history paths; the single-param REST tickers endpoint always passes 0 or 1 selector, so it can never trip the count cap.
- More than 4 getMempoolFilters requests in flight on one websocket connection (accepted but not yet written back) are rejected with a `mempool_filters_limit` error.

## Tests
- `server/websocket_test.go` — mempool filters slot capping, including release after a successful write (via `outputLoop` over a live connection) and drain-on-close
- `api/fiat_rates_api_test.go`, `api/fiat_balance_history_test.go` — currency normalization/limits
- `bchain/coins/eth/contract_test.go` — malformed ERC-1155 batch logs

### DIFF
--- a/api/fiat_balance_history.go
+++ b/api/fiat_balance_history.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -13,14 +12,6 @@ func normalizeBalanceHistoryPathLabel(pathLabel string) string {
 		return "unknown"
 	}
 	return pathLabel
-}
-
-func normalizeCurrenciesToLowercase(currencies []string) []string {
-	currenciesLowercase := make([]string, len(currencies))
-	for i := range currencies {
-		currenciesLowercase[i] = strings.ToLower(currencies[i])
-	}
-	return currenciesLowercase
 }
 
 func buildBalanceHistoryTimestamps(histories BalanceHistories) []int64 {
@@ -180,7 +171,10 @@ func (w *Worker) setFiatRateToBalanceHistories(histories BalanceHistories, curre
 		return nil
 	}
 	pathLabel = normalizeBalanceHistoryPathLabel(pathLabel)
-	currenciesLowercase := normalizeCurrenciesToLowercase(currencies)
+	currenciesLowercase, err := normalizeFiatCurrencies(currencies)
+	if err != nil {
+		return err
+	}
 	timestamps := buildBalanceHistoryTimestamps(histories)
 	tickers, batchFetchValid, reason, returnedTickers, err := w.lookupBalanceHistoryBatchTickers(timestamps, pathLabel, len(histories))
 	if batchFetchValid {
--- a/api/fiat_balance_history.go
+++ b/api/fiat_balance_history.go
@@ -166,23 +166,21 @@ func (w *Worker) applyFallbackTickersToBalanceHistories(histories BalanceHistori
 	w.observeBalanceHistoryFiatDuration(pathLabel, "fallback", stats.status(), fallbackStarted)
 }
 
-func (w *Worker) setFiatRateToBalanceHistories(histories BalanceHistories, currencies []string, pathLabel string) error {
+// setFiatRateToBalanceHistories fills FiatRates of the passed histories.
+// currenciesLowercase must already be normalized by normalizeFiatCurrencies;
+// both entry points (GetBalanceHistory, GetXpubBalanceHistory) do so.
+func (w *Worker) setFiatRateToBalanceHistories(histories BalanceHistories, currenciesLowercase []string, pathLabel string) {
 	if len(histories) == 0 || w.fiatRates == nil || !w.fiatRates.Enabled {
-		return nil
+		return
 	}
 	pathLabel = normalizeBalanceHistoryPathLabel(pathLabel)
-	currenciesLowercase, err := normalizeFiatCurrencies(currencies)
-	if err != nil {
-		return err
-	}
 	timestamps := buildBalanceHistoryTimestamps(histories)
 	tickers, batchFetchValid, reason, returnedTickers, err := w.lookupBalanceHistoryBatchTickers(timestamps, pathLabel, len(histories))
 	if batchFetchValid {
 		applyBatchTickersToBalanceHistories(histories, tickers, currenciesLowercase)
-		return nil
+		return
 	}
 	glog.Errorf("Error finding tickers for %d timestamps (returned %d, reason %s). Error: %v", len(timestamps), returnedTickers, reason, err)
 	w.incrementBalanceHistoryFiatFallback(pathLabel, reason)
 	w.applyFallbackTickersToBalanceHistories(histories, currenciesLowercase, pathLabel)
-	return nil
 }
--- a/api/fiat_rates_api.go
+++ b/api/fiat_rates_api.go
@@ -13,15 +13,42 @@ import (
 // MaxFiatRatesTimestamps limits batch fiat-rate lookups to bounded request work.
 const MaxFiatRatesTimestamps = 1000
 
-// removeEmpty removes empty strings from a slice.
-func removeEmpty(stringSlice []string) []string {
-	ret := make([]string, 0, len(stringSlice))
-	for _, str := range stringSlice {
-		if str != "" {
-			ret = append(ret, str)
+// MaxFiatRatesCurrencies limits explicit currency selectors. An empty list still
+// means "all available currencies", so legitimate callers do not need a large
+// explicit list.
+const MaxFiatRatesCurrencies = 128
+
+// MaxFiatRatesCurrencyCodeLength bounds each selector before it is copied into
+// per-result response maps.
+const MaxFiatRatesCurrencyCodeLength = 16
+
+// normalizeFiatCurrencies trims, lowercases, deduplicates, and bounds explicit
+// fiat currency selectors before any per-result response maps are allocated.
+func normalizeFiatCurrencies(currencies []string) ([]string, error) {
+	capacity := len(currencies)
+	if capacity > MaxFiatRatesCurrencies {
+		capacity = MaxFiatRatesCurrencies + 1
+	}
+	seen := make(map[string]struct{}, capacity)
+	normalized := make([]string, 0, capacity)
+	for _, currency := range currencies {
+		currency = strings.ToLower(strings.TrimSpace(currency))
+		if currency == "" {
+			continue
+		}
+		if len(currency) > MaxFiatRatesCurrencyCodeLength {
+			return nil, NewAPIError(fmt.Sprintf("currency code too long, max %d", MaxFiatRatesCurrencyCodeLength), true)
+		}
+		if _, found := seen[currency]; found {
+			continue
+		}
+		seen[currency] = struct{}{}
+		normalized = append(normalized, currency)
+		if len(normalized) > MaxFiatRatesCurrencies {
+			return nil, NewAPIError(fmt.Sprintf("too many currencies, max %d", MaxFiatRatesCurrencies), true)
 		}
 	}
-	return ret
+	return normalized, nil
 }
 
 func copyTickerRates(rates map[string]float32) map[string]float32 {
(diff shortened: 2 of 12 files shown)


server/websocket.go
```diff
@@ -871,6 +911,17 @@ func (s *WebsocketServer) onRequest(c *websocketChannel, req *WsReq) {
...
+		if req.Method == "getMempoolFilters" {
+			if !c.acquireMempoolFiltersSlot() {
+				e := resultError{}
+				e.Error.Message = "mempool_filters_limit"
+				data = e
+				s.metrics.WebsocketRequests.With(common.Labels{"method": methodLabel, "status": "failure"}).Inc()
+				glog.Warning("Client ", c.id, " exceeded getMempoolFilters response limit, ", c.ip)
+				return
+			}
+			release = c.releaseMempoolFiltersSlot
```
pragmaxim, line 929: The `getMempoolFilters` cap (4) bounds **concurrent compute + queue + write together**, not just "responses not yet written" as the field comment implies. The slot is acquired at the *start* of `onRequest` and released by `finalize()` only *after* the response is written by `outputLoop`'s `WriteJSON` (or drained on close). Because each request runs in its own goroutine (`inputLoop` spawns `go onRequest`, up to 48 pending), two consequences follow:

- A well-behaved client that pipelines >4 `getMempoolFilters` back-to-back during sync gets spurious `mempool_filters_limit` errors even though it reads responses promptly.
- A slow-but-still-alive client whose `WriteJSON` blocks toward the 60s write deadline pins all 4 slots for that window, rejecting every further `getMempoolFilters`.

If the intent is to bound **memory of large queued filter responses**, this is working as designed — please make the comment say so. If the intent is to bound **compute concurrency**, release the slot right after `f(s, c, req)` returns rather than at write time. Worth a deliberate decision either way.

api/fiat_rates_api.go
```diff
@@ -13,15 +13,42 @@ import (
...
+	}
+	seen := make(map[string]struct{}, capacity)
+	normalized := make([]string, 0, capacity)
+	for _, currency := range currencies {
+		currency = strings.ToLower(strings.TrimSpace(currency))
+		if currency == "" {
+			continue
+		}
+		if len(currency) > MaxFiatRatesCurrencyCodeLength {
+			return nil, NewAPIError(fmt.Sprintf("currency code too long, max %d", MaxFiatRatesCurrencyCodeLength), true)
```
pragmaxim, line 41: Behavior change worth a release note: a currency selector >16 chars now returns a public `APIError` (→ HTTP 400). The REST tickers handler (`server/public.go:1792-1795`) takes the raw `?currency=` value whole and wraps it as a single-element slice — it never splits on commas. So `?currency=usd,eur,gbp,jpy` (17 chars), which previously flowed through `removeEmpty` and returned `200` with `{"usd,eur,gbp,jpy": -1}`, now returns `400 "currency code too long, max 16"`. Arguably an improvement (comma-joined currencies were never supported), but it's a deterministic 200→400 change for any client relying on the old tolerant `-1` behavior.

api/fiat_rates_api.go
```diff
@@ -13,15 +13,42 @@ import (
...
+// normalizeFiatCurrencies trims, lowercases, deduplicates, and bounds explicit
+// fiat currency selectors before any per-result response maps are allocated.
+func normalizeFiatCurrencies(currencies []string) ([]string, error) {
+	capacity := len(currencies)
+	if capacity > MaxFiatRatesCurrencies {
+		capacity = MaxFiatRatesCurrencies + 1
+	}
+	seen := make(map[string]struct{}, capacity)
+	normalized := make([]string, 0, capacity)
+	for _, currency := range currencies {
```
pragmaxim, line 35: Minor (defense-in-depth): there's no upfront `len(currencies)` guard. The "too many currencies" check at line 47 fires only when a *unique* code is appended (after the dedup `continue` at 42-44), so an all-duplicate input (`["usd","usd",…]`) collapses to one entry and the cap never trips — the full slice is still iterated (`TrimSpace`/`ToLower`/map-lookup per element). It's bounded by the 4 MiB WS read limit (~700K entries ≈ a few ms), so low impact, but a one-line `if len(currencies) > MaxFiatRatesCurrencies { return ... }` before the loop would bound the work by raw input length and match the cap's intent.

api/fiat_balance_history.go
```diff
@@ -180,7 +171,10 @@ func (w *Worker) setFiatRateToBalanceHistories(histories BalanceHistories, curre
 		return nil
 	}
 	pathLabel = normalizeBalanceHistoryPathLabel(pathLabel)
-	currenciesLowercase := normalizeCurrenciesToLowercase(currencies)
+	currenciesLowercase, err := normalizeFiatCurrencies(currencies)
```
pragmaxim: Redundant double normalization. Both callers — `GetBalanceHistory` (api/worker.go) and `GetXpubBalanceHistory` (api/xpub.go) — already call `normalizeFiatCurrencies(currencies)` in place before invoking `setFiatRateToBalanceHistories`, which then normalizes the identical slice again here. It's idempotent (no wrong output), but it's a second trim/lowercase pass plus a throwaway `seen` map + slice on every balance-history request. Consider dropping this call and relying on the already-normalized slice from the entry points.

bchain/coins/eth/contract.go
```diff
@@ -154,40 +197,65 @@ func processERC1155TransferBatchEvent(l *bchain.RpcLog) (transfer *bchain.TokenT
...
+	if err != nil {
+		return nil, err
+	}
+	if endValues > len(data) {
+		return nil, errors.New("ERC1155 TransferBatch, invalid values data length")
+	}
+	maxInt := uint64(^uint(0) >> 1)
+	if countValues > maxInt {
+		return nil, errors.New("ERC1155 TransferBatch, invalid data length")
+	}
```
pragmaxim: Dead code: this `countValues > maxInt` guard is statically unreachable. By the time control reaches here, the `endValues > len(data)` check above already bounds `countValues` to `(len(data)-base)/64` — far below `maxInt` — so `int(countValues)` can never truncate. (The real load-bearing overflow guard is the `count > (maxInt-base)/evmWordHex` check inside `erc1155BatchArrayEnd`.) These four lines can be replaced with `count := int(countValues)`.

pragmaxim (commented):
Automated review of the cap-limit changes. The core logic is sound — the ERC-1155 `TransferBatch` bounds/overflow hardening in particular holds up well (candidate concerns about `countIds` misuse and `erc1155BatchOffsetHex` rejecting valid `<0x40` offsets were checked and refuted). Findings below are behavior-change + cleanup, no crashes.

Inline comments cover the actionable items. One extra very-minor nit not inline (the function isn't in the diff): `getFiatRatesResult` (api/fiat_rates_api.go:81 and :104) still calls `strings.ToLower` on each selector, but the slice it receives is already lowercased by `normalizeFiatCurrencies` — redundant, bounded work. (Note line 72 is *not* redundant; it iterates raw ticker-map keys.)

api/fiat_rates_api.go
```diff
@@ -13,15 +13,40 @@ import (
...
+		return nil, NewAPIError(fmt.Sprintf("too many currencies, max %d", MaxFiatRatesCurrencies), true)
+	}
+	seen := make(map[string]struct{}, len(currencies))
+	normalized := make([]string, 0, len(currencies))
+	for _, currency := range currencies {
+		currency = strings.ToLower(strings.TrimSpace(currency))
+		if currency == "" {
+			continue
 		}
+		if len(currency) > MaxFiatRatesCurrencyCodeLength {
```
pragmaxim, line 40: **PR-description inaccuracy (not a code issue).** The PR body says `?currency=usd,eur,gbp,jpy` on the REST tickers endpoint "now returns `400 \"currency code too long, max 16\"`". It doesn't: `apiTickers` (server/public.go) never comma-splits — it passes the whole query value as a single selector (`currencies = []string{currency}`). That string is 15 chars, which is `<= 16`, so this check passes and the endpoint still returns `200` with `{"usd,eur,gbp,jpy": -1}`, exactly as before. The 400 only triggers at >=17 chars (e.g. `usd,eur,gbp,cad,jpy`, 19 chars). Could you correct the example in the description?

While here: it's worth noting in the description that the count cap (128) is unreachable on this single-param endpoint (it always passes 0 or 1 selector) — the cap only bites on the array-accepting WS `getCurrentFiatRates`/`getFiatRatesForTimestamps` and balance-history paths. The code itself is correct and internally consistent; only the prose example is off.

server/websocket.go
```diff
@@ -601,6 +639,7 @@ func (s *WebsocketServer) outputLoop(c *websocketChannel) {
 	for m := range c.out {
 		c.conn.SetWriteDeadline(time.Now().Add(defaultTimeout))
 		err := c.conn.WriteJSON(m)
+		c.finalize(m)
```
pragmaxim, line 642: **Test-coverage gap (optional).** This `c.finalize(m)` after a successful `WriteJSON` is the primary slot-release path for a live connection, but no test exercises it. `TestWebsocketMempoolFilterResponseSlots` covers only the raw semaphore, and `TestWebsocketCloseOutReleasesQueuedMempoolFilterResponses` covers only the enqueue-then-`CloseOut` drain.

If this line were ever removed, every successful `getMempoolFilters` response would leak a slot and the connection would be permanently locked out after 4 responses — yet both new unit tests *and* the existing e2e suite (which sends only 2 such requests against a cap of 4 and asserts payloads only) would still pass, so the regression would ship undetected.

Consider a test that drives a release-bearing `WsRes` through `outputLoop` (or `onRequest` end-to-end) and asserts `len(c.mempoolFiltersSlots) == 0` after the write, plus an end-to-end check that a 5th in-flight request gets `mempool_filters_limit`.

pragmaxim (commented):
Two nit-level comments below. The rest of the review found no correctness or security issues — the websocket slot lifecycle and the ERC-1155 bounds math both hold up, all tests pass with `-tags unittest`, and `go vet` is clean.

pragmaxim (approved):

PR #1583, opened 2026-06-22
fix(tron): the receipt for unsolidified block should not be set to ni…

### LINKED ISSUE #1576: Disperency in token transfers on Tron
When accessing `/api/v2/address/TXy5qqpAJNykdqsMU2dZM7B7mZbVF2izAN`, we does not receive data on the TRC20 USDT transfer and also cannot see the balance of this token. Example response:

```
{
    "page": 1,
    "totalPages": 1,
    "itemsOnPage": 1000,
    "address": "TXy5qqpAJNykdqsMU2dZM7B7mZbVF2izAN",
    "balance": "30000000",
    "unconfirmedBalance": "0",
    "unconfirmedTxs": 0,
    "txs": 2,
    "nonTokenTxs": 2,
    "txids": [
        "755ff6cf885cb1454733f12c3f7429270dac9fd0d117a40fabdd13ea2d1da9f7",
        "9145c1d9294e78cc3f4203f1531bb15135f7df267ef5c617c5cd3284b7d4449c"
    ],
    "nonce": "0",
    "chainExtraData": {
        "payloadType": "tron",
        "payload": {
            "availableStakedBandwidth": 0,
            "totalStakedBandwidth": 0,
            "availableFreeBandwidth": 600,
            "totalFreeBandwidth": 600,
            "availableEnergy": 0,
            "totalEnergy": 0,
            "stakingInfo": {
                "stakedBalance": "0",
                "stakedBalanceEnergy": "0",
                "stakedBalanceBandwidth": "0",
                "unstakingBatches": [],
                "totalVotingPower": "0",
                "availableVotingPower": "0",
                "votes": [],
                "unclaimedReward": "0",
                "delegatedBalanceEnergy": "0",
                "delegatedBalanceBandwidth": "0"
            }
        }
    }
}
```

However, there is a USDT TRC20 transfer and balance at this address (link).

This information can also be obtained by querying the blockchain; by calling `/api/v2/address/TXy5qqpAJNykdqsMU2dZM7B7mZbVF2izAN?details=txs&contract=TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t`. Example response:

```
{
    "address": "TXy5qqpAJNykdqsMU2dZM7B7mZbVF2izAN",
    "balance": "30000000",
    "unconfirmedBalance": "0",
    "unconfirmedTxs": 0,
    "txs": 2,
    "nonTokenTxs": 2,
    "nonce": "0",
    "tokens": [
        {
            "type": "TRC20",
            "standard": "TRC20",
            "name": "Tether USD",
            "contract": "TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t",
            "transfers": 0,
            "symbol": "USDT",
            "decimals": 6,
            "balance": "760000000"
        }
    ],
    "chainExtraData": {
        "payloadType": "tron",
        "payload": {
            "availableStakedBandwidth": 0,
            "totalStakedBandwidth": 0,
            "availableFreeBandwidth": 600,
            "totalFreeBandwidth": 600,
            "availableEnergy": 0,
            "totalEnergy": 0,
            "stakingInfo": {
                "stakedBalance": "0",
                "stakedBalanceEnergy": "0",
                "stakedBalanceBandwidth": "0",
                "unstakingBatches": [],
                "totalVotingPower": "0",
                "availableVotingPower": "0",
                "votes": [],
                "unclaimedReward": "0",
                "delegatedBalanceEnergy": "0",
                "delegatedBalanceBandwidth": "0"
            }
        }
    }
}
```

Information about this transfer is also available by referring to `/api/v2/tx/b7a97862b0c719b714f0cf7e250ebc2dcdf1e5f05c54ddb21ff3a748c1aa45e4`

We are building the Blockbook from commit 29c73c2 on the master branch. Could you please clarify whether this behavior is expected?

cranycrane:
Closes #1576.

## Problem

Token transfers on Tron were missing from address/contract indexes for transactions in not-yet-solidified blocks. During indexing, `GetBlock` dropped the receipt (and its event logs) for unsolidified blocks. Since Tron derives token transfers from receipt logs, the transfer participants never reached the indexes until the block solidified and was re-processed.

## Fix

- Renamed `buildTxFromHTTPData`'s `isSolidified` param → `keepReceipt` to name what it controls.
- `GetBlock` now always keeps the receipt, so logs are retained for indexing regardless of solidity.
- `GetTransaction`/mempool paths are behavior-unchanged (still pending until solid).

## Tests

`TestTronBuildTxFromHTTPData_KeepReceiptControlsTokenLogs`: `keepReceipt=false` → nil receipt, no transfers; `keepReceipt=true` → receipt/logs retained and TRC-20 transfer decoded correctly.

### DIFF
--- a/bchain/coins/tron/tronrpc.go
+++ b/bchain/coins/tron/tronrpc.go
@@ -347,21 +347,6 @@ func (b *TronRPC) GetBlockInfo(hash string) (*bchain.BlockInfo, error) {
 }
 
 func (b *TronRPC) getBestHeader() (bchain.EVMHeader, error) {
-	// During initial sync (before ZeroMQ is initialized) there is no push-based
-	// tip refresh, so always read the latest header from the backend.
-	if b.mq == nil {
-		_, err := b.refreshBestHeaderFromChain()
-		if err != nil {
-			return nil, err
-		}
-		b.bestHeaderLock.Lock()
-		defer b.bestHeaderLock.Unlock()
-		if b.bestHeader == nil || b.bestHeader.Number() == nil {
-			return nil, errors.New("best header is nil")
-		}
-		return b.bestHeader, nil
-	}
-
 	b.bestHeaderLock.Lock()
 	cachedHeader := b.bestHeader
 	cachedAt := b.bestHeaderTime
@@ -384,22 +369,6 @@ func (b *TronRPC) getBestHeader() (bchain.EVMHeader, error) {
 	return b.bestHeader, nil
 }
 
-// setBestHeader stores h as the cached tip and reports whether it changed.
-//
-// Unlike EthereumRPC.setBestHeader this is intentionally NON-monotonic: a lower
-// height is accepted. Tron's tip is never taken from the feed header (the ZeroMQ
-// notification carries none) — it is always an HTTP re-query (refreshBestHeaderFromChain),
-// refreshed on every notification and on a tronBestHeaderMaxAge timer, so the cache
-// is meant to track whatever the backend currently reports. Accepting a lower height
-// is what lets a genuine rollback surface immediately to resyncIndex, so Tron is not
-// subject to the frozen-tip masking that the EVM monotonic guard introduces (and which
-// EVM has to undo with a watchdog regress).
-//
-// Tradeoff: with a load-balanced Tron RPC, a single lagging node answering a re-query
-// could regress the tip and trip a spurious fork in resyncIndex (the case the EVM
-// monotonic guard exists to prevent). That is acceptable for the common single-node
-// java-tron backend; if Tron is ever fronted by a load balancer, port the EVM pattern
-// here (monotonic hot path + on-advance liveness + allowRegress watchdog poll).
 func (b *TronRPC) setBestHeader(h bchain.EVMHeader) bool {
 	if h == nil || h.Number() == nil {
 		return false
--- a/bchain/coins/tron/tronrpc.go
+++ b/bchain/coins/tron/tronrpc.go
@@ -287,6 +287,16 @@ func (b *TronRPC) GetBestBlockHash() (string, error) {
 
 // GetBlockHash returns block hash in Tron API format (without 0x prefix).
 func (b *TronRPC) GetBlockHash(height uint32) (string, error) {
+	if b.isBlockSolidified(uint64(height)) && b.solidityNodeHTTP != nil {
+		ctx, cancel := context.WithTimeout(b.requestContext(), b.Timeout)
+		defer cancel()
+		hash, err := b.requestBlockHashByNum(ctx, height, true)
+		if err == nil {
+			return hash, nil
+		}
+		glog.V(1).Infof("Tron native getblock hash lookup failed for height %d, falling back to JSON-RPC: %v", height, err)
+	}
+
 	hash, err := b.EthereumRPC.GetBlockHash(height)
 	if err != nil {
 		return "", err
--- a/bchain/coins/tron/tronrpc.go
+++ b/bchain/coins/tron/tronrpc.go
@@ -697,11 +697,11 @@ func (b *TronRPC) computeBlockConfirmations(blockNumber uint64) (uint32, error) 
 	return uint32(bestHeight - blockNumber + 1), nil
 }
 
-func (b *TronRPC) buildTxFromHTTPData(txByID *tronGetTransactionByIDResponse, txInfo *tronGetTransactionInfoByIDResponse, blockTime int64, confirmations uint32, internalData *bchain.EthereumInternalData, isSolidified bool) (*bchain.Tx, error) {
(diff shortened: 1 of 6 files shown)


bchain/coins/tron/tronrpc.go
```diff
@@ -948,7 +927,9 @@ func (b *TronRPC) GetBlock(hash string, height uint32) (*bchain.Block, error) {
 			txInternalData = &internalData[i]
 		}
 
-		rebuiltTx, err := b.buildTxFromHTTPData(txByID, txInfo, bbh.Time, confirmations, txInternalData, isSolidified)
+		// Keep receipts for block indexing even before solidity. If logs are dropped
+		// here, token-transfer participants never reach the address/contract indexes.
+		rebuiltTx, err := b.buildTxFromHTTPData(txByID, txInfo, bbh.Time, confirmations, txInternalData, true)
```
pragmaxim, line 932: This fixes the **indexing** path, but the **serving** path stays asymmetric: `GetTransaction` (line 1044) and `GetTransactionForMempool` (line 1074) still pass `isSolidified`/`false`, so for a TRC-20 transfer in a mined-but-not-yet-solidified block (~19-block window) the transfer is indexed *with* its receipt here, yet a live tx-detail fetch of the same tx returns `Receipt=nil`, pending status, and zero token transfers. The address page would then list the transfer while the transaction page shows none — the same index-vs-detail disparity #1576 targets, on the serving side. The PR description says the `GetTransaction`/mempool paths are intentionally left pending until solid; flagging to confirm that user-visible inconsistency is acceptable, or whether `GetTransaction` should also keep the receipt once the tx is in a block.

bchain/coins/tron/tronrpc.go
```diff
@@ -347,21 +357,6 @@ func (b *TronRPC) GetBlockInfo(hash string) (*bchain.BlockInfo, error) {
...
-		}
-		b.bestHeaderLock.Lock()
-		defer b.bestHeaderLock.Unlock()
-		if b.bestHeader == nil || b.bestHeader.Number() == nil {
-			return nil, errors.New("best header is nil")
-		}
-		return b.bestHeader, nil
-	}
-
 	b.bestHeaderLock.Lock()
```
pragmaxim, line 360: Heads-up on the removed `if b.mq == nil { refreshBestHeaderFromChain() }` block: `getBestHeader` now serves a cached tip for up to `tronBestHeaderMaxAge` (30s), and the watchdog that refreshes it only starts in `InitializeMempool`, which runs *after* the initial `ResyncIndex`. So during initial bulk sync (and in polling-only mode before mempool init) the target tip and `computeBlockConfirmations` can lag the chain by up to 30s where they were previously always fresh. This looks like the deliberate trade-off from the "reduce requests per block" commit and is benign for far-behind sync — just confirming it's intended and that the syncer can't briefly treat the stale tip as fully-synced.

bchain/coins/tron/tronrpc.go
```diff
@@ -384,22 +379,6 @@ func (b *TronRPC) getBestHeader() (bchain.EVMHeader, error) {
...
-// is what lets a genuine rollback surface immediately to resyncIndex, so Tron is not
-// subject to the frozen-tip masking that the EVM monotonic guard introduces (and which
-// EVM has to undo with a watchdog regress).
-//
-// Tradeoff: with a load-balanced Tron RPC, a single lagging node answering a re-query
-// could regress the tip and trip a spurious fork in resyncIndex (the case the EVM
-// monotonic guard exists to prevent). That is acceptable for the common single-node
-// java-tron backend; if Tron is ever fronted by a load balancer, port the EVM pattern
-// here (monotonic hot path + on-advance liveness + allowRegress watchdog poll).
 func (b *TronRPC) setBestHeader(h bchain.EVMHeader) bool {
```
pragmaxim, line 382: This branch deleted the doc comment that explained `setBestHeader` is *intentionally* non-monotonic — it accepts a lower height so a genuine rollback surfaces immediately to `resyncIndex`, unlike `EthereumRPC.setBestHeader`'s monotonic guard, plus the load-balancer caveat. The body is unchanged, so that rationale is still accurate. Consider restoring a trimmed version: without it, a future maintainer "harmonizing" Tron with the EVM path could re-add a monotonic guard and silently reintroduce frozen-tip masking on rollbacks.

pragmaxim (commented):
Reviewed the branch. I pushed two fixes directly:

- **Drain the HTTP response body before close** (`tronhttp.go`) — `net/http` only pools a keep-alive connection when the body is read to EOF and then closed. `Request()` closed without draining (the `>=300` path reads nothing; `json.Decoder` can leave trailing bytes), so connections were torn down instead of reused — defeating the `MaxIdleConns*` tuning this branch added for sync speed.
- **Drop the dead `isSolidified` param from `requestBlockHashByNum`** — the only caller (the `GetBlockHash` fast path) is guarded by `isBlockSolidified() && solidityNodeHTTP != nil` and always passed `true`, so the `/wallet/getblock` branch was unreachable.

The three inline comments below are design calls I left for you rather than changing unilaterally.

pragmaxim (approved):

PR #1590, opened 2026-06-27
Adjust REST/UI rate-limit defaults for individual user usage

pragmaxim:
## What

Adjust the REST/UI per-client rate-limit **defaults** to fit how the public HTTP surface is actually used — by individual people, not high-throughput clients.

| `<network>_REST_UI_*` | Before | After |
|---|---|---|
| `RATE_LIMIT` (requests / min) | 180 | **20** |
| `BURST` | 40 | **20** |

`RATE_WINDOW` (1m) and `MAX_CONCURRENT` (12) are unchanged. Both adjusted values remain overridable via the existing `<network>_REST_UI_*` env vars.

## Why

The public HTTP surface (explorer UI pages + REST API) is meant for individual human use; **Trezor Suite talks to the WebSocket interface, not REST/UI**, which has its own separate limits. The previous defaults were therefore much more generous than real browsing requires.

That headroom is the problem under crawling: the explorer renders each page as a single HTML request (e.g. `/address/0x…`), and a distributed crawler spread across many source IPs can sustain a large aggregate request rate while every individual per-IP key stays well under 180/min — so the limiter never rejects anything.

Tightening the per-client budget makes the limits match individual usage and raises the cost of distributed crawling:

- **20 req/min** = 20 explorer pages per minute sustained, which comfortably covers a person reading the explorer, while forcing a crawler onto roughly **10× more source IPs** to hold a given aggregate rate.
- **Burst 20** still allows a short flurry of page loads (since each page is one request), keeping the UI snappy for real users.

This is a defaults/tuning change only; it is not a substitute for edge-level controls (e.g. Cloudflare rate-limiting / bot management on `/address/`, `/search/`, `/xpub/` and locking the origin to the CDN), which remain the right tool against a genuinely distributed crawl.

## Changes

- `server/rest_rate_limiter.go` — lower `defaultRestUIRateLimit` (180→20) and `defaultRestUIBurst` (40→20).
- `server/rest_rate_limiter_test.go` — update the default-locking assertions.
- `docs/env.md` — document the new defaults.

## Testing

`go test -tags unittest ./server/ -run 'RateLimit|ClientIP|RouteMatching|Limiter|Bypass|TrackingCap'` passes.

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -2018,7 +2018,7 @@ func (w *Worker) balanceHistoryForTxid(addrDesc bchain.AddressDescriptor, txid s
 // (0) via <NET>_{WS,REST}_BALANCE_HISTORY_MAX_TXS (<NET>_BALANCE_HISTORY_MAX_TXS
 // sets both).
 const (
-	DefaultBalanceHistoryMaxTxsREST = 250000
+	DefaultBalanceHistoryMaxTxsREST = 100000
 	DefaultBalanceHistoryMaxTxsWS   = 1000000
 )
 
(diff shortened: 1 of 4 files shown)


cranycrane (approved):
LGTM

PR #1591, opened 2026-06-27
fix(eth): prevent panic parsing non-32-byte-aligned input data

pragmaxim:
## What

Fix a panic in `EthereumParser.ParseInputData` when a transaction's input data is **not 32-byte aligned**.

## Symptom

Observed in production while rendering / crawling an Ethereum address page:

```
E0627 20:59:11 dataparser.go:254] ParseInputData recovered from panic:
  runtime error: index out of range [349] with length 349, 0000...0020...
… eth.tryParseParams … dataparser.go:224
… eth.(*EthereumParser).ParseInputData … dataparser.go:283
… api.(*Worker).GetTransactionFromBchainTx … server.(*PublicServer).explorerAddress
```

The panic is caught by `ParseInputData`'s deferred `recover()`, so it does **not** crash the process — but it returns `nil` and dumps a full `debug.PrintStack()` for **every** offending transaction, which is heavy log spam when an affected address is rendered or crawled.

## Root cause

`tryParseParams` sizes its bookkeeping slice as `processed := make([]bool, len(data)/64)` — one entry per 32-byte ABI word. Standard ABI-encoded calldata is always a multiple of 32 bytes, but a malformed / non-standard transaction is not. When marking the words occupied by a dynamic field (`bytes` / `string` / slice), the rounded-up word count could advance `dynIndex` one slot past `processed`, because the byte-level fit check (`de <= len(data)`) allows content to extend into a trailing partial word that the integer division dropped. Writing `processed[dynIndex]` then indexed out of range.

## Fix

Bound the dynamic-field marking loop: if `dynIndex` reaches `len(processed)`, the declared layout runs past the data, so reject the signature (`return nil, 0, false`) instead of writing out of range — exactly how the parser already treats other malformed input. With the data genuinely word-aligned (the valid-ABI case) the bound is never hit, so well-formed parsing is unchanged.

This eliminates both the panic and the per-transaction recover/stack-dump log spam.

## Testing

Added `TestParseInputDataMisalignedDynamicData`, a minimal reproducer with a non-32-byte-aligned `bytes` field. Verified it reproduces the original panic (same `dataparser.go:224` site) without the fix and passes with it; the existing `TestParseInputData` table still passes.

```
go test -tags unittest ./bchain/coins/eth/ -run TestParseInputData
```

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -1088,7 +1088,13 @@ func (w *Worker) getEthereumContractBalance(addrDesc bchain.AddressDescriptor, i
 				b, err = w.chain.EthereumTypeGetErc20ContractBalance(addrDesc, c.Contract)
 				if err != nil {
 					// return nil, nil, nil, errors.Annotatef(err, "EthereumTypeGetErc20ContractBalance %v %v", addrDesc, c.Contract)
-					glog.Warningf("EthereumTypeGetErc20ContractBalance addr %v, contract %v, %v", addrDesc, c.Contract, err)
+					if errors.Cause(err) == eth.ErrInvalidErc20Balance {
+						// Benign and high-volume: unparseable/empty balanceOf from a dead or non-ERC20-conforming
+						// token. Tracked via the "invalid" metric; logged at V(2) to avoid flooding logs.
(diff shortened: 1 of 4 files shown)


cranycrane (approved):
LGTM

PR #1594, opened 2026-06-29
test(eth): cover processCallTrace internal-transfer extraction

pragmaxim:
## Background

NowNodes reported that Polygon tx
`0x23d7acf0d6a0de63b3cdd6e4bc53c60d4a48f19b797d57f00c267a3e0a6f25cc` produces no
internal transactions in Blockbook, while
`0xba9e993b81714c6312bbfcebf0b929f575e71776cc91db31c13b918baa5d98fb` does, and
hypothesised that `processCallTrace` was dropping certain call-trace patterns
(DELEGATECALL etc.).

## Investigation (against the Polygon `bor` archive backend)

I pulled the real `callTracer` traces for both transactions from the configured
Polygon archive RPC and compared them with what Blockbook's API returns.

`0x23d7…` is a **WMATIC `withdraw(uint256)`** (`0x2e1a7d4d`, contract
`0x0d500b1d…` = WMATIC), block **15526560** (June 2021). Its trace is:

```
CALL  value=0x0            0x749da3b1… (user)  -> 0x0d500b1d… (WMATIC)
└─ CALL value=0x4d15bb279a6b725  0x0d500b1d… (WMATIC) -> 0x749da3b1… (user)   input=0x
```

The inner frame is a **plain value-bearing `CALL`** (~0.347 MATIC paid back to the
user) — **not** a DELEGATECALL. So the original DELEGATECALL hypothesis does not
apply, and by the flattening rules `processCallTrace` *should* emit a transfer.

I confirmed it does: feeding this verbatim trace through the current
`processCallTrace` produces exactly one internal transfer
(`CALL 0x0d500b1d… → 0x749da3b1…`, value `347159468387514149`). The new test
`TestProcessCallTrace_RealWMATICWithdraw` pins this.

So the extraction code is **correct**, yet Blockbook's API returns
`internalTransfers: null` for this tx. The reason is **stale indexed data**:

- Blockbook computes internal-transaction data when a block is fetched
  (`getInternalDataForBlock` → `debug_traceBlockByHash`) and does **not
  automatically re-trace a block that was already synced without an internal-data
  error** (the refetch routine only revisits blocks flagged with one, e.g. trace
  timeouts).
- Block 15526560 was synced before internal-transaction data was being captured
  for that range. **All four** transactions in that block that have value-bearing
  internal CALLs return `internalTransfers: null`, whereas the recent `0xba9e…`
  (block 87524447) returns all six of its internal transfers.
- The block trace today is perfectly aligned (224 txs ↔ 224 trace entries, each
  carrying `txHash`), so the Polygon bridge-padding path is not involved and there
  is no misalignment.
- Sampling value-bearing-internal-CALL transactions across the whole chain shows a
  **sharp, monotonic sync-era cutoff** on this backend: `internalTransfers` is
  `null` at/below block ~**49.5M** and populated at/above ~**49.6M**. `0x23d7…`
  (block 15.5M) is far below it; several other `null` txs below the cutoff are
  themselves WMATIC withdraws structurally identical to `0x23d7…`. Every recent
  value-bearing internal CALL sampled (89.1M–89.3M) is present — i.e. no live
  extraction bug.

**Conclusion: not a `processCallTrace` bug.** The missing internal transactions
are pre-capture history. Blockbook records internal data only when a block is
fetched and has **no targeted historical backfill** — the automatic refetch
routine revisits only blocks flagged with an internal-data error (these were not),
rollback only works downward from the tip, and the bulk-connect path is the
initial-sync writer (not safe to re-run in place). So such blocks are repopulated
only by a full resync of the index. Note the impact is scoped: eth-type balances
are served live from the node (`eth_getBalance`), so they remain correct — only
the `internalTransfers` list and the address↔tx history for addresses involved
*solely* via an internal transfer are affected.

## What this PR changes

**Tests only — no production code is touched.** The extraction path was previously
untested, which is what made correct behaviour look like a bug. The emit rules,
the `DELEGATECALL` double-counting rationale, and the stale-data/no-backfill notes
live in the test file header (rather than in `ethrpc.go`), with the tests as their
executable proof:

- Adds `ethrpc_calltrace_test.go`:
  - `TestProcessCallTrace_RealWMATICWithdraw` — drives the **verbatim** `0x23d7…`
    trace and asserts the WMATIC-withdraw transfer is extracted.
  - `TestProcessCallTrace_TransferRules` — every branch: value `CALL`, `DELEGATECALL`
    ignored vs. a real `CALL` nested under it, a skipped `DELEGATECALL` sibling
    between two recorded `CALL`s (index stability), zero-value `CALL` with/without
    `ProcessZeroInternalTransactions`, `CALLCODE`/`STATICCALL` fall-through,
    `CREATE`/`CREATE2`, `SELFDESTRUCT`, deep nesting, error propagation,
    empty/undecodable `value` (incl. the `err == nil` guard).
  - `TestProcessCallTrace_ContractLifecycleBlockHeight` — create/destruct record shape.

## Answers to NowNodes

1. **Conditions to create an internal transaction** — `CREATE`/`CREATE2` and
   `SELFDESTRUCT` always; any other frame (`CALL`, plus `CALLCODE`/`STATICCALL` by
   fall-through) when its `value` decodes and is `> 0` (or
   `ProcessZeroInternalTransactions` is set); `DELEGATECALL` never (its `value` is
   inherited `msg.value`, no value moves), though its children are still traversed.
   The top-level frame's value is the regular tx value, not an internal transfer.
2. **Known limitations** — none that explain this report. (Minor, harmless edges:
   a reverted value `CALL` is still recorded — over-, not under-report; a value
   `CALLCODE` would be recorded though it moves no balance — extinct in practice;
   Polygon state-sync/bridge deposits move value outside the call trace entirely.)
3. **Is `0x23d7…` expected?** — The empty result is **not** a code bug. Its
   value-bearing internal CALL is extracted correctly by current code; the `null`
   is stale data from when that old block was indexed (below the backend's
   internal-tx capture cutoff). There is no targeted backfill; the index-complete
   options are a full resync, or fetching internal transfers on demand per tx via
   `debug_traceTransaction` (callTracer) against a tracing archive node — which is
   exactly what Blockbook does internally.
4. **Expose the full call-trace hierarchy?** — Out of scope here; a larger feature
   with DB-schema/reindex implications. This PR documents and tests the current
   simplified representation.

## Testing

`go test -tags unittest ./bchain/coins/eth/` passes. Traces and API responses were
obtained read-only from the project's configured Polygon `bor` archive backend.

### DIFF
--- a/bchain/coins/eth/ethrpc_calltrace_test.go
+++ b/bchain/coins/eth/ethrpc_calltrace_test.go
@@ -1,0 +1,396 @@
+package eth
+
+import (
+	"encoding/json"
+	"math/big"
+	"testing"
+
+	"github.com/trezor/blockbook/bchain"
+)
+
+// These tests document and pin the rules processCallTrace uses to flatten a callTracer frame
+// tree into the simplified EthereumInternalTransfer list. Only frames that represent an actual
+// movement of native value (or a contract lifecycle event) become a transfer record:
+//
+//   - CREATE / CREATE2  -> always emitted (Type CREATE), and the new contract is recorded.
+//   - SELFDESTRUCT      -> always emitted (Type SELFDESTRUCT), and the destroyed contract is recorded.
+//   - DELEGATECALL      -> never emitted. Since geth v1.11 the bor/geth callTracer reports the
+//     inherited msg.value on DELEGATECALL frames (https://github.com/ethereum/go-ethereum/issues/26726),
+//     but a DELEGATECALL runs another contract's code in the caller's own context and moves no
+//     value, so emitting it would double-count the value already accounted for by the parent CALL.
+//     Its children are still traversed, so a genuine value-bearing CALL nested under a DELEGATECALL
+//     is still captured.
+//   - any other frame (CALL, and by fall-through CALLCODE / STATICCALL or any unrecognised type)
+//     -> emitted as Type CALL only when its value field decodes cleanly AND (value > 0 ||
+//     ProcessZeroInternalTransactions). STATICCALL cannot carry value and CALLCODE is effectively
+//     extinct, so neither is special-cased; a value-bearing CALLCODE would, like DELEGATECALL, be
+//     over-counted if one ever appeared.
+//
+// A transaction whose only sub-frames are DELEGATECALLs (inherited value) and/or zero-value CALLs
+// legitimately produces no transfers: its single native value movement is the top-level call
+// value, exposed as the regular transaction value rather than as an internal transfer.
+//
+// Separately, processCallTrace runs only when a block's internal data is fetched
+// (getInternalDataForBlock, during sync or the internal-data refetch routine); Blockbook does not
+// automatically re-trace an already-synced block that completed without an internal-data error.
+// So a transaction that genuinely contains a value-bearing internal CALL but shows no internal
+// transfers in the API was most likely indexed before internal-transaction data was captured for
+// its block range — only a full resync repopulates it (there is no targeted historical backfill).
+// TestProcessCallTrace_RealWMATICWithdraw pins that the extraction itself is correct, so such data
+// is missing only from the stored index, not lost from the chain.
+
+// newCallTraceRPC builds an EthereumRPC suitable for exercising processCallTrace directly.
+func newCallTraceRPC(processZero bool) *EthereumRPC {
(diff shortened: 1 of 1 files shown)


pragmaxim:
This was just a proof there is no bug.

PR #1596, opened 2026-07-01
fix(grafana): correct stale instant-tier claim in EIP-1559 max-fee panel

pragmaxim:
- Reviewed the recent Grafana/fee-related commits for the suspected mixing of "Fiat Rates" and "Fees" concepts — confirmed they are cleanly separated (distinct panel keys, metric prefixes, and Go structs), no changes needed there.
- Found and fixed a real, unrelated staleness bug: `a2ec3b3d` remapped 1inch's fee tiers (medium→low, high→medium, instant→high), so 1inch no longer exposes a separate instant tier. The `fees.eip1559_max_fee` panel description still claimed 1inch was a "provider that returns four tiers", which is no longer true and would send an on-call engineer chasing a phantom instant-tier discrepancy.

### DIFF
--- a/configs/grafana/panels.yaml
+++ b/configs/grafana/panels.yaml
@@ -665,7 +665,7 @@ panels:
         legend: '{{instance}} - {{provider}} - {{status}}'
   fees.eip1559_max_fee:
     title: EIP-1559 maxFeePerGas by tier (Gwei)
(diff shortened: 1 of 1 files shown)


cranycrane (approved):
LGTM

PR #1605, opened 2026-07-04
fix(eth): avoid int64 overflow in on-chain base fee conversion

pragmaxim:
**Severity: minor** — theoretical today, but a cheap and correct fix worth taking.

The pull-path EIP-1559 estimate (`EthereumTypeGetEip1559Fees`) decoded `baseFeePerGas` as `uint64` and then cast it to `int64` for `big.NewInt`. A value above `math.MaxInt64` (~9.22e18 wei) would wrap to a negative `big.Int`.

Not reachable on Ethereum mainnet today, but possible on high-fee L2s. Switching to `new(big.Int).SetUint64()` also makes this consistent with the header path in `attachBlockGas`, which already handles it correctly.

Follow-up from review of the EIP-1559 fields work (already merged to master).


cranycrane (approved):
LGTM

PR #1610, opened 2026-07-08
fix(eth): recover GetTransaction on archive backends with pruned tx index

### LINKED ISSUE #1609: EVM GetTransaction fails on archive backends that prune the eth_getTransactionByHash index (Base)
## Summary

`EthereumRPC.GetTransaction` fails with `ErrTxNotFound` ("Tx not found") for mined transactions that are older than the backend's `eth_getTransactionByHash` retention window, even when the transaction is fully retained and reachable via `eth_getBlockByHash` and `eth_getTransactionReceipt`.

This surfaced in CI after the Base backend was pointed at the QuikNode archive endpoint, failing `TestIntegration/base=main/rpc/GetTransaction` (and it would equally fail the API-suite `GetTransaction` / `WsGetTransaction` for Base). It is **not** caused by the PR that first tripped it (that PR only touches server/db admin + rpcCall-allowlist code).

## Root cause

`GetTransaction` relies solely on `eth_getTransactionByHash` (`bchain/coins/eth/ethrpc.go`). The QuikNode Base archive node prunes the **transaction-by-hash index** to roughly the last ~1–1.5M blocks (~1 month) while remaining a full archive for block bodies and receipts. For an older transaction the call returns `null`, which blockbook maps to `ErrTxNotFound`.

## Evidence (live node, fixture tx `0xd20ea523…`, block 45629012)

| RPC method | result |
|---|---|
| `eth_getTransactionByHash` | ❌ `null` (consistent) |
| `eth_getBlockByHash` (full txs) | ✅ block, 252 txs, tx present |
| `eth_getTransactionReceipt` | ✅ receipt with `blockHash`/`blockNumber`/`transactionIndex` |

Recent and ~1M-block-old txs on the same node **do** resolve via `eth_getTransactionByHash`; only txs beyond the pruned window fail — confirming an index-retention limit, not a reorg or data loss.

## Impact — this is a runtime dependency, not just a test concern

**EVM transaction bodies are never persisted during sync**, so blockbook must fetch them from the backend on demand. This makes the bug a real production data-availability gap on Base, not merely a CI failure.

Evidence from the call graph:

- The `cfTransactions` column family is written in exactly one place — `RocksDB.PutTx` (`db/rocksdb.go:1912`) — which is called from exactly one place: the TxCache **on a cache miss** (`db/txcache.go:92`).
- The EthereumType sync path (`db/rocksdb_ethereumtype.go`) only ever *deletes* from `cfTransactions` (reorg/disconnect, line 1358). Sync indexes address→txid mappings and internal data, but **not** full transaction bodies.
- Therefore for EVM, tx bodies live only in a lazily-populated cache:
  ```
  db/txcache.go:38  TxCache.GetTransaction
    ├─ hit  → served from RocksDB (cfTransactions)
    └─ miss → chain.GetTransaction(txid)  →  eth_getTransactionByHash   ← fails here
              then PutTx (populate cache)
  ```

Runtime paths that reach `chain.GetTransaction` for EVM (all via the `ChainEthereumType` branch at `api/worker.go:1991`):

- `GET /api/v2/tx/{txid}`, `/api/v1/tx/{txid}`, HTML `/tx/{txid}` — `server/public.go:955,1275,1523`.
- WebSocket `getTransaction` — `server/websocket.go:957`.
- Address / xpub transaction listings — each tx detail is fetched the same way.
- EVM mempool sync — `bchain/mempool_ethereum_type.go:49` → `GetTransactionForMempool` → `GetTransaction`.

Consequence: **any Base transaction older than the node's tx-by-hash retention window that is not already cached returns "not found" from the live API/WS** — e.g. opening an old transaction, or paging into the older history of an address. blockbook's own address index still references those txids, so the gap is visible, not silent. Bitcoin-type chains are unaffected: they persist `TxAddresses` at sync time and take a different branch (`api/worker.go:1980`).

- **CI**: Base integration tests fail (the first place it surfaced).

## Fix

`GetTransaction` now falls back to reconstructing the transaction from its receipt (for the block hash) + the full block body when `eth_getTransactionByHash` returns null. Genuinely-unknown txs still return `ErrTxNotFound`; the primary path (and its behavior for all other chains) is unchanged.

Longer term, consider pointing Base at a backend that retains the full transaction-by-hash index, or confirm the fallback is acceptable as the permanent behavior for index-pruned archive backends.

pragmaxim:
## TL;DR

Without this fix, transactions from the pruned period are simply **broken** — they return `ErrTxNotFound` from the live API/WS even though the node fully retains them. The recovery path added here only ever runs in exactly that situation: when the primary lookup returns `null` for a transaction the node actually retains. It trades a wrong "not found" answer for a correct one, and cannot regress any path where the primary lookup already works.

## Problem

`EthereumRPC.GetTransaction` relies solely on `eth_getTransactionByHash`, which returns `null` for mined transactions older than the backend's transaction-by-hash retention window on some archive nodes (observed on the QuikNode Base backend, which prunes that index to ~1 month while remaining a full archive for block bodies and receipts). blockbook maps the `null` to `ErrTxNotFound`.

This is a **runtime dependency**, not just a test issue: EVM transaction bodies are never persisted during sync (`cfTransactions` is written only lazily by the TxCache on a miss; the eth sync path only deletes from it). So the first API/WS access of any eth transaction — and any tx never accessed or evicted — fetches from the backend via `chain.GetTransaction`. Consequently, on Base, transactions older than the node's window that aren't already cached return "not found" from the live API/WS (e.g. opening an old transaction, or paging into the older history of an address).

Full analysis and evidence in #1609.

## Fix

Fall back to reconstructing the transaction from its receipt (which still carries the block hash) plus the full block body when `eth_getTransactionByHash` returns `null`. Genuinely-unknown transactions still return `ErrTxNotFound`, and the primary path is unchanged for backends that retain the full index.

## Verification

- Reproduced the exact failure locally against the live `base_archive` node: `TestIntegration/base=main/rpc/GetTransaction` → "Tx not found".
- After the fix, that test passes, along with the full `base` rpc suite and `optimism` (whose `GetTransaction` still takes the unchanged primary path — no regression).
- `go build ./bchain/coins/eth/` and the eth unit tests pass.

Fixes #1609

### DIFF
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1727,6 +1727,49 @@ func (b *EthereumRPC) removeTransactionFromMempool(txid string) {
 	}
 }
 
+// recoverMinedTransaction reconstructs a mined transaction that eth_getTransactionByHash
+// failed to return. Some archive backends prune the transaction-by-hash index beyond a
+// recent window while still serving full block bodies and receipts (observed on QuikNode
+// Base), so an older-but-retained transaction reads as missing. The receipt still carries
+// the block hash, from which the full block body yields the transaction. Returns nil when
+// the backend genuinely has no record of the transaction (so the caller falls back to
+// ErrTxNotFound) or when the recovery lookups fail; failures are logged, not propagated,
+// to preserve the not-found semantics of the primary lookup.
+func (b *EthereumRPC) recoverMinedTransaction(txid string) *bchain.RpcTransaction {
+	ctx, cancel := context.WithTimeout(context.Background(), b.Timeout)
+	defer cancel()
+	// eth_getTransactionReceipt keeps working on such backends; the block hash is enough
+	// to fetch the block body, which contains the full transaction.
+	var receipt struct {
+		BlockHash string `json:"blockHash"`
+	}
+	if err := b.RPC.CallContext(ctx, &receipt, "eth_getTransactionReceipt", ethcommon.HexToHash(txid)); err != nil {
+		glog.Warningf("recoverMinedTransaction %s: eth_getTransactionReceipt failed: %v", txid, err)
+		return nil
+	}
+	if receipt.BlockHash == "" {
+		// No receipt: the transaction is genuinely unknown to the backend.
+		return nil
+	}
+	raw, err := b.getBlockRaw(receipt.BlockHash, 0, true)
+	if err != nil {
+		glog.Warningf("recoverMinedTransaction %s: getBlockRaw %s failed: %v", txid, receipt.BlockHash, err)
+		return nil
+	}
+	var body rpcBlockTransactions
+	if err := json.Unmarshal(raw, &body); err != nil {
+		glog.Warningf("recoverMinedTransaction %s: decode block %s failed: %v", txid, receipt.BlockHash, err)
+		return nil
+	}
+	for i := range body.Transactions {
+		if strings.EqualFold(body.Transactions[i].Hash, txid) {
+			return &body.Transactions[i]
+		}
+	}
+	glog.Warningf("recoverMinedTransaction %s: not present in block %s body", txid, receipt.BlockHash)
+	return nil
+}
+
 // GetTransaction returns a transaction by the transaction ID.
 func (b *EthereumRPC) GetTransaction(txid string) (*bchain.Tx, error) {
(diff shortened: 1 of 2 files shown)


cranycrane (approved):
LGTM

PR #1613, opened 2026-07-09
fix(eth): prevent slice-bounds panic from ENS log name-length overflow

pragmaxim:
## Problem

`getEnsRecord` (`bchain/coins/eth/dataparser.go`) reads a 32-byte name-length word `c` from an attacker-controlled ENS `NameRegistered` log and computes a slice end `de = 194+64+(int(c)<<1)`. `int(c)<<1` overflows: Go's signed left shift wraps silently, so for `c ∈ [0x7FFFFFFFFFFFFF7F, 0x7FFFFFFFFFFFFFFF]` (129 values) `de` lands in `[0, 256]`. The guard `de > len(l.Data) || de < 0` does not catch `de` in that range, so `l.Data[194+64 : de]` is a `low > high` slice expression and panics (`slice bounds out of range [258:0]`).

The call runs in the `GetBlock` log-processing goroutine (`ethrpc.go:1628`), which has no `recover()` (nor does the `db` sync path). An unrecovered panic in a goroutine terminates the whole process, and because it fires before the block is committed, the same block is re-fetched on restart → deterministic crash-loop.

Trigger: any contract emitting a `LOG3` with the ENS `NameRegistered` topic, 3 topics, `len(Data) ≥ 322`, and the length word set to `0x…7fffffffffffff7f`. Unauthenticated, low-cost availability DoS of the EVM backend.

## Changes

- **`getEnsRecord`**: guard the slice start — `de < 194+64` instead of `de < 0`. After the guard `258 ≤ de ≤ len(l.Data)`, so the slice can never panic (32- and 64-bit).
- **`processParam`**: same-class fix `de < d` for the sibling `data[d:de]` slice. Already under `ParseInputData`'s `recover()`, so no live crash — correctness only. `count==0` (`de==d`) still accepted.
- **`GetBlock`**: `recover()` in the log goroutine as defense-in-depth, returning an error instead of crashing (adds `runtime/debug` import).
- **Tests**: regression cases for `de=0`, `de=256`, and the `processParam` variant. Verified to panic with the fix reverted; return `nil` with it.

## Verification

`go vet`, `gofmt`, `go build`, and `go test -tags unittest ./bchain/coins/eth/` all pass.

## Notes

- The `recover()` returns a non-retryable error; the parallel range-sync worker (`db/sync.go:677`) would retry the same block rather than skip it — process stays up, but a deterministic panic would stall that worker. Moot once the guard removes the panic; flagged as a follow-up.
- The sibling `getInternalDataForBlock` goroutine (`ethrpc.go:1632`) has the same structural exposure and no `recover()`.

### DIFF
--- a/bchain/coins/eth/dataparser.go
+++ b/bchain/coins/eth/dataparser.go
@@ -156,7 +156,7 @@ func processParam(data string, index int, dataOffset int, t *abi.Type, processed
 		offset <<= 1
 		d = int(offset) + dataOffset
 		dynIndex := d >> 6
-		if d+64 > len(data) || d < 0 {
+		if d+64 > len(data) || d < 0 || d+64 < d {
 			return nil, 0, false
 		}
 		// get element count of dynamic type
@@ -314,17 +314,18 @@ func getEnsRecord(l *rpcLogWithTxHash) *bchain.AddressAliasRecord {
 		if err != nil {
 			return nil
 		}
+		const nameStart = 194 + 64
 		// int(c)<<1 can overflow: c is attacker-controlled (up to 2^63-1) and Go's
 		// signed left shift wraps silently, so de can land anywhere in [0, 257]
-		// below the name start offset (194+64). The lower bound must be the slice
-		// start index, not 0 — a de below the start makes l.Data[194+64:de] a
-		// low>high slice expression that panics. Checking de < 194+64 guarantees
-		// 194+64 <= de <= len(l.Data), so the slice below can never panic.
-		de := 194 + 64 + (int(c) << 1)
-		if de > len(l.Data) || de < 194+64 {
+		// below nameStart. The lower bound must be the slice
+		// start index, not 0 — a de below the start makes l.Data[nameStart:de] a
+		// low>high slice expression that panics. Checking de < nameStart guarantees
+		// nameStart <= de <= len(l.Data), so the slice below can never panic.
+		de := nameStart + (int(c) << 1)
+		if de > len(l.Data) || de < nameStart {
 			return nil
 		}
-		b, err := hex.DecodeString(l.Data[194+64 : de])
+		b, err := hex.DecodeString(l.Data[nameStart:de])
 		if err != nil {
 			return nil
 		}
--- a/bchain/coins/eth/dataparser.go
+++ b/bchain/coins/eth/dataparser.go
@@ -169,8 +169,12 @@ func processParam(data string, index int, dataOffset int, t *abi.Type, processed
 		dynIndex++
 		if t.T == abi.StringTy || t.T == abi.BytesTy {
 			d += 64
+			// count<<1 can overflow and wrap below d (same class as the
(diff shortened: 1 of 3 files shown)


bchain/coins/eth/dataparser.go
```diff
@@ -169,8 +169,12 @@ func processParam(data string, index int, dataOffset int, t *abi.Type, processed
 		dynIndex++
 		if t.T == abi.StringTy || t.T == abi.BytesTy {
 			d += 64
+			// count<<1 can overflow and wrap below d (same class as the
+			// getEnsRecord bug); the lower bound must be the slice start d, not 0,
+			// so that d <= de <= len(data) and data[d:de] below can never panic.
+			// count==0 gives de==d, which is still accepted.
 			de := d + (count << 1)
-			if de > len(data) || de < 0 {
+			if de > len(data) || de < d {
```
cranycrane, line 177: Non-blocking, same-class follow-up: one guard earlier in this function (`offset <<= 1` at line 156 and the `d+64 > len(data) || d < 0` check at line 159), the addition can itself overflow. A crafted offset word near 2^62 puts `d` in `[2^63−64, 2^63−1]`, so `d+64` wraps negative — both guards pass and `data[d:d+64]` panics. Like this fix, it's under `ParseInputData`'s `recover()`, so it's correctness-only rather than a crash, but since this PR is already fixing this exact class in this function it may be worth folding in — e.g. `if d+64 > len(data) || d < 0 || d+64 < d` (or validate `offset` before shifting).

bchain/coins/eth/dataparser.go
```diff
@@ -310,8 +314,14 @@ func getEnsRecord(l *rpcLogWithTxHash) *bchain.AddressAliasRecord {
...
 		}
+		// int(c)<<1 can overflow: c is attacker-controlled (up to 2^63-1) and Go's
+		// signed left shift wraps silently, so de can land anywhere in [0, 257]
+		// below the name start offset (194+64). The lower bound must be the slice
+		// start index, not 0 — a de below the start makes l.Data[194+64:de] a
+		// low>high slice expression that panics. Checking de < 194+64 guarantees
+		// 194+64 <= de <= len(l.Data), so the slice below can never panic.
 		de := 194 + 64 + (int(c) << 1)
-		if de > len(l.Data) || de < 0 {
+		if de > len(l.Data) || de < 194+64 {
```
cranycrane: Nit: `194+64` now appears three times in this function (guard, comment, slice expression). A local `const nameStart = 194 + 64` would make the guard's intent (`de < nameStart`) self-evident and keep the three sites from drifting.

cranycrane (commented):
Reviewed — the fix is minimal and provably correct on both 32- and 64-bit, the regression vectors check out, and the `recover()` is appropriately defense-in-depth rather than the primary mitigation. Two minor, non-blocking observations inline.

cranycrane (approved):
LGTM

PR #1624, opened 2026-07-13
fix(eth): prevent duplicate ERC-721 holdings

### LINKED ISSUE #1619: eth: ERC721 NFT holdings can accumulate duplicate token IDs
### Summary

The per-address NFT holdings index for ERC721 tokens can store the same token ID multiple times. Repeated `Transfer` logs for the same `(address, contract, tokenId)` add duplicate entries to the address's holdings, producing an incorrect ownership list.

### Details

`Ids.insert` and `unpackedIds.insert` insert a token ID at its sorted position without checking whether an equal ID already exists:

- `search()` returns the first index whose value is `>= id`; `insert()` then splices the new value in unconditionally, so an equal ID is duplicated rather than treated as a no-op.
- `remove()` deletes only a single matching entry, so N duplicate inserts followed by one transfer-out leave N-1 stale copies behind.
- There is no dedup on the pack/read path, and `SortAddressContracts` (the sort migration) sorts but does not deduplicate, so existing duplicates persist permanently.

Because `Transfer` events are emitted by arbitrary contracts, the indexer cannot assume a given `(contract, tokenId)` appears at most once; even for well-behaved contracts a re-emitted `Transfer` yields a wrong holdings list.

ERC1155 is not affected — `MultiTokenValues.upsert` aggregates by ID, so it cannot accumulate duplicate entries.

### Location

- `db/rocksdb_ethereumtype.go` — `Ids.insert` (~L47), `unpackedIds.insert` (~L1597)
- Called from `addToContract` (~L435)

### Impact

Account-details / token APIs report the same ERC721 token ID more than once for an address, i.e. incorrect NFT holdings/ownership.

### Suggested fix

In `insert`, treat an existing equal ID as a no-op (reject the duplicate). Consider deduplicating during the sort migration so already-indexed rows are repaired.

cranycrane:
Fixes #1619.

## Summary

- Make ERC-721 holding insertion idempotent so one address cannot store the same token ID more than once.
- Preserve ERC-721 holdings across self-transfers and their rollback while retaining transaction indexing.
- Add a read-only RocksDB checker that reports duplicate ERC-721 holdings and tells the administrator to reindex; it never repairs or normalizes the database.

## Bug scenario

If `tokenId 42` was transferred from A to B and the same `Transfer` log was emitted or replayed again, Blockbook previously credited B twice and stored `[42, 42]`. A later transfer from B to C removed only one copy, leaving a stale `42` under B while also adding it to C. Since transfer logs come from arbitrary contracts, the indexer cannot assume they are unique.

## Validation

- Full Go unit test suite
- Blockbook connectivity integration suite
- Targeted database and checker tests
- Ethereum OpenAPI end-to-end suite


db/rocksdb_ethereumtype.go
```diff
@@ -669,11 +680,12 @@ func (d *RocksDB) processContractTransfers(blockTx *ethBlockTx, tx *bchain.Tx, a
 			glog.Warningf("rocksdb: processContractTransfers %v, tx %v, transfer %v", err, tx.Txid, t)
 			continue
 		}
-		if err = d.addToAddressesAndContractsEthereumType(to, blockTx.btxID, int32(i), contract, t, true, addresses, addressContracts); err != nil {
+		eq := bytes.Equal(from, to)
+		mutateHoldings := !eq || t.Standard != bchain.NonFungibleToken
```
pragmaxim: The `mutateHoldings` invariant is derived in two places with different phrasings — here (`!eq || t.Standard != NonFungibleToken`) and in `disconnectAddress` (`selfTransfer && transferStandard != NonFungibleToken`). They're consistent, but a reader has to verify that by hand. Consider a small shared helper (e.g. `isNFTSelfTransfer(from, to, standard)`) or at least a comment cross-referencing the disconnect site, so the two can't drift.

db/rocksdb_ethereumtype.go
```diff
@@ -417,6 +420,14 @@ func addToContract(c *unpackedAddrContract, contractIndex int, index int32, cont
 			s.Add(s, v)
 		}
 	}
+	if !mutateHoldings {
```
pragmaxim, line 431: This early return is generic, but the only callers passing `mutateHoldings=false` are NFT self-transfers. If a future caller passed `false` for an ERC-20/1155, it would silently skip value normalization / upsert. Worth a one-line comment noting the flag is specific to the NFT self-transfer case so the assumption is explicit.

db/rocksdb_ethereumtype_test.go
```diff
@@ -43,6 +52,103 @@ func makeTestAddrDesc(seed int) bchain.AddressDescriptor {
...
+		t.Fatalf("Ids after repeated transfer-in = %v, want [42]", c.Ids)
+	}
+
+	addToContract(c, 0, transferFrom, contract, transfer, false, true)
+	if len(c.Ids) != 0 {
+		t.Fatalf("Ids after transfer-out = %v, want empty", c.Ids)
+	}
+}
+
+func TestERC721SelfTransferDoesNotMutateHoldings(t *testing.T) {
```
pragmaxim, line 116: This exercises the connect side via `processContractTransfers`, but there's no full `ConnectBlock -> DisconnectBlock` round-trip test for an NFT self-transfer. The connect/disconnect symmetry (holdings untouched on both sides, tx count balanced) is exactly where a regression would hide. A round-trip test would harden this path — the general suite passes, so this is hardening rather than a defect.

pragmaxim (approved):

PR #1626, opened 2026-07-13
feat(websocket): make per-connection pending request limit env-configurable

knaperek:
## Problem

`maxWebsocketPendingRequests` (48) is a hard-coded const. It caps how many requests a single websocket connection may have executing concurrently; on breach the connection is closed (`pending_requests_limit`).

Interactive clients such as Trezor Suite never come near it, but self-hosted deployments with internal batch consumers do: our exchange backend scans several thousand deposit addresses by pipelining `getAccountInfo` requests over one connection, and every scan is disconnected mid-batch. The adjacent websocket limits (`<network>_WS_GETACCOUNTINFO_LIMIT`, `<network>_WS_MESSAGE_RATE_LIMIT`, `<network>_WS_IP_BLOCK_DURATION`) are already env-configurable so operators can tune them for trusted internal traffic - this one is not.

## Change

- Read the cap from `<network>_WS_PENDING_REQUESTS_LIMIT` via the existing `parseNonNegativeIntEnv` helper, in `configurePendingRequestsLimit` mirroring `configureMessageRateLimit`.
- Default remains **48**; behavior with the variable unset is unchanged. `0` disables the cap (nil slot channel; the server-wide work limit remains as a backstop against unbounded concurrency).
- The effective value is logged at startup, like the other WS limiters.
- Documented in `docs/env.md`.

## Tests

`Test_WebsocketClosesWhenPendingRequestLimitExceeded` is refactored into a helper parameterized by the limit (still covering the default), and the new `Test_WebsocketPendingRequestLimitEnvOverride` sets `FAKE_WS_PENDING_REQUESTS_LIMIT=3` and verifies the connection closes at the 4th concurrent request. Full suite passes.

The patch has been running in production on our ETH archive instance since 2026-07-13.


pragmaxim:
Hi @knaperek, I forgot to add it, thank you :+1:

pragmaxim (approved):

PR #1628, opened 2026-07-14
fix(eth): gate alternative-provider nonce lookups to recent private senders

### LINKED ISSUE #1627: eth: every address request sends eth_getTransactionCount to the alternative provider, causing constant 429 rate-limit errors
## Problem

On Ethereum Blockbook instances configured with an alternative send-tx provider (`*_ALTERNATIVE_SENDTX_URLS`), the error log fills with:

```
E0714 08:43:51.980500 ... ethrpc.go] Alternative provider failed for eth_getTransactionCount: 429 Too Many Requests: {"id":null,"error":{"code":-32005,"message":"Rate limit exceeded"},"jsonrpc":"2.0"}, falling back to primary RPC
```

## Cause

`EthereumTypeGetNonces` (`bchain/coins/eth/ethrpc.go`) routes the `eth_getTransactionCount` lookup to the alternative provider **unconditionally** whenever one is configured — for every address, on every address detail request. Since this runs on the hottest API endpoint (`/api/v2/address/...`), the provider's per-key RPS quota is exhausted immediately.

The intent of the routing is correct: transactions submitted privately through the alternative provider never reach the public mempool, so only that provider knows the true `pending` nonce for those senders. But almost all addresses being queried have no privately-submitted tx in flight, and for them the alternative-provider call is pure waste.

Compounding factors in `AlternativeSendTxProvider.getNonces` (`bchain/coins/eth/alternativesendtx.go`):
- no caching, rate limiting, or 429 backoff
- `callHttpRawResult` dials a brand-new RPC client for every call (no connection reuse)
- with `WithConfirmedNonce`, each request is a 2-element batch (`pending` + `latest`), which many providers count as 2 requests against quota

## Consequences

- **Stale pending nonce in the one case the feature exists for**: when the 429 forces fallback to the primary RPC, a sender with an in-flight privately-submitted tx gets a stale `pending` nonce — a wallet building the next tx from it would reuse the nonce and replace the private tx.
- **Quota exhaustion affecting broadcasts**: the nonce reads burn the same API-key quota as `eth_sendRawTransaction`, so an actual private broadcast can itself get rate-limited under address-request load.
- **Latency**: every address request pays a failed provider round-trip (dial + call + 429) before the primary call.
- **Log noise**: `glog.Errorf` at request rate floods the ERROR log and drowns out real provider outages.

## Proposed fix

1. **Gate the alternative-provider nonce path to senders that need it.** The provider already maintains the alternative mempool cache (`mempoolTxs`, when `*_ALTERNATIVE_FETCH_MEMPOOL_TX=TRUE`). Route `getNonces` through the provider only when that cache holds a pending tx from the queried address; otherwise go straight to the primary RPC. Caveat: when running with `*_ALTERNATIVE_SENDTX_ONLY=TRUE` but without the mempool cache, keep always asking the provider.
2. **Add a client-side rate limiter and a short (few seconds) per-address nonce TTL cache** as protection against bursts, and reuse a persistent RPC client instead of dialing per call.
3. **Demote the log**: a 429 with a working fallback is expected operation — log at warning/verbose level or throttle (first occurrence + counter) so genuine provider outages stay visible.

cranycrane:
Fixes #1627

## Why

On instances with `*_ALTERNATIVE_SENDTX_URLS` configured, `EthereumTypeGetNonces` routed `eth_getTransactionCount` to the alternative provider on a bare nil-check — for **every** address detail request, the hottest API endpoint. This exhausted the provider's per-key rate quota, flooding the error log with 429s, and the forced fallback to the primary RPC handed a stale pending nonce to exactly the private-tx senders the feature exists for. Almost none of the queried addresses have a privately submitted tx in flight, so nearly all of that provider traffic was waste.

## What & why, per commit

- **Gate nonce lookups to recent private senders** — only a sender with a private tx in flight needs the provider's nonce view. The sender is decoded from the raw tx hex at send time (rather than gating on the alternative mempool cache), because that works in every config mode and survives a failed post-send fetch-back — the cases where cache-based gating would silently regress to stale nonces. The fallback log drops to a warning since, once gated, it only fires for a genuine provider failure. This removes ~100% of the steady-state provider traffic without new config knobs.
- **Route nonce reads to the provider that accepted the send** — a broadcast succeeds if *any* configured URL accepts it, but nonce reads always hit `urls[0]`, which may never have seen the tx. Recording the accepting URL keeps the read on the one provider guaranteed to know the transaction.
- **Release nonce routing when the sender's private txs settle** — otherwise address polling keeps burning provider quota for the full cache timeout after the tx is already mined/superseded, and the gate's benefit erodes.
- **Raise the pending-nonce answer to the alternative mempool floor** — applied to both the primary-RPC fallback *and* the provider's own answer. The primary cannot see private pending txs at all, and Blink-style relays stop counting a still-pending tx at the `pending` tag while Blockbook deliberately keeps exposing it (see `reconcileMempoolTxs`). The floor (`max(cached nonce)+1`) guarantees the reported pending nonce never contradicts Blockbook's own pending view, so a wallet cannot accidentally replace the sender's in-flight private transaction.

## Deployment note

The routing state is in process memory. With multiple Blockbook instances behind a load balancer, only the instance that accepted the transaction routes the sender's nonce queries to the alternative provider — websocket clients are naturally sticky to one instance, but plain REST polling may hit another instance and receive the public backend's view (pre-PR behavior for that instance). A restart clears the state as well; exposure is bounded by `alternativeMempoolTxTimeout`. The gating therefore assumes request affinity for send+poll flows, which websocket clients get inherently. An env-var opt-out was added during review and deliberately removed again as speculative config surface for a deployment shape production does not run; it can be re-added compatibly if a non-affine deployment ever needs it. Documented in `docs/env.md`.

Deliberately *not* included (kept minimal): client-side rate limiter / nonce TTL cache / persistent RPC clients — the gating removes the traffic those would protect against; they can be added later if ever needed. `EthereumTypeEstimateGas` has the same unconditional-provider pathology and is tracked in #1629.

## Testing

- `go test -tags unittest -run 'Nonce|AlternativeSendTx' ./bchain/coins/eth/` — full suite passes, including regression tests: zero provider round-trips for a non-sender address, gated sender served by the provider, floor applied over both the fallback and a successful provider answer, sender released on eviction (4 scenarios), and accepting-URL routing on both the pending-only and batched paths.
- Each commit compiles and passes vet/gofmt independently.


cranycrane:
Addressed point by point:

1. **Floor on successful provider answers** — applied (fa0a5a0d): `raiseToPendingFloor` now wraps both the provider answer and the primary fallback, making the "never answer below our own pending view" invariant unconditional. Covered by `TestEthereumTypeGetNonces_AlternativeProvider_ProviderAnswerRaisedToCacheFloor`.
2. **Per-instance state behind a load balancer** — documented (66c9bca3) in `docs/env.md` and the PR's "Deployment note".
3. **Premature-release window in `releaseRecentSender`** — documented as accepted residual risk in the function comment (66c9bca3), per the comment-only suggestion.
4. **Batched nonce path with a recorded URL** — test added (9b89eb1f); the `withConfirmed` batch branch is now exercised against the accepting provider.

cranycrane:
**[P1] Cross-replica routing state** — partially applied (31cba411). Blockbook has no cross-instance store to share the map through, so instead of new infrastructure the PR adds `*_ALTERNATIVE_NONCE_FOR_ALL_ADDRESSES=TRUE` as an explicit opt-out for deployments without request affinity (websocket flows are naturally sticky, so Suite is unaffected). Documented in `docs/env.md`. *(Superseded — see follow-up comment below.)*

**[P2] Same-second release window** — applied (1a338dbe). Replaced the wall-clock + 1s-slack comparison with a monotonic send-generation counter; sub-second send sequences are now ordered exactly. Remaining documented gap: an *uncached older* send outliving an evicted newer one, largely excluded by nonce sequencing. Tests updated, plus a new test pinning the generation stamping.

cranycrane:
Update on **[P1]**: the `*_ALTERNATIVE_NONCE_FOR_ALL_ADDRESSES` flag was removed again (75774202) — it guarded a deployment shape production does not run (wallet flows are websocket-sticky), and setting it "to be safe" would silently recreate the #1627 quota exhaustion. The per-instance limitation stays documented in `docs/env.md` and the `useForNonces` comment; re-adding an opt-out later is a backward-compatible change if a non-affine deployment ever needs it.

cranycrane:
Confirmed and fixed (8411393e). `handleMempoolTransaction` read the sender's *current* generation after the fetch-back round-trip, so a concurrent send could attach its generation to the wrong transaction — in the described interleaving, A gets cached as gen 2 and its eviction releases the routing while B is still privately pending. Fixed as suggested: `registerSuccessfulSend` returns the generation it assigned and `SendRawTransaction` carries that exact value into `handleMempoolTransaction(txid, gen)`. `TestAlternativeSendTxProviderHandleMempoolTransactionStampsOwnGeneration` encodes this exact interleaving end-to-end.

cranycrane:
1. **Fixed** (07b2c0b2) — the floor now applies whenever a provider is configured, closing the routing-vs-cache expiry window. Covered by `TestEthereumTypeGetNonces_AlternativeProvider_FloorAppliedWithoutRouting`.
2. **Agreed, left as-is** — rare path, uncontended map lookups; merging the gate with URL selection isn't worth the API muddying.
3. **Hardened** (9729aefa) — registration now keys on `acceptedURL != ""` instead of `retErr == nil`, decoupling it from the helper's non-empty-result contract.

pragmaxim (approved):

PR #1630, opened 2026-07-15
feat(xpub): make xpub expansion and cache caps configurable per coin

pragmaxim:
## Summary

Replaces three hard-coded xpub constants (`xpubCacheMaxEntries=128`, `xpubCacheExpirationSeconds=3600`, `maxXpubAddressDerivations=20002`) with a per-coin configurable `XpubConfig`, following the same `additional_params` pattern as `MissingBlockRetry`.

**Motivation:** the 128-entry xpub cache was too small for busy instances. With hundreds of concurrent Trezor Suite users the cache thrashes — every request evicts the previous user's data and forces a full re-derive + re-sort, wasting CPU and defeating the cache. The default is raised to 1024, and operators can now tune all four knobs per coin.

## Changes

- Add `XpubConfig` wire struct (`bchain/types.go`) with `MaxCacheEntries`, `MaxCacheExpirationSeconds`, `DefaultAddressesGap`, `MaxAddressesGap`
- Add `XpubConfig` field + `XpubConfigOverride` method to `BitcoinRPC` and `EthereumRPC` configs; forward through `blockChainWithMetrics`
- Add resolved `XpubConfig` type with `DefaultXpubConfig` / `ApplyXpubConfig` in `api/xpub.go`
- Thread config into `evictXpubCacheItemsLocked`, `trimXpubCacheItemsLocked`, `validateXpubScanLimits`, `xpubScanAddresses` instead of package-level constants
- Resolve config inside `NewWorker` via a duck-typed interface assertion against `chain` (no call-site changes)

### Default config (applies when a coin's JSON has no `xpubConfig` block)

| Field | Default | |
|---|---|---|
| `MaxCacheEntries` | 1024 | was 128 |
| `MaxCacheExpirationSeconds` | 3600 | unchanged |
| `DefaultAddressesGap` | 20 | unchanged |
| `MaxAddressesGap` | 10000 | unchanged |

To override per coin, add to `additional_params` in `configs/coins/*.json`:
```json
"xpubConfig": { "maxCacheEntries": 2048 }
```

## Config invariants

Both the default and override paths run through a single `finalizeXpubConfig` helper that computes the derived `MaxAddressDerivations` from `MaxAddressesGap` and clamps `DefaultAddressesGap` to never exceed `MaxAddressesGap` — so no path can hand `NewWorker` a config that rejects all scans or produces an out-of-range default gap.

## Testing

- `go build ./api/...` passes
- `go test -tags unittest ./api/` — xpub tests pass, including new coverage of the default config path (`TestDefaultXpubConfigDerivedFields`) and the override clamp/derive invariants (`TestApplyXpubConfigClampsAndDerives`)

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -34,6 +34,7 @@ type Worker struct {
 	is                *common.InternalState
 	fiatRates         *fiat.FiatRates
 	metrics           *common.Metrics
+	xpubConfig        XpubConfig
 }
 
 var getTickersForTimestamps = func(fr *fiat.FiatRates, timestamps []int64, vsCurrency string, token string) (*[]*common.CurrencyRatesTicker, error) {
@@ -49,6 +50,17 @@ type contractInfoCache = map[string]*bchain.ContractInfo
 
 // NewWorker creates new api worker
 func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool, txCache *db.TxCache, metrics *common.Metrics, is *common.InternalState, fiatRates *fiat.FiatRates) (*Worker, error) {
+	// Resolve per-chain xpub config override, if any.
+	xpubCfg := DefaultXpubConfig()
+	if provider, ok := chain.(interface {
+		XpubConfigOverride() *bchain.XpubConfig
+	}); ok {
+		if override := provider.XpubConfigOverride(); override != nil {
+			xpubCfg = ApplyXpubConfig(override)
+			glog.Infof("xpub: xpubConfig override applied: maxCacheEntries=%d maxCacheExpirationSeconds=%d defaultAddressesGap=%d maxAddressesGap=%d",
+				xpubCfg.MaxCacheEntries, xpubCfg.MaxCacheExpirationSeconds, xpubCfg.DefaultAddressesGap, xpubCfg.MaxAddressesGap)
+		}
+	}
 	w := &Worker{
 		db:                db,
 		txCache:           txCache,
@@ -60,6 +72,7 @@ func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool, 
 		is:                is,
 		fiatRates:         fiatRates,
 		metrics:           metrics,
+		xpubConfig:        xpubCfg,
 	}
 	if w.chainType == bchain.ChainBitcoinType {
 		w.initXpubCache()
--- a/bchain/types.go
+++ b/bchain/types.go
@@ -448,3 +448,22 @@ type MissingBlockRetry struct {
 	// yields errResync to the outer machinery.
 	MaxStallMs int `json:"maxStallMs,omitempty"`
 }
+
+// XpubConfig carries per-chain overrides for xpub address expansion and
+// caching. All fields are optional; zero / missing values fall back to the
+// api package's built-in defaults. Operators set this under
+// `additional_params.xpubConfig` in `configs/coins/*.json`.
+type XpubConfig struct {
+	// MaxCacheExpirationSeconds is the TTL for cached xpub-to-address
+	// derivations (seconds). Zero / missing uses the default (3600).
+	MaxCacheExpirationSeconds int `json:"maxCacheExpirationSeconds,omitempty"`
+	// MaxCacheEntries is the hard cap on the number of cached xpubs.
+	// Zero / missing uses the default (1024).
+	MaxCacheEntries int `json:"maxCacheEntries,omitempty"`
+	// DefaultAddressesGap is the default BIP44 gap limit when the
+	// caller does not supply one (zero / missing → 20).
+	DefaultAddressesGap int `json:"defaultAddressesGap,omitempty"`
+	// MaxAddressesGap is the maximum user-supplied gap value the
+	// server will accept (zero / missing → 10000).
+	MaxAddressesGap int `json:"maxAddressesGap,omitempty"`
+}
--- a/api/xpub.go
+++ b/api/xpub.go
@@ -15,15 +15,63 @@ import (
 	"github.com/trezor/blockbook/db"
 )
 
-const defaultAddressesGap = 20
-const maxAddressesGap = 10000
-
 const txInput = 1
 const txOutput = 2
 
-const xpubCacheExpirationSeconds = 3600
-const xpubCacheMaxEntries = 128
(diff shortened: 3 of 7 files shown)


api/worker.go
```diff
@@ -49,6 +50,15 @@ type contractInfoCache = map[string]*bchain.ContractInfo
 
 // NewWorker creates new api worker
 func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool, txCache *db.TxCache, metrics *common.Metrics, is *common.InternalState, fiatRates *fiat.FiatRates) (*Worker, error) {
+	// Resolve per-chain xpub config override, if any.
+	xpubCfg := DefaultXpubConfig()
+	if provider, ok := chain.(interface {
+		XpubConfigOverride() *bchain.XpubConfig
+	}); ok {
+		if override := provider.XpubConfigOverride(); override != nil {
+			xpubCfg = ApplyXpubConfig(override)
```
cranycrane, line 59: **Log the effective config at startup.** This mirrors the `MissingBlockRetry` pattern, and that path logs one `sync: missingBlockRetry override applied: …` line at startup so operators can confirm the resolved values (see `docs/sync.md`). Here the only logging is a warning on *negative* input in `ApplyXpubConfig`. Since the whole motivation is operator tuning of busy instances, emitting one line with the four effective values (at least when an override is present) would match the precedent and make misconfiguration diagnosable.

For example, when `override != nil`:
```go
xpubCfg = ApplyXpubConfig(override)
glog.Infof("xpub: xpubConfig override applied: maxCacheEntries=%d maxCacheExpirationSeconds=%d defaultAddressesGap=%d maxAddressesGap=%d",
    xpubCfg.MaxCacheEntries, xpubCfg.MaxCacheExpirationSeconds, xpubCfg.DefaultAddressesGap, xpubCfg.MaxAddressesGap)
```

bchain/types.go
```diff
@@ -448,3 +448,22 @@ type MissingBlockRetry struct {
 	// yields errResync to the outer machinery.
 	MaxStallMs int `json:"maxStallMs,omitempty"`
 }
+
+// XpubConfig carries per-chain overrides for xpub address expansion and
+// caching. All fields are optional; zero / missing values fall back to the
+// api package's built-in defaults. Operators set this under
+// `additional_params.xpubConfig` in `configs/coins/*.json`.
+type XpubConfig struct {
```
cranycrane, line 456: **Add operator-facing documentation.** `MissingBlockRetry` — the pattern this mirrors — ships a full knob table + JSON example in `docs/sync.md`, and `additional_params` is described in `docs/config.md`. `XpubConfig` currently has no discoverable reference outside this struct's comments and the PR body: operators have nowhere to find the four knobs, their defaults, the 128→1024 change, or the clamp behavior. Please add a short section (mirroring the `missingBlockRetry` table in `docs/sync.md`, or a subsection in `docs/config.md`) covering the fields, defaults, the `<= 0` → default semantics, and a note that raising `maxCacheEntries` increases resident memory so hosts can be sized accordingly.

api/worker.go
pragmaxim: Done in e08a65c — `NewWorker` now logs one `xpub: xpubConfig override applied: maxCacheEntries=… maxCacheExpirationSeconds=… defaultAddressesGap=… maxAddressesGap=…` line at startup when an override is present, matching the `missingBlockRetry` precedent.

bchain/types.go
pragmaxim: Done in e08a65c — added an `xpubConfig` subsection under `additional_params` in `docs/config.md`: a knob table with the four fields and their defaults, the `<= 0` → default (and negative → ignored+logged) semantics, a note that raising `maxCacheEntries` increases resident memory so hosts can be sized accordingly, and a JSON override example.

bchain/coins/eth/ethrpc.go
```diff
@@ -317,6 +320,16 @@ func (b *EthereumRPC) MissingBlockRetryOverride() *bchain.MissingBlockRetry {
 	return b.ChainConfig.MissingBlockRetry
 }
 
+// XpubConfigOverride exposes the per-chain xpub expansion and cache config
+// override (or nil to use built-in defaults). Consumed by api.NewWorker via
+// a duck-typed interface assertion.
+func (b *EthereumRPC) XpubConfigOverride() *bchain.XpubConfig {
```
cranycrane: **This (and the `XpubConfig` field at line 117) is dead config on EVM chains — consider dropping it.** xpub / BIP44 gap-scanning is a UTXO-only feature in Blockbook. The cache is only initialized for Bitcoin-type chains (`api/worker.go`: `if w.chainType == bchain.ChainBitcoinType { w.initXpubCache() }`), and the sole reader, `getXpubData`, bails out immediately for anything else (`api/xpub.go`: `if w.chainType != bchain.ChainBitcoinType { return ..., ErrUnsupportedXpub }`).

So on an `EthereumRPC`-backed coin, `XpubConfigOverride()` is called and `ApplyXpubConfig` runs, but the resolved config is stored on the worker and **never read**. Unlike `MissingBlockRetry` (which EVM chains genuinely use via the shared sync worker), this got copied to `eth.Configuration` mechanically where it has no effect.

It's functionally harmless, but misleading: an operator can add an `xpubConfig` block to an EVM coin's config and it will be silently ignored — no error, no log line.

Suggest removing the `XpubConfig` field and this `XpubConfigOverride()` method from `eth/ethrpc.go` and keeping the override Bitcoin-only. The duck-typed assertion in `NewWorker` degrades gracefully: if `EthereumRPC` doesn't implement the method, the assertion simply fails and the worker falls back to `DefaultXpubConfig()` (which it never reads on EVM anyway). Nothing else needs to change.

pragmaxim: Good catch — done in aae3a52. Removed the `XpubConfig` field and `XpubConfigOverride()` method from `eth/ethrpc.go`; xpub scanning is UTXO-only, so the config was stored on the worker but never read on EVM. The override is now Bitcoin-only. The duck-typed assertion in `blockChainWithMetrics.XpubConfigOverride()` / `NewWorker` degrades gracefully — `EthereumRPC` no longer implements the method, so the assertion fails and the worker falls back to `DefaultXpubConfig()` (which EVM never reads anyway). Also updated the commit message to reflect Bitcoin-only.

cranycrane (approved):
LGTM

PR #1632, opened 2026-07-16
fix(eth): treat "contract not exist" eth_call errors as non-retriable

cranycrane:
## Problem

On a TRON Blockbook deployment, the logs show pairs of warnings for token `balanceOf` calls:

```
W contract.go:629] erc20 batch eth_call failed for 0x70a9…cc29: Smart contract is not exist.
W contract.go:633] erc20 single eth_call fallback failed for 0x70a9…cc29: Smart contract is not exist.
```

These occur when an address recorded as a token contract isn't backed by live code on the node — typically self-destructed contracts, or **fake/spam TRC20 transfer logs** referencing non-contract addresses (very common on TRON). java-tron answers such `eth_call`s with `Smart contract is not exist`.

`isNonRetriableEthCallError` did not recognize this message, so the batch fallback classified it as *retriable* and re-issued each one as a single call. That retry is deterministic and guaranteed to fail — it only wastes an RPC round-trip and logs a second warning per contract.

This is not a Trezor Suite or node bug: Suite just requests token balances, and the node correctly reports the address isn't a contract. It's benign Blockbook log-noise plus wasted RPC.

## Fix

Classify `smart contract is not exist` (and the generic `no contract` / `not deployed` variants from other backends) as non-retriable. The batch element is then skipped instead of retried, removing both the redundant call and both warning lines. Account-info responses are unchanged — the balance stays `nil` for such contracts, exactly as before.

## Tests

- `TestErc20BalancesBatchNonRetriableSkipsFallback` — feeds the real TRON error string as a batch-element error and asserts **zero** single-call fallbacks and a `nil` balance.
- `TestIsNonRetriableEthCallError` — table test covering the new strings, the pre-existing ones, and negative cases (transient RPC / timeout stay retriable).

### DIFF
--- a/bchain/coins/eth/contract.go
+++ b/bchain/coins/eth/contract.go
@@ -663,7 +663,11 @@ func isNonRetriableEthCallError(err error) bool {
 		strings.Contains(msg, "invalid opcode") ||
 		strings.Contains(msg, "out of gas") ||
 		strings.Contains(msg, "stack underflow") ||
-		strings.Contains(msg, "revert")
+		strings.Contains(msg, "revert") ||
+		// Java-Tron throws these messages
+		strings.Contains(msg, "smart contract is not exist") ||
+		strings.Contains(msg, "no contract") ||
+		strings.Contains(msg, "not deployed")
 }
 
 // GetTokenURI returns URI of non fungible or multi token defined by token id
--- a/bchain/coins/eth/contract_batch_test.go
+++ b/bchain/coins/eth/contract_batch_test.go
@@ -499,3 +499,74 @@ func TestEthereumTypeGetErc20ContractBalancesPartialError(t *testing.T) {
 		t.Fatalf("expected balance[1] to be nil, got %v", balances[1])
 	}
 }
+
+// A deterministic per-element error (e.g. TRON's "Smart contract is not exist"
+// for a dead/fake TRC20 contract) must not trigger a single-call fallback: the
+// retry cannot succeed and only amplifies RPC load and log noise.
+func TestErc20BalancesBatchNonRetriableSkipsFallback(t *testing.T) {
+	addr := common.HexToAddress("0x0000000000000000000000000000000000000011")
+	contractA := common.HexToAddress("0x00000000000000000000000000000000000000aa")
+	contractB := common.HexToAddress("0x00000000000000000000000000000000000000bb")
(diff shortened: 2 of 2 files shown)


bchain/coins/eth/contract.go
```diff
@@ -663,7 +663,11 @@ func isNonRetriableEthCallError(err error) bool {
 		strings.Contains(msg, "invalid opcode") ||
 		strings.Contains(msg, "out of gas") ||
 		strings.Contains(msg, "stack underflow") ||
-		strings.Contains(msg, "revert")
+		strings.Contains(msg, "revert") ||
+		// Java-Tron throws these messages
+		strings.Contains(msg, "smart contract is not exist") ||
```
pragmaxim, line 668: **Issue: Missing Nethermind pattern — `"no code at given address"`**

Nethermind and some geth versions return `"no code at given address"` when `eth_call` targets a non-existent contract. This string contains **neither** `"no contract"` **nor** `"not deployed"` **nor** `"smart contract is not exist"`. On Nethermind-backed chains, every dead-contract `balanceOf` call will still trigger the useless single-call fallback retry — exactly the problem this PR aims to solve.

**Recommendation**: Add `"no code at given address"` to the non-retriable patterns in `isNonRetriableEthCallError`. This is a well-known, deterministic geth/Nethermind error that is safe to treat as non-retriable.

bchain/coins/eth/contract.go
```diff
@@ -663,7 +663,11 @@ func isNonRetriableEthCallError(err error) bool {
...
+		strings.Contains(msg, "no contract") ||
+		strings.Contains(msg, "not deployed")
```
pragmaxim, line 670: **🔴 Issue: Pattern `"not deployed"` is dangerously broad**

The substring `"not deployed"` matches any error message containing those words — including transient conditions like `"contract not deployed yet"` during node sync/reorg or `"contract deployment not finished yet"`. These **could resolve on retry**, but now the code silently produces a `nil` balance with no warning log and no metric. Misclassification is invisible to operators until users report missing token balances.

**Recommendation**: Either (a) match a more specific phrase (e.g., `"is not deployed"`), or (b) remove this pattern entirely if no canonical TRON client emits it exactly. At minimum, add a warning log on the non-retriable skip path to make false positives detectable.

cranycrane:
not needed anymore

PR #1633, opened 2026-07-17
fix(btc): accept array-form "warnings" from Bitcoin Core 25.0+

torkelrogstad:
Bitcoin Core 25.0 changed the "warnings" field of getblockchaininfo and getnetworkinfo from a single string to an array of strings. Blockbook types the field as a plain string, so GetChainInfo fails against any Core 25+ node with:

  cannot unmarshal array into Go struct field .result.warnings of type string

This blocks initialization: getBlockChainWithRetry never succeeds, so blockbook cannot connect to a modern bitcoind (reproduced against Core 31.1).

Add a coreWarnings type whose UnmarshalJSON accepts either shape: an array (joined with spaces) or the legacy string (still produced by -deprecatedrpc=warnings). Use it for the "warnings" field of ResGetBlockChainInfo and ResGetNetworkInfo, casting back to string at the use sites. zec reuses btc's ResGetNetworkInfo and gets the same cast.

Tests cover both wire shapes (string, empty/single/multi-element array, null) plus a struct-level regression for the getblockchaininfo array case.

### DIFF
--- a/bchain/coins/btc/bitcoinrpc.go
+++ b/bchain/coins/btc/bitcoinrpc.go
@@ -8,6 +8,7 @@ import (
 	"math/big"
 	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


pragmaxim (approved):
Looks good, thank you @torkelrogstad ... Did you actually sync blockbook with Core 31.1 up to chain tip?

cranycrane (approved):
verified, LGTM

PR #1635, opened 2026-07-18
feat(explorer): collapse long input/output lists in tx detail

ranjbar-dev:
Transactions with more than 8 inputs or outputs collapse behind a toggle by default, so scrolling past a large tx (e.g. a batch payout or coinjoin) on the address/block/xpub pages doesn't require paging through hundreds of rows. Uses Bootstrap's existing collapse component already loaded site-wide, no new JS.


j1warren:
applied it to 0.4.0 setup, works good!

2 remarks: I don't like how the button looks and it doesn't take you back to the same tx when you collapse the list.
Also when expanded it may be convenient to have buttons (to collapse it back) both at start and at the end of the list.

pragmaxim:
Thanks for this — nice, self-contained improvement, and reusing the existing Bootstrap collapse (no new JS) is a clean approach. 👍

One thing I wanted to check on the default-collapsed behavior, mostly for the aggregate pages (address / xpub / block) rather than the standalone tx page:

Since Bootstrap `collapse` uses `display:none`, a collapsed `>8` input/output list is skipped by the browser's in-page find (Ctrl+F), and the `tx-own` ("your address") highlight won't show until the user expands. On the address/xpub pages people often scan specifically for *their own* address across many txs, so hiding those rows by default could occasionally hide the very thing they're looking for.

Was collapsing-by-default on those aggregate pages intentional, or would it make sense to only collapse on the standalone tx detail page (or default-expanded there)? Happy either way — just wanted to make sure the trade-off was considered. Thanks again!

ranjbar-dev:
Good catch @pragmaxim , and no — I hadn't fully considered the Ctrl+F / tx-own interaction on the aggregate pages. That's a real trade-off.
I think the cleanest fix is to only collapse-by-default on the standalone tx detail page, and leave address/xpub/block pages expanded (or raise the threshold there significantly, e.g. only collapse if >20-30 rows). The original goal was really about the standalone tx page anyway — that's where a single coinjoin/batch payout tx dominates the whole page. On the aggregate pages the list is usually shorter per-tx already, so collapsing there adds an extra click for the "did I get paid" scan without saving much space.
I'll push an update that scopes the default-collapse to the tx detail view only. Let me know if you'd rather see a different threshold instead of removing it entirely on those pages.

ranjbar-dev:
@j1warren  Thanks for testing it on 0.4.0! Addressing both:

Button styling — will restyle to match the existing UI more closely rather than the default Bootstrap toggle look.
Collapsing back to the same tx / duplicate toggle at bottom — good idea, I'll add a matching "collapse" control at the end of the list so you don't have to scroll back up to close it.

pragmaxim:
One small thing on the new bottom toggle: it's rendered as a sibling *after* the collapsible `<div>` (outside it), so when the list is collapsed both the top and bottom buttons stay visible — you end up with two identical "N inputs ▼" buttons stacked with the hidden list between them.

Since the bottom control is really only useful for collapsing *after* expanding, would it make sense to move it *inside* the collapsible div (as its last child)? That way it only appears while the list is expanded, and the collapsed state shows just the single top toggle. Thanks again for iterating on this!

ranjbar-dev:
Good catch — bottom toggle should've been nested inside the collapsible <div>, not a sibling. Fixed, pushed now. @pragmaxim

pragmaxim (approved):
LGTM

j1warren:
I'm testing this on https://btc1.trezor.io/ and I don't see the collapse button scrolling back to tx. Was it working before?

ranjbar-dev:
> I'm testing this on https://btc1.trezor.io/ and I don't see the collapse button scrolling back to tx. Was it working before?

yes it was working about week a go that i was using the btc1.trezor.io

j1warren:
https://btc1.trezor.io/tx/8538acbb3ccb8aab30eeff1e13709b669fb272baf37f9711c53b5e017935ac7b#txvout-8538acbb3ccb8aab30eeff1e13709b669fb272baf37f9711c53b5e017935ac7b

jumping back on collapsing doesn't work for me on FF and Vivaldi

cranycrane:
hey guys, you are right, we forgot to regenerate the assets so the styles are applied - https://github.com/trezor/blockbook/pull/1777

PR #1637, opened 2026-07-21
feat(hyperevm): add Hyperliquid (HyperEVM) mainnet support

cranycrane:
## Summary

Adds Blockbook indexing support for **HyperEVM** — Hyperliquid's EVM layer (mainnet chain ID `999`, native token HYPE, EIP-1559). Mainnet-only, self-hosted archive node, following the same integration surface as the existing Ethereum-L2 coins.

## What's included

- **RPC shim** (`bchain/coins/hyperevm`) embedding the Ethereum driver. A custom `Initialize` accepts network id `999` and uses the Arbitrum idiom (`MainNetChainID = 999`) so `getInfo` reports `mainnet` and Ethereum-only ENS stays disabled.
- **`GetBlockHash` override** returning the hash the backend reports (`eth_getBlockByNumber(...).hash`) instead of go-ethereum's RLP recomputation of the header. The two disagree at the HyperEVM genesis — go-ethereum recomputes `0x0466…8f8bd5` while reth-hl's canonical hash is `0xd8fcc13b…895f0` (zero-valued Prague/Cancun genesis fields folded into the hash under different rules), so sync asked for a block by the mis-derived hash, got `null`, and wedged at height 0. Blocks are stored under the backend-reported hash, so reading that same field at *every* height keeps the stored hash and all later tip/reorg comparisons consistent by construction, instead of depending on reth-hl and go-ethereum agreeing for non-genesis headers. No extra RPC cost — `HeaderByNumber` issues the same `eth_getBlockByNumber(height, false)` call.
- **Coin config** `hyperevm_archive`: EIP-1559 fees with the Infura Gas API (`networks/999`) as the alternative provider and native `eth_feeHistory` as fallback, 1 s average block time, mempool sync disabled (nanoreth is not a mempool participant), CoinGecko `hyperevm`/`hyperliquid` fiat rates.
- **Backend**: nanoreth (`reth-hl`) serving HTTP **+ WebSocket** JSON-RPC (Blockbook dials a WebSocket at startup) and ingesting pre-executed blocks + receipts from Hyperliquid's S3 stream. Installs the pinned release binary `nb-20260613` (amd64), verified by sha256.
- Regenerated `docs/ports.md`, an integration-test entry, and the captured RPC fixture.

## Why nanoreth (WebSocket + archive)

Blockbook's Ethereum driver dials a `wss://` endpoint at startup and needs historical state. HyperEVM's public RPC has neither (HTTP-only, ~12 h retention). nanoreth is the self-hostable node that provides both.

## Fees

`eip1559Fees: true` plus `alternative_estimate_fee: infura` against `https://gas.api.infura.io/v3/${api_key}/networks/999/suggestedGasFees` — chain 999 is in the Infura Gas API's supported-network list, matching how the other archive EVM coins are configured. Consequence for deploy: `INFURA_API_KEY` must be present in the environment or chain initialization fails fast (`EthereumRPC.initAlternativeFeeProvider`). When the provider's cache is stale or unready, fee estimation falls back to on-chain `eth_feeHistory`.

## Internal transactions

`processInternalTransactions: true` with `debug,trace` in the HTTP/WS API lists. nanoreth supports `debug_traceBlockByHash`, so `callTracer`-based internal-tx extraction works on this backend. A trace failure would not break sync — `GetBlock` records it as `InternalDataError` with bounded retry — but it is not expected to be needed here.

## Test status

Full CI (unit → connectivity → integration) is green for `hyperevm_archive`: connectivity runs with `BB_TEST_BACKEND_CONNECTIVITY=1` against the node RPC and the Blockbook API, and the RPC suite passes against the captured `tests/rpc/testdata/hyperevm_archive.json`. Dev/staging currently points Blockbook at a hosted nanoreth RPC endpoint.

## Still open (self-hosted deploy)

- [ ] Provision AWS S3 credentials for the backend system user (deploy-time secret, not shipped) — needed for the self-hosted nanoreth node, not for the current staging setup
- [ ] Confirm `debug`/`trace` on the self-hosted `reth-hl` build once it is running (documented as supported by nanoreth; validated so far against a hosted nanoreth endpoint)
- [ ] arm64: the pinned release is amd64-only — build from source if deploying on arm64

## Notes

- The `docs/ports.md` diff also corrects pre-existing staleness in unrelated rows (the port-registry generator is the canonical source for that file).

### DIFF
--- a/bchain/coins/hyperevm/hyperevmrpc.go
+++ b/bchain/coins/hyperevm/hyperevmrpc.go
@@ -76,6 +76,51 @@ func (b *HyperevmRPC) Initialize() error {
 	return nil
 }
 
+// GetBlockHash returns the hash of the block at the given height.
+//
+// For every height except genesis this defers to the base implementation, which
+// derives the hash via go-ethereum's HeaderByNumber().Hash() — i.e. it RLP-hashes
+// the header fields rather than trusting the node's "hash" field. That works for
+// blocks >= 1, but for the HyperEVM genesis go-ethereum recomputes 0x0466…8f8bd5,
+// which does NOT match reth-hl's canonical genesis hash 0xd8fcc13b…895f0 (the
+// genesis header carries zero-valued Prague/Cancun fields that reth-hl folded into
+// the canonical hash under different rules). Sync then asks the backend for a block
+// by that mis-derived hash, gets null, and wedges at height 0 forever.
+//
+// Return the hash the backend actually reports for the genesis header instead, so
+// the stored genesis and every subsequent tip/reorg comparison stay consistent.
+func (b *HyperevmRPC) GetBlockHash(height uint32) (string, error) {
+	if height == 0 {
+		raw, err := b.GetBlockRawByHashOrHeight("", 0, false)
+		if err != nil {
+			return "", errors.Annotate(err, "genesis")
+		}
+		var h struct {
+			Hash string `json:"hash"`
+		}
+		if err := json.Unmarshal(raw, &h); err != nil {
+			return "", errors.Annotate(err, "genesis")
+		}
+		if h.Hash == "" {
+			return "", bchain.ErrBlockNotFound
+		}
+		return h.Hash, nil
+	}
+	return b.EthereumRPC.GetBlockHash(height)
+}
+
+// GetBlock returns the block of the given hash and height. Genesis is fetched by
+// number (empty hash routes getBlockRaw to eth_getBlockByNumber) because
+// GetBlockHash's caller passes go-ethereum's mis-derived genesis hash; fetching by
+// number sidesteps the eth_getBlockByHash lookup that would return null. All other
+// heights keep the hash-based path so reorg detection during sync stays intact.
+func (b *HyperevmRPC) GetBlock(hash string, height uint32) (*bchain.Block, error) {
+	if height == 0 {
+		hash = ""
+	}
+	return b.EthereumRPC.GetBlock(hash, height)
+}
+
 func (b *HyperevmRPC) ResolveENS(name string) (*bchain.ENSResolution, error) {
 	return b.EthereumRPC.ResolveENS(name)
 }
--- a/bchain/coins/hyperevm/hyperevmrpc.go
+++ b/bchain/coins/hyperevm/hyperevmrpc.go
@@ -1,0 +1,81 @@
+package hyperevm
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/golang/glog"
+	"github.com/juju/errors"
+	"github.com/trezor/blockbook/bchain"
+	"github.com/trezor/blockbook/bchain/coins/eth"
+)
+
+const (
+	// MainNet is the chain ID of HyperEVM (Hyperliquid) mainnet
+	MainNet eth.Network = 999
+)
+
+// HyperevmRPC is an interface to JSON-RPC HyperEVM service.
+type HyperevmRPC struct {
+	*eth.EthereumRPC
+}
+
+// NewHyperevmRPC returns new HyperevmRPC instance.
+func NewHyperevmRPC(config json.RawMessage, pushHandler func(bchain.NotificationType)) (bchain.BlockChain, error) {
+	c, err := eth.NewEthereumRPC(config, pushHandler)
+	if err != nil {
+		return nil, err
+	}
+
+	s := &HyperevmRPC{
+		EthereumRPC: c.(*eth.EthereumRPC),
+	}
+
+	return s, nil
+}
+
+// Initialize hyperevm rpc interface
+func (b *HyperevmRPC) Initialize() error {
+	b.OpenRPC = eth.OpenRPC
+
+	rc, ec, err := b.OpenRPC(b.ChainConfig.RPCURL, b.ChainConfig.RPCURLWS)
+	if err != nil {
+		return err
+	}
+
+	// set chain specific
+	b.Client = ec
+	b.RPC = rc
+	b.NewBlock = eth.NewEthereumNewBlock()
+	b.NewTx = eth.NewEthereumNewTx()
+
+	ctx, cancel := context.WithTimeout(context.Background(), b.Timeout)
+	defer cancel()
+
+	id, err := b.Client.NetworkID(ctx)
+	if err != nil {
+		return err
+	}
+
+	// parameters for getInfo request
+	switch eth.Network(id.Uint64()) {
+	case MainNet:
+		b.MainNetChainID = MainNet
+		b.Testnet = false
+		b.Network = "livenet"
+	default:
+		return errors.Errorf("Unknown network id %v", id)
+	}
+
+	if err = b.InitAlternativeProviders(); err != nil {
+		return err
+	}
+
+	glog.Info("rpc: block chain ", b.Network)
+
+	return nil
+}
+
+func (b *HyperevmRPC) ResolveENS(name string) (*bchain.ENSResolution, error) {
+	return b.EthereumRPC.ResolveENS(name)
+}
--- a/bchain/coins/hyperevm/hyperevmrpc.go
+++ b/bchain/coins/hyperevm/hyperevmrpc.go
@@ -109,18 +109,6 @@ func (b *HyperevmRPC) GetBlockHash(height uint32) (string, error) {
 	return b.EthereumRPC.GetBlockHash(height)
 }
 
-// GetBlock returns the block of the given hash and height. Genesis is fetched by
-// number (empty hash routes getBlockRaw to eth_getBlockByNumber) because
-// GetBlockHash's caller passes go-ethereum's mis-derived genesis hash; fetching by
-// number sidesteps the eth_getBlockByHash lookup that would return null. All other
-// heights keep the hash-based path so reorg detection during sync stays intact.
-func (b *HyperevmRPC) GetBlock(hash string, height uint32) (*bchain.Block, error) {
-	if height == 0 {
-		hash = ""
-	}
-	return b.EthereumRPC.GetBlock(hash, height)
-}
-
 func (b *HyperevmRPC) ResolveENS(name string) (*bchain.ENSResolution, error) {
 	return b.EthereumRPC.ResolveENS(name)
 }
--- a/bchain/coins/hyperevm/hyperevmrpc.go
+++ b/bchain/coins/hyperevm/hyperevmrpc.go
@@ -76,37 +76,27 @@ func (b *HyperevmRPC) Initialize() error {
 	return nil
 }
 
-// GetBlockHash returns the hash of the block at the given height.
-//
-// For every height except genesis this defers to the base implementation, which
-// derives the hash via go-ethereum's HeaderByNumber().Hash() — i.e. it RLP-hashes
-// the header fields rather than trusting the node's "hash" field. That works for
-// blocks >= 1, but for the HyperEVM genesis go-ethereum recomputes 0x0466…8f8bd5,
-// which does NOT match reth-hl's canonical genesis hash 0xd8fcc13b…895f0 (the
-// genesis header carries zero-valued Prague/Cancun fields that reth-hl folded into
-// the canonical hash under different rules). Sync then asks the backend for a block
-// by that mis-derived hash, gets null, and wedges at height 0 forever.
-//
-// Return the hash the backend actually reports for the genesis header instead, so
-// the stored genesis and every subsequent tip/reorg comparison stay consistent.
+// GetBlockHash returns the hash reported by the backend, not go-ethereum's RLP
+// recomputation of the header used by the base implementation. The two disagree at the
+// HyperEVM genesis, which wedged sync at height 0. Blocks are stored under the reported
+// hash, so read that field at every height to keep the comparisons consistent.
 func (b *HyperevmRPC) GetBlockHash(height uint32) (string, error) {
-	if height == 0 {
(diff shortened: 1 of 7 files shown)


tests/tests.json
```diff
@@ -314,6 +314,13 @@
                 "WsGetInfo", "WsGetBlockHash", "WsGetTransaction", "WsGetAccountInfoBasicEVM", "WsGetAccountInfoConfirmedNonceEVM","WsGetAccountInfoEVM", "WsGetAccountInfoTxidsConsistencyEVM", "WsGetAccountInfoTxsConsistencyEVM", "WsGetAccountInfoContractFilterEVM", "WsPing", "WsGetBlock", "WsGetBalanceHistory", "WsGetTransactionSpecific"],
         "rpc": ["GetBlock", "GetBlockHash", "GetTransaction", "EstimateFee", "GetBlockHeader", "EthCallBatch"]
     },
+    "hyperevm_archive": {
```
pragmaxim, line 317: ~~**CI will fail: enabled entry with no deployed backend + skeleton fixture.**~~

**Retracted — I was wrong.** I'd assumed the tests would dial the config's `127.0.0.1` template and fail without a deployed nanoreth. They won't: `lookupEnvWithArchiveFallback` (build/tools/templates.go) overrides the RPC URL from `BB_DEV_RPC_URL_HTTP_HYPEREVM_ARCHIVE` / `..._WS_...`, which CI exports from the repo variables — and **both of those already exist** and point at a QuickNode `hype-mainnet` endpoint.

I verified the endpoint end to end, and it covers everything this entry exercises:

| check | result |
| --- | --- |
| `eth_chainId` | `0x3e7` = 999 ✓ |
| WebSocket upgrade | `101 Switching Protocols` ✓ |
| archive state (`eth_getBalance` @ block `0x1`) | returns a value, no "missing trie node" ✓ |
| `debug_traceBlockByHash` + `callTracer` | returns full nested traces ✓ |

And the fixture is **not** a skeleton — `tests/rpc/testdata/hyperevm_archive.json` matches the live endpoint exactly at height 42176091: hash `0x14e51adc…5217`, time `1785740657`, 7 txs, size 4251, and `debug_traceBlockByHash` on that block returns the fixture's first txid. It was clearly captured from this endpoint.

So no `"disabled": true` needed — this entry should pass as-is. Sorry for the noise.

bchain/coins/hyperevm/hyperevmrpc.go
```diff
@@ -0,0 +1,114 @@
...
+// the header fields rather than trusting the node's "hash" field. That works for
+// blocks >= 1, but for the HyperEVM genesis go-ethereum recomputes 0x0466…8f8bd5,
+// which does NOT match reth-hl's canonical genesis hash 0xd8fcc13b…895f0 (the
+// genesis header carries zero-valued Prague/Cancun fields that reth-hl folded into
+// the canonical hash under different rules). Sync then asks the backend for a block
+// by that mis-derived hash, gets null, and wedges at height 0 forever.
+//
+// Return the hash the backend actually reports for the genesis header instead, so
+// the stored genesis and every subsequent tip/reorg comparison stay consistent.
+func (b *HyperevmRPC) GetBlockHash(height uint32) (string, error) {
```
pragmaxim, line 83: ~~**Genesis-only override may be insufficient — validate non-genesis hashes.**~~

**Tested — the genesis-only scope is correct. Keep this exactly as written.**

I was concerned that reconciling only height 0 left N≥1 depending on go-ethereum's RLP recomputation matching reth-hl's canonical hash, and that any divergence there would make `handleFork` (db/sync.go:333-349) disconnect to genesis on every resync. So I compared `HeaderByNumber(n).Hash()` — exactly what `eth.EthereumRPC.GetBlockHash` returns — against the node-reported `hash`, across the chain, on the QuickNode `hype-mainnet` endpoint the RPC vars point at:

| height | result |
| --- | --- |
| **0** | **DIVERGE** — derived `0x0466e6a9…8f8bd5`, node `0xd8fcc13b…895f0` |
| 1, 2, 3 | match |
| 1 000 / 100 000 / 1 000 000 / 10 000 000 | match |
| 42176091 (fixture block) | match |
| tip−100, tip−5 (tip 42 271 537) | match |

1 mismatch in 11 — and it's precisely height 0, with exactly the two hashes named in your comment. So the diagnosis is right, the fix is right, and the genesis-only scope is right: post-genesis headers RLP-recompute consistently, so the base implementation's cheaper `HeaderByNumber` path is correct for N≥1.

Disregard my suggestion to return the node-reported hash for all heights — it would trade a header fetch for a full block fetch on the fork-scan path to solve a problem that doesn't exist. This is the better implementation.

One note for the future rather than now: this validates the endpoint actually in the serving path. If nanoreth ever becomes the backend, re-run the same comparison there, since a different node implementation is exactly where this class of divergence originates.

pragmaxim (commented):
**Superseded — see the APPROVED review below (LGTM).** Every finding in this pass was tested against the live QuickNode `hype-mainnet` endpoint and retracted; the inline comments are amended individually with the evidence. Leaving them in place rather than deleting so the reasoning is auditable.

configs/coins/hyperevm_archive.json
```diff
@@ -0,0 +1,82 @@
...
+                "averageBlockTimeMs": 1000,
+                "missingBlockRetry": {
+                    "retryDelayMs": 1000,
+                    "recheckThreshold": 10,
+                    "tipRecheckThreshold": 3,
+                    "maxStallMs": 60000
+                },
+                "address_aliases": true,
+                "eip1559Fees": true,
+                "alternative_estimate_fee": "infura",
```
pragmaxim, line 64: **Fee config contradicts the PR description; Infura coverage for chain 999 is undocumented.**

_Edited: I originally wrote that Infura's gas API is "unlikely to serve chain 999." I've since tested that and it was unsupported speculation — retracted, with what I actually verified below._

The substantive point stands: the PR summary says "native `eth_feeHistory` EIP-1559 fees (no external provider)", but this selects the Infura gas API for `networks/999`. Either the summary or the config should change.

Also note `initAlternativeFeeProvider` (eth/ethrpc.go:891) **fails fast** without `INFURA_API_KEY` — this makes the key a hard startup dependency on deploy, not a soft enhancement.

What I verified against Infura:

- **The URL form is correct and current.** `https://gas.api.infura.io/v3/${api_key}/networks/${chainId}/suggestedGasFees` is exactly the shape documented in the Gas API quickstart. No issue there.
- **Infura supports HyperEVM as a first-class network** — dedicated reference section, mainnet HTTPS + WebSocket (`wss://hyperevm-mainnet.infura.io/ws/v3/<KEY>`), a testnet endpoint, and HyperEVM-specific methods such as `eth_bigblockgasprice`. So Gas API coverage for 999 is entirely plausible.
- **But Gas API network coverage is undocumented.** The Gas API quickstart points to `/get-started/endpoints/#gas-api` as the supported-network list, and that table is **empty** (header row only, in both the rendered page and the `.md` export). All Gas API pages use `chainId = 1` in examples and never enumerate networks.
- **It can't be probed without a valid key.** The API validates only that the chain ID is numeric (`/networks/abc/...` → 404) and resolves network support strictly *after* auth, so `1`, `999`, `4663` and even a nonexistent `999999` all return an identical 401. No differential to exploit.

So this needs a real key to settle. The cheapest check is the existing metric `eth_eip1559_fee_source_total{source=...}` (common/metrics.go:103): if `source="provider"` increments, Infura serves 999; if only `onchain_fallback` ever increments, the provider config is dead weight and `eip1559Fees: true` alone already gives the native `eth_feeHistory` fees the summary describes.

Worth running that same check against **robinhood/4663**, which uses the identical pattern — because the fallback is graceful, a non-functioning provider config is otherwise silent.

configs/coins/hyperevm_archive.json
```diff
@@ -0,0 +1,82 @@
...
+                "alternative_estimate_fee_params": "{\"url\": \"https://gas.api.infura.io/v3/${api_key}/networks/999/suggestedGasFees\", \"periodSeconds\": 10}",
+                "mempoolTxTimeoutHours": 12,
+                "processInternalTransactions": true,
```
pragmaxim, line 67: ~~**`processInternalTransactions: true` is unvalidated against nanoreth.**~~

**Retracted — tested and it works.** I ran the exact call Blockbook makes for internal transactions (`debug_traceBlockByHash` with `{"tracer": "callTracer"}`, ethrpc.go:1520) against the QuickNode `hype-mainnet` endpoint that `BB_DEV_RPC_URL_HTTP_HYPEREVM_ARCHIVE` and `BB_PROD_RPC_URL_HTTP_HYPEREVM_ARCHIVE` both point at. It returns full nested `calls` trees, including on the fixture block. So `processInternalTransactions: true` is correct as configured — leave it on.

Two small things that still stand:

1. **`trace` in the API lists is dead weight.** Blockbook only ever calls `debug_traceBlockByHash`; it never uses the Parity-style `trace_*` namespace anywhere in the EVM driver. `--http.api`/`--ws.api` in `hyperevm_archive.sh` can drop `trace` and keep `debug`.
2. **Re-validate if nanoreth is ever actually the backend.** Both the dev and prod RPC vars currently point at QuickNode, so — since `WantsBackendService` is derived from the effective RPC URL — no local backend service is wanted and the nanoreth package isn't in the serving path today. Worth confirming whether `hyperevm_archive.sh` / the `nb-20260613` pin are intended for a later self-hosted deploy or are vestigial. If nanoreth does become the backend, tracing needs re-checking there specifically: it ingests pre-executed blocks + receipts from S3 rather than re-executing them, and `callTracer` requires re-execution against parent state.

For the record on the failure mode, had it not been supported: `getInternalDataForBlock` errors are tolerated (ethrpc.go:1657, `// error fetching internal data does not stop the block processing`) — sync would continue, storing `InternalDataError` per block and silently omitting internal txs, plus one failed trace RPC per block at a 1 s block time. Moot here, but that's why it was worth checking rather than assuming.

pragmaxim (approved):
LGTM ✅

This supersedes my earlier review. I raised three blocking-ish concerns there and **tested all three against the QuickNode `hype-mainnet` endpoint that `BB_{DEV,PROD}_RPC_URL_*_HYPEREVM_ARCHIVE` point at — every one was wrong.** The inline comments are amended individually; summary:

| my concern | verdict |
| --- | --- |
| `tests.json` entry needs `"disabled": true` — CI will dial localhost | **Wrong.** `lookupEnvWithArchiveFallback` overrides the RPC URL from the repo vars, and both already exist. Endpoint verified: `eth_chainId` = `0x3e7` (999), WS `101 Switching Protocols`, archive state at block `0x1`, `debug` namespace live. |
| fixture is a placeholder skeleton | **Wrong.** Matches the endpoint exactly at height 42176091 — hash `0x14e51adc…5217`, time `1785740657`, 7 txs, size 4251 — and tracing that block returns the fixture's first txid. |
| `processInternalTransactions: true` unvalidated | **Wrong.** `debug_traceBlockByHash` + `{"tracer":"callTracer"}` returns full nested `calls` trees. Correct as configured. |
| genesis-only `GetBlockHash` override may be insufficient | **Wrong, and the fix is well-scoped.** Compared `HeaderByNumber(n).Hash()` against the node-reported hash at 11 heights (0, 1, 2, 3, 1e3, 1e5, 1e6, 1e7, the fixture block, tip−100, tip−5): **1 mismatch, exactly height 0**, with precisely the two hashes your comment names. Post-genesis headers RLP-recompute consistently, so deferring N≥1 to the cheaper base implementation is right. Disregard my suggested alternative — it was strictly worse. |

I also walked back a claim that Infura's gas API is "unlikely to serve chain 999": that was unverified speculation. It can't be probed without a valid key (network support resolves strictly after auth, so `999` and a nonexistent `999999` both return an identical 401), Infura's own Gas API supported-network table is empty in their docs, and Infura does serve HyperEVM as a first-class network — including `wss://hyperevm-mainnet.infura.io/ws/v3/<key>`.

Nice work on the genesis hash in particular — that's a subtle failure mode that would have wedged sync at height 0 with no obvious cause, and the comment explaining it is the kind of thing that saves the next person hours.

Non-blocking follow-ups, all optional and none worth holding the merge for:

1. **Drop `trace` from `--http.api`/`--ws.api`** in `hyperevm_archive.sh` — the EVM driver only ever calls `debug_traceBlockByHash`, never the Parity-style `trace_*` namespace.
2. **PR description touch-ups:** the summary says "native `eth_feeHistory` EIP-1559 fees (no external provider)" while the config selects Infura, and says HyperEVM's public RPC has no WebSocket, which isn't accurate (Infura publishes one, and the endpoint we actually use serves WSS). Archive depth is still a fair reason to prefer a dedicated backend.
3. **`INFURA_API_KEY` becomes a hard startup dependency** via `initAlternativeFeeProvider` (eth/ethrpc.go:891) — worth confirming it's provisioned wherever this deploys. If Infura turns out not to serve 999, `eth_eip1559_fee_source_total{source=...}` will show only `onchain_fallback`, and `eip1559Fees: true` alone already gives the native fees the description advertises.
4. **Is the nanoreth package in the serving path?** Both dev and prod RPC vars point at QuickNode, and `WantsBackendService` derives from the effective RPC URL — so no local backend is wanted today. Worth noting whether `hyperevm_archive.sh` and the `nb-20260613` pin are for a later self-hosted deploy or vestigial. If nanoreth does become the backend, re-run the genesis-hash comparison and the `debug_traceBlockByHash` check against it, since it ingests pre-executed blocks rather than re-executing them.
5. `mempoolTxTimeoutHours: 12` is inert alongside `disableMempoolSync: true`.

Issue #1573, opened 2026-06-18
EVM alt-mempool: cancelled / replacement txs via a private relay (Blink) linger until timeout


pragmaxim:
Follow-up to #1572 (which fixed the over-aggressive `provider_missing` eviction in the alternative-provider EVM send-tx mempool reconciliation introduced in #1500/#1562). This issue tracks a remaining behavior gap around **cancelled / replacement** transactions submitted through a private relay.

### Glossary

- **RBF — Replace-By-Fee.** Re-submitting a transaction that reuses the **same account nonce** as an already-pending one but with a higher fee. Since Ethereum mines at most one tx per `(account, nonce)`, the replacement supersedes the original. Two flavours:
  - **bump-fee** — same recipient/amount, higher fee, to get the original mined sooner.
  - **cancel** — a zero-value **self-send** (`from == to == account`) at the same nonce, used purely to consume the nonce so the original can never be mined. This is the standard way to "cancel" a pending EVM transaction.
- **Private relay (Blink / Blinklabs).** Keeps the tx in a **private mempool** and hands it to block builders directly instead of the public p2p mempool. It exposes no mempool listing, and `eth_getTransactionByHash` can return `null` for a tx it has accepted but not surfaced. Blockbook tracks these in `AlternativeSendTxProvider`.
- **`confirmedNonce`.** Mined-only account nonce — `eth_getTransactionCount(addr, "latest")`; exposed on the address API since #1562. A pending tx whose nonce is `< confirmedNonce` is **superseded** and can never be mined.

### Background — Blink's cancel behaviour

Blink confirmed (June 2026):
1. **No cancel-by-txid endpoint** (and they won't add one).
2. Cancel is done with a **same-nonce self-send** (standard EVM RBF-cancel).
3. Two behaviours for that self-send:
   - **default** — forwarded to a builder, mined, **spends gas** → nonce consumed.
   - **"drop" mode** (opt-in) — relay **drops both** the original and the cancel self-send, submits neither, **no gas, nonce NOT consumed on-chain**.

### How it maps onto `reconcileMempoolTxs` (`bchain/coins/eth/alternativesendtx.go`)

- **Default-mode cancel (self-send mined):** consumes nonce *N* → `confirmedNonce` → *N+1* → the original (nonce *N*) hits `txNonce < confirmedNonce` → evicted promptly via **`nonce_superseded`**. #1572's "run the nonce check even when the provider reports the tx missing" change is exactly what makes this prompt. ✅
[...]

PR #1638, opened 2026-07-21
eth: fix alt-provider quota burn (#1629), lingering cancelled/replaced private txs (#1573) and trezor-suite#30846

### LINKED ISSUE #1629: eth: every gas estimate is sent to the alternative provider, burning its rate-limit quota
## Problem

`EthereumTypeEstimateGas` (`bchain/coins/eth/ethrpc.go:1961`, provider block at ~1985) routes **every** `eth_estimateGas` request to the alternative send-tx provider's `urls[0]` whenever `*_ALTERNATIVE_SENDTX_URLS` is configured — the guard is a bare nil-check, with a fresh RPC client dialed per call and a **silent** fallback to the primary backend on error (the provider error is discarded, so the problem is invisible in logs).

This is the same pathology as #1627, one function further down the file: the routing is unconditional, but almost no request needs the provider. The path is hit from the websocket `estimateFee` handler (`server/websocket.go:1124`) — i.e. every time a user sits on a wallet's send screen, with estimates re-firing as the amount is edited.

## Example

- An ETH instance has `ETH_ALTERNATIVE_SENDTX_URLS` pointed at a private relay whose API key allows ~10 req/s.
- Alice types an amount into the send form. The wallet requests a gas estimate; Blockbook sends `eth_estimateGas` **to the relay**, although Alice never submitted a private transaction — the primary backend could have answered, and the relay call spends quota for nothing.
- With ~50 users composing transactions concurrently, the relay quota is exhausted → 429s → every estimate pays a wasted round-trip (dial + call + rejection) before the primary answers, making the send form slower for everyone.
- Bob, who *actually* just broadcast a private transaction through the relay, is the one harmed: the quota his nonce lookup and next broadcast need was burned by everyone else's gas estimates.

## Why not simply "never use the provider for estimates"

There is one legitimate case: a sender with a private transaction still pending. If Bob composes a second transaction that builds on the pending state of his first (private) one, only the relay sees that pending transaction and can estimate correctly — the primary backend cannot.

## Proposed fix

Reuse the gating introduced by #1628 for nonces: route `eth_estimateGas` through the alternative provider only for recent private senders (`useForNonces` / the `recentSenders` map), directed at the URL that accepted their send (`nonceURL`); all other requests go straight to the primary backend. Also stop discarding the provider error silently — log it at warning level like the nonce fallback does.

Related: #1627, #1628 (deliberately scoped the estimateGas path out).

pragmaxim:
Fixes #1629 and #1573 in the EVM alternative send-tx provider, plus hardening from an xhigh review of the surrounding subsystem. Also the backend counterpart to **trezor/trezor-suite#30846** — a wallet told "send failed" while Blockbook has already broadcast risks a double payment when the user retries.

Follow-up https://gist.github.com/pragmaxim/f71b03a4ed4c0030b6cba7fb0a302e96

## #1629 — gas estimates burned the relay's rate quota
`EthereumTypeEstimateGas` routed *every* `eth_estimateGas` to the alternative provider (always `urls[0]`, errors swallowed). The WS `estimateFee` handler fires on every send-form keystroke, so concurrent users drained a private relay's quota even without using private txs. Now routed only for recent private senders (`useForNonces` + `nonceURL`), matching the nonce path; others hit the primary backend. Provider errors log and fall back. New `eth_alternative_estimate_gas_requests_total` metric + Grafana panel.

## #1573 — cancelled / replaced private txs lingered until timeout
A Blink drop-mode cancel or an accepted-but-unsurfaced RBF replacement left the superseded tx "Unconfirmed" for up to the 5-min cache timeout, because eviction waited on a fetch-back that never came. Now a same-`(from, nonce)` predecessor is retired the moment the relay ACKs its replacement (decoded from the raw hex), independent of fetch-back — deliberately distinct from a non-authoritative empty `getTransactionByHash` probe.

## Counterpart to trezor-suite#30846 — a relay-accepted send must never look failed
The suite PR gives a transaction push its own 60 s deadline and stops showing "send failed" when the backend is merely slow. Both only hold if Blockbook answers within the wallet's budget **and** surfaces every send the relay accepted, whatever the relay does afterwards. The issues fixed here:

- **an accepted send is cached from its own signed bytes** — an accepted send used to be cached only if the post-send `eth_getTransactionByHash` round-trip surfaced it; a relay that stopped surfacing (or never surfaced) what it accepted left the send exposed *nowhere* — not served as pending, not in the address index, not raising the pending-nonce floor — so a wallet that retried re-signed at the next nonce and both transactions mined. The signed bytes carry everything a pending transaction needs, so the send path now decodes once and caches + indexes the transaction before the relay is ever asked; the fetch-back only refreshes that entry with the relay's own view.
- **the send path stays inside the wallet's deadline** — the broadcast to every configured relay used to run sequentially, so one unresponsive relay held the wallet's answer for its full `rpc_timeout` before the next URL was even tried, and the post-send fetch-back sat in the wait too. Both are gone: the broadcast runs concurrently (worst case a single `rpc_timeout`, not N), and the fetch-back runs in the background, so the wallet's own deadline is no longer consumed by relay handling.
- **the background fetch-back updates, never re-creates** — now that the fetch-back runs off the send path, its answer can straddle a block; re-inserting the pending body over a transaction that mined and was cleared by block sync would flip it back to Unconfirmed and re-index it as pending. It only updates a still-cached entry (or evicts on a mined answer), capped at 16 in flight so a burst of accepted sends cannot grow goroutines, sockets or relay-quota usage without bound.

## Review-driven hardening (same subsystem)
- **estimate-gas fallback**: build the primary-RPC deadline *after* the alt round-trip so a slow relay can't hand the fallback an already-expired context; also fall back (not error) on a malformed provider result.
- **rejected send**: skip the mempool-cache path when no relay accepted, so a rejected send no longer fans out `eth_getTransactionByHash(0x0…0)` and logs a spurious error.
- **metrics gating**: lifecycle counters/residence recorded only by the goroutine that actually removed the entry — no more double-counting across concurrent reconcile / read-path / RBF evictions.
- **generation ordering**: an older submission's slow fetch-back can no longer evict or shadow a newer same-nonce replacement.
- **read-path eviction**: routes through `removeMempoolTx`, so an expired private tx is cleared from the wrapped mempool even when the caller's primary lookup errors.

## Docs
`docs/evm-send.md` — Mermaid flowchart of the broadcast + pending-tx cache lifecycle; fee estimation cross-linked to `fees.md`; what a relay-accepted send guarantees (see above) stated precisely.

## Testing
`go build ./...`, `go vet`, and the eth + common suites pass; each fix ships a regression test.


cranycrane (commented):
Reviewed at `53ee56fd`. The design is sound and the three fixes are real — caching from the signed bytes is the right call, and the nonce-pin trade-off it accepts is correctly documented and instrumented. `go build`/`go vet`/`gofmt` clean, eth + common suites pass, `-race -count=30` on the concurrency tests clean, merges cleanly on current `master`.

Worth calling out as a bonus: this removes a real pre-existing panic. In geth v1.17.0 `newModernSigner` panics on `chainID == nil || chainID.Sign() <= 0`, and `LatestSignerForChainID` routes every non-nil chain id into it — including the zero an unprotected pre-EIP-155 legacy tx decodes to. `master`'s `alternativeTxSender` therefore panicked on the send path for such a raw tx. The new `HomesteadSigner` guard is correct, and it cannot silently recover a *wrong* sender (signature validation rejects `v ∈ {35,36}` rather than returning a garbage address).

Below are the major findings only; medium/minor ones (stale comments, doc inaccuracies, test gaps, bucket change, comment density) are separate.

---

### 1. The cached tx body is shared by pointer and mutated with no lock

`alternativesendtx.go:778` returns the live `storedTx.tx`. `ethrpc.go:1859` then calls `EthTxToTx(..., fixEIP55=true)`, which **writes** `tx.From`/`tx.To` in place (`ethparser.go:137,143`) holding no lock — while `pendingNonceFloor:541`, `txSenderAndNonce` (via `insertMempoolTx:677`, `evictReplacedByNonce:720`) and `refreshCachedTransaction:347` read those same fields under `mempoolTxsMux`. Reproducible under `-race`.

Pre-existing on `master`, but this PR is what makes it worth fixing here: it creates an entry for *every* accepted send instead of only surfaced ones, adds three new readers (two on a background goroutine), and the send path itself drives the writer via `cacheMempoolTransaction:637` → `AddTransactionToMempool` → `createTxEntry` → `GetTransaction`.

- On EVM the damage is bounded — both values are 42 chars, so a torn read still yields a valid string, and `HexToAddress` is case-insensitive. But the cached body is permanently rewritten to EIP-55, which silently falsifies `decodeAlternativeSendTx`'s own documented "lower-case hex addresses" contract after the first read.
- On Tron it is not bounded: `tronparser.go:38` sets `FormatAddressFunc = ToTronAddressFromAddress`, rewriting `From` to a 34-char base58 `T…`, so `txSenderAndNonce` then decodes the wrong address — no RBF retirement, and `pendingNonceFloor` stops covering the tx, i.e. exactly the nonce-reuse exposure this PR exists to close. Latent today (no in-repo config sets `TRON_ALTERNATIVE_SENDTX_URLS`) but `TronRPC` embeds `*eth.EthereumRPC`, so it is env-reachable.

**Fix:** one line in `GetTransaction` — `cp := *storedTx.tx; return &cp, true`. A shallow copy suffices for a pending tx.

### 2. `evictReplacedByNonce` retires at most one predecessor

The `break` at `:733` (with a single `rbfTxid`/`rbfTime`) means one pass evicts one victim. There are two eviction passes per send (`:220` pre-cache, `:632` post-insert), and `insertMempoolTx` only skips on a *strictly higher* gen already visible. With three concurrent sends for one `(from, nonce)` whose `register → slotSupersededBy → insert` windows interleave, all three insert, the newest pass evicts one, and the mid pass finds the newest protected — leaving **two entries cached and address-indexed for one nonce slot**, the state `insertMempoolTx`'s own comment says must never exist. It does not self-heal: a later gen-4 send again evicts only one. Reproduced 174/200 trials by driving the real call sequence.

Impact: two `Unconfirmed` txs for one nonce served to the wallet until the winner mines or the 5-min timeout; `rbf_replaced` under-counts. Needs a fee-bump storm to hit, but the send-generation machinery exists precisely to make this impossible.

**Fix:** collect all victims into a slice, drop the `break`, meter each successful `removeMempoolTx`. That is sufficient — the max-gen send then clears every lower-gen entry, and no lower-gen send can insert afterwards. Also `:700-702` says "removes **any** cached transaction that shares…", which is currently false.

### 3. A dropped fetch-back in the decode-failure branch loses the accepted send entirely

`:240` sends the `decErr != nil` fetch-back through `inBackground`, and `startBackground:282` **drops** the work when 16 are already in flight, with only a `glog.Warningf`. In that branch nothing was cached and `registerSuccessfulSend` never ran — so the accepted send is served nowhere, not address-indexed, and raises no pending-nonce floor. That is precisely the failure this PR exists to prevent, and it is unmetered (`observeSendNotSurfaced` does not fire on a drop, so the dashboard cannot see it).

`inBackground`'s justification at `:250-254` — "reconcileMempoolTxs revisits every cached entry within a minute anyway" — holds for the `:236` call site and is false for this one.

**Fix:** run the decode-failure fetch-back synchronously (it is the only thing that can expose the tx), or at minimum `observeSendNotSurfaced("dropped")` when `startBackground` refuses, and split the comment per call site.

### 4. The fallback tail still exceeds Suite's new 60 s push deadline

`ethrpc.go:2248-2267`: when no relay accepts and `_ALTERNATIVE_SENDTX_ONLY` is not `TRUE`, the handler falls through to the primary `eth_sendRawTransaction` under `b.Timeout` — 25 s (relays, now concurrent) + 25 s ≈ 50 s — and on the `disableMempoolSync` coins (arbitrum, base, bsc, optimism, polygon) `AddTransactionToMempool` then issues a primary `eth_getTransactionByHash` under another 25 s → **≈75 s**, past the 60 s that trezor/trezor-suite#30846 installs.

Pre-existing, and the wallet gets an error on that path anyway — but it is the one remaining route where the user is told "send failed" while the primary may have broadcast. At minimum `docs/evm-send.md:67-72` should stop asserting "the wallet's answer waits for the broadcast and nothing else"; better, skip the fallback `AddTransactionToMempool` round-trip.

### 5. Deployment order: trezor-suite#30846 has to land first

The suite side (`PUSH_TRANSACTION_TIMEOUT = 60_000`, no retry — a Blockbook error and a socket timeout are indistinguishable to `sendFormThunks`) is satisfied by the concurrent broadcast with margin, and returning the signed-bytes hash rather than the relay's echo is an *improvement* for it, since `synchronizeSentTransactionThunk` keys its pending row on that txid.

But the coupling runs both ways: caching every accepted send means `OnNewTx` now pushes pending txs Blockbook previously never surfaced — including contract creations, which serialise `addresses: null`. Pre-#30846 `getEthereumRbfParams` dereferences `vout[0].addresses![0]` and throws inside `addTransaction`, failing the whole account batch. Reachable today for *surfaced* private txs, so not new, but this PR widens the exposure and old Suite builds stay in the field. Worth stating the order in the PR body/changelog.

---

bchain/coins/eth/alternativesendtx.go
```diff
@@ -218,87 +554,215 @@ func (p *AlternativeSendTxProvider) releaseRecentSender(sender ethcommon.Address
...
+
+// cacheMempoolTransaction caches a pending transaction body under txid and indexes it in the wrapped
+// Blockbook mempool, ordering itself against concurrent sends for the same (from, nonce) slot through
+// the send generation gen. Reached from the send path, and from handleMempoolTransaction when the
+// signed bytes did not decode.
+func (p *AlternativeSendTxProvider) cacheMempoolTransaction(txid string, tx *bchain.RpcTransaction, gen uint64) {
+	from, nonce, decoded := txSenderAndNonce(tx)
+
+	// checked before the cache scan below: the scan only sees slots that produced an entry
+	if decoded && p.slotSupersededBy(from, nonce, gen) {
```
cranycrane, line 873: **Correctness: a decode-failure *replacement* is silently dropped, restoring the #1573 symptom on the one path built to prevent it.**

On the decode-failure path `gen` is always 0 — `registerSuccessfulSend` runs only in the `decErr == nil` branch (L229) — so the expose fetch-back at L259 calls `handleMempoolTransaction(txid, 0)` -> `cacheMempoolTransaction(txid, relayBody, 0)`. Here `from`/`nonce` decode from the *relay's* body, so `decoded == true`, and `slotSupersededBy(from, nonce, 0)` returns `s.gen > 0` — true for **any** earlier accepted send on that slot.

Failure scenario:
1. Send A at nonce 5 decodes -> gen 1, cached, `acceptedSlots[{from,5}] = 1`.
2. Send B at nonce 5 whose raw hex `UnmarshalBinary`/`types.Sender` cannot handle (a tx type newer than the pinned go-ethereum, or any sender-recovery failure). The relay accepts B.
3. `slotSupersededBy(from, 5, 0)` -> `1 > 0` -> `true` -> early return.

B is cached nowhere, indexed nowhere, and never raises the pending-nonce floor, while A stays "Unconfirmed" for the full 3h window. No `send_not_surfaced{reason="dropped"}` fires either, because `inBackground` *did* take the work — so the one alerting signal for this class stays silent.

`insertMempoolTx` (L703) carries the same asymmetry (`st.gen > gen` with gen 0), so bypassing `slotSupersededBy` alone would not fix it. The gen-0 "unknown order" sentinel needs distinguishing from "oldest" — e.g. allocate the generation before attempting the decode, or carry an explicit `hasGen bool`.

Reproduced independently by two reviewers against 653b2805.

bchain/coins/eth/alternativesendtx.go
```diff
@@ -343,8 +817,7 @@ func (p *AlternativeSendTxProvider) watchMempoolTxs() {
 	}
 }
 
-// shutdown stops the background mempool reconciliation goroutine. Safe to call on a
-// nil receiver and more than once.
+// shutdown stops the background mempool reconciliation goroutine. Safe on a nil receiver and repeated.
 func (p *AlternativeSendTxProvider) shutdown() {
```
cranycrane, line 1056: **`shutdown()` never drains the background fetch-backs.**

It only closes `p.stop`; nothing waits on `backgroundCount`/`exposeCount` — even though `waitForRefreshes` (L356) already implements exactly that wait and is currently test-only.

Two consequences:

- `startBackground`'s `stopped()` check (L302) runs *before* `backgroundMux` is taken, so a send racing shutdown still spawns a goroutine.
- `EthereumRPC.Shutdown` (ethrpc.go:995) returns with fetch-backs in flight. `waitForSignalAndShutdown` then proceeds past `public.Shutdown`, and a surviving fetch-back can still reach `cacheMempoolTransaction` -> `p.mempool.AddTransactionToMempool` -> `createTxEntry` -> `OnNewTx` -> a websocket push *after* the public server is down, and on into the deferred RocksDB close.

It is also a guaranteed lost send: the goroutine's work is abandoned mid-flight, and L259 suppresses the `dropped` metric via `!p.stopped()`, so it is not counted.

Suggest calling `waitForRefreshes` from `shutdown()` under a bounded timeout and moving the `stopped()` check inside the mutex. Note `rpc.Client.Close` is a no-op over HTTP, so an in-flight probe can outlive shutdown by a full `rpc_timeout` regardless — the drain needs its own deadline rather than relying on the client closing.

bchain/coins/eth/alternativesendtx.go
```diff
@@ -218,87 +554,215 @@ func (p *AlternativeSendTxProvider) releaseRecentSender(sender ethcommon.Address
...
+// cachedNoncesFor returns the account nonces of the private transactions the cache holds for addr.
+// Decoding through txSenderAndNonce keeps a body with no sender out of the answer instead of folding it
+// into the zero address's floor.
+func (p *AlternativeSendTxProvider) cachedNoncesFor(addr ethcommon.Address) map[uint64]struct{} {
+	nonces := make(map[uint64]struct{})
 	p.mempoolTxsMux.Lock()
 	defer p.mempoolTxsMux.Unlock()
-	var floor uint64
-	var found bool
 	for _, storedTx := range p.mempoolTxs {
```
cranycrane, line 734: **Performance: this full-cache scan is now amplified roughly 36x on the hottest EVM endpoint.**

The O(cache) scan under `mempoolTxsMux` is pre-existing (the base `pendingNonceFloor` did the same), so this is not a new defect — but three changes here compound it:

- retention default 5 min -> 3 h, so steady-state cache depth grows ~36x;
- the send path now caches unconditionally from the signed bytes, where before an entry existed only if the fetch-back surfaced the tx;
- `cachedNoncesFor` heap-allocates a `map[uint64]struct{}` per call, where the base accumulated a single max into a scalar.

`EthereumTypeGetNonces` calls `raiseToPendingFloor` whenever a provider is configured — not only for gated senders (ethrpc.go:2409) — and `api/worker.go:1273` runs that on **every** address-info request for any address with indexed data: Suite account polling and websocket `getAccountInfo`. Each call holds the mutex while running `hexutil.DecodeUint64` + `HexToAddress` per entry.

The same mutex takes three more full scans per accepted send (`evictReplacedByNonce` twice, `insertMempoolTx`, plus `removeTransaction`'s sender scan). Nothing bounds cache depth except relay acceptance rate x 3 h.

The comment at ethrpc.go:2407 — "a local scan of a usually-empty map, so it costs nothing on the hot path" — was fair at 5 min but is hard to defend at 3 h. Given the 20k-websocket target in AGENTS.md, consider indexing `mempoolTxs` by sender.

bchain/coins/eth/ethrpc.go
```diff
@@ -1977,19 +2023,49 @@ func (b *EthereumRPC) EthereumTypeEstimateGas(params map[string]interface{}) (ui
...
+			}
 		}
+		// Logged rather than swallowed: a silent provider error hides quota exhaustion (#1629).
+		b.observeAlternativeEstimateGasRequest("error")
+		glog.Warningf("Alternative provider failed for eth_estimateGas: %v, falling back to primary RPC", err)
 	}
+	// Deadline built here rather than at function entry: the provider round-trip above runs on its own
+	// context and a slow or rate-limited relay can burn most of b.Timeout, so a context created at entry
+	// could already be expired by the time the fallback reaches the healthy primary (#1629).
+	ctx, cancel := context.WithTimeout(context.Background(), b.Timeout)
```
cranycrane: **Building the deadline here is the right fix, but it removes the only overall bound on the call.**

`p.rpcTimeout` and `b.Timeout` are both `ChainConfig.RPCTimeout` (25 s on ethereum/base/arbitrum/bsc). Pre-PR the entry-created context capped total wall time at ~25 s; now a hanging relay burns 25 s and the primary then gets a *fresh* 25 s, so the worst case doubles to 50 s.

That matters specifically on this endpoint. With a blackholing relay — the #1629 scenario — a recently-sent-privately address's `estimateFee` polls each hold a `pendingRequestsLimit` slot on the websocket connection for 50 s instead of 25 s, halving send-form throughput exactly when the relay is already degraded.

Suggest capping the fallback at `b.Timeout - elapsed` with a floor, or giving the provider leg a shorter context for this call.

Worth a test as well: the two existing fallback tests use instant-error providers, so they pass with the context built at *either* position. Nothing currently pins the behaviour this comment describes — a slow provider plus a short `b.Timeout` would.

docs/evm-send.md
```diff
@@ -0,0 +1,180 @@
+# EVM transaction submission
+
+How Blockbook broadcasts a **Trezor Suite** EVM transaction through the private send-tx relay and
+tracks it in its own pending-transaction cache until the chain confirms or supersedes it. Fee
+estimation is covered in [fees.md](/docs/fees.md).
+
+The private-relay path — the alternative send-tx provider and its pending-tx cache — is active only
+when a coin is configured with `*_ALTERNATIVE_SENDTX_URLS`, `*_ALTERNATIVE_SENDTX_ONLY` and
+`*_ALTERNATIVE_FETCH_MEMPOOL_TX`. Without them Blockbook sends through the normal backend RPC and the
```
cranycrane: **Factually wrong: the relay broadcast does not require all three variables.**

With only `*_ALTERNATIVE_SENDTX_URLS` set the relay path **is** active — `NewAlternativeSendTxProvider` returns non-nil on URLs alone (alternativesendtx.go:131) and `EthereumRPC.SendRawTransaction` routes through it (ethrpc.go:2288). Only the *pending-tx cache* needs all three, via the `p.onlyAlternative && p.fetchMempoolTx` gate at alternativesendtx.go:240.

As written this tells an operator that a URLs-only config does not use the relay, which is the opposite of what happens — and that is the config most likely to be deployed first. Suggest splitting the sentence: broadcast requires `_URLS`; the pending-tx cache additionally requires `_ONLY` and `_FETCH_MEMPOOL_TX`.

docs/evm-send.md
```diff
@@ -0,0 +1,180 @@
...
+  from an *empty* `eth_getTransactionByHash` probe, which
+  is **not** authoritative: a private relay stops surfacing a still-mineable tx while it stays
+  broadcast, so `provider_missing` is kept until the timeout. `mined` and `nonce_superseded` are
+  reconcile's only deterministic early evictions.
+- **Send generations order concurrent same-nonce sends.** Each accepted send gets a monotonic
+  generation; an older submission's slow fetch-back neither caches itself over, nor evicts, a newer
+  replacement that already holds the nonce slot.
+- **Every exit funnels through `removeMempoolTx`** (see the diagram for what it clears), which records
+  the lifecycle metric exactly once — gated on the actual removal, so concurrent reconcile /
+  read-path / RBF evictions of the same entry don't double-count.
```
cranycrane: **Both halves of this are inaccurate.**

*"records the lifecycle metric exactly once"* — `removeMempoolTx` calls `removeTransaction(txid, "")`, and the empty action means it records **nothing** (alternativesendtx.go:1116-1123, 1163). The metering lives in the callers (`evictMempoolTx`, `evictReplacedByNonce`, the `GetTransaction` timeout path), gated on its bool return. The gating claim is correct; the attribution is not.

*"Every exit funnels through `removeMempoolTx`"* — `sync_removed` does not: ethrpc.go:1782 calls `RemoveTransaction` -> `removeTransaction(txid, "sync_removed")` directly. That is the exit L149-150 says "should dominate", so the claimed universal funnel misses the most common one, and the diagram at L40 omits the edge entirely.

The same misattribution appears at docs/evm-send-mempools.md:63 and :83 — L83 doubly so, since nothing is metered at the delegate boundary.

docs/evm-send.md
```diff
@@ -0,0 +1,180 @@
...
+- **Removals are not pushed to the wallet.** Blockbook pushes only *added* txs; a wallet learns a
+  pending tx is gone on its next account re-fetch (the initiating device also removes it
+  optimistically). The cache timeout, at the pending window's length, is only the backstop, so `mined`
+  and `nonce_superseded` retire an entry in practice.
```
cranycrane: Contradicts L149-150 in this same file, which says `sync_removed` "should dominate, `mined` should be rare". Both cannot hold.

configs/grafana/panels.yaml:178 and configs/metrics.yaml:230 side with L149-150, so this line is the one to correct.

cranycrane (commented):
Review of 653b2805. A full pass over the subsystem produced a longer list; filed here are the five findings I consider worth acting on. Each was traced against the PR head and diffed against the base at 180c2cd2.

Two are correctness (the gen-0 replacement drop and the missing shutdown drain), one is performance on the hottest EVM endpoint, one is a latency-bound regression, and four inline notes cover factual errors in `docs/evm-send.md`.

The core fixes are sound — I verified that the key regression tests genuinely fail on the pre-PR code, `decodeAlternativeSendTx` is field-complete against `bchain.RpcTransaction`, `GasPrice = GasFeeCap` matches geth's own pending-tx encoding, and there is no lock inversion across the three mutexes. `go build` and `go vet` are clean.

Two things not filed as inline comments:

- The default pending window moves 5 min -> 3 h and `mempoolRetentionInverted` makes an explicit `mempoolTxTimeout` at or below the cache retention fatal at startup. That is the most operationally consequential change in the PR and is not mentioned in the description. It needs a changelog entry and an upgrade note — `changelog.md:63` still advertises #1500 ("Decrease mempool timeout for MEV-protected coins") in the same Unreleased section this reverses.
- 49 of the PR's tests are `httptest`-based and could not be executed in my environment (loopback bind denied). Of the 108 that could run, all pass under `-race -count=5` with no races. The httptest ones still need a run on a machine that permits listeners.

docs/evm-send.md
```diff
@@ -0,0 +1,180 @@
...
+`blockbook_eth_alternative_pending_floor_raised_total{source="provider"}`: it fires today *because*
+the relay drops a still-cached transaction from its pending count, so it collapsing toward zero is
+what says the relay is answering over the full window.
+
+## Observability
+
+Prometheus counters for the cache lifecycle:
+
+- `blockbook_eth_alternative_mempool_reconciliation_events_total{action}` — cache exits by reason
+  (`mined`, `nonce_superseded`, `provider_missing`, `timeout`, `rbf_replaced`, `sync_removed`) plus
```
cranycrane, line 167: The `sync_removed` definition is incomplete. "block sync indexing the tx's block, or the read path finding it mined" omits the third trigger: ethrpc.go:1889 also fires it when the read path finds the tx **unknown** — a null response whose pruned-index recovery also failed.

The code's own comment at alternativesendtx.go:1127-1128 gets this right; this doc is the stale copy.

bchain/coins/eth/ethrpc.go
```diff
@@ -53,10 +53,17 @@ const (
...
-	// Alternative/private relays expire pending txs quickly, so local pending state
-	// must not inherit the legacy hour-scale public mempool timeout.
-	defaultMempoolTxTimeoutWithAlternativeProvider = 10 * time.Minute
-	defaultAlternativeMempoolTxTimeout             = 5 * time.Minute
+	// defaultAlternativePendingTxWindow is how long a transaction broadcast through a private relay is
+	// treated as still pending. A relay dropping it from eth_getTransactionByHash and from its pending
+	// nonce count is not the same thing as it being gone: it stays broadcast and builders can still
+	// include it, which Blinklabs puts at around three hours. Sizing pending state below the relay's
+	// window made Blockbook stop showing a live transaction and hand its nonce back out (#1573, #1675).
+	defaultAlternativePendingTxWindow = 3 * time.Hour
```
cranycrane, line 59: **Upgrade hazard: this changes deployed behaviour by ~36x and is not mentioned in the PR description.**

| | base (180c2cd2) | this PR |
|---|---|---|
| alt cache retention | 5 min (`defaultAlternativeMempoolTxTimeout`) | **3 h** |
| wrapped mempool, provider enabled | 10 min (`defaultMempoolTxTimeoutWithAlternativeProvider`) | **3 h 30 min** |

The rationale — a relay keeps a tx mineable for ~3 h after it stops surfacing it — is sound and well argued in the comment above. Two consequences are worth stating where an operator will actually read them:

- A private tx that is accepted and then dropped (the Blink drop-mode cancel) now holds its nonce slot and reads "Unconfirmed" for **3 hours** instead of 5 minutes. Reconcile deliberately does not evict on `!known`, and `transactionSupersededByNonce` only fires on `txNonce < confirmed`, so nothing retires it early. `raiseToPendingFloor` keeps handing out the nonce above it for that whole window, and `stranded` stays false, so no metric fires either.
- Steady-state cache depth grows ~36x, which is what turns the `cachedNoncesFor` scan on the address-info path into a real cost (see my comment on alternativesendtx.go:564).

`changelog.md:63` still advertises #1500 *"Decrease mempool timeout for MEV-protected coins"* in the same Unreleased section that this reverses. Both cannot ship in one release without confusing whoever reads the notes — please add a changelog entry for this change and reconcile the #1500 line.

bchain/coins/eth/ethrpc.go
```diff
@@ -708,6 +749,13 @@ func (b *EthereumRPC) CreateMempool(chain bchain.BlockChain) (bchain.Mempool, er
 		if err != nil {
 			return nil, err
 		}
+		// Checked here, not in Validate: the effective retention depends on the env-configured provider
+		// existing. Inverted, private transactions silently lose their address index while still being
+		// served as pending (#1573), so it is rejected rather than warned about - and rejected before the
+		// mempool is assigned, so a second call cannot hand out a mempool with no provider wired to it.
+		if b.alternativeSendTxProvider != nil && mempoolRetentionInverted(b.alternativeSendTxProvider.mempoolTxsTimeout, mempoolTxTimeout) {
```
cranycrane: **This makes a previously-valid config fatal at startup, and the failure mode is noisier than it looks.**

`MempoolTxTimeoutDuration` returns an explicit `mempoolTxTimeout` verbatim, and `mempoolRetentionInverted` (L186) is `>=`. With the retention default now 3 h, any relay-enabled deployment carrying an explicit `mempoolTxTimeout` at or below 3 h fails to start. `"10m"` — the *previous* release's documented default for exactly this case — and `"0s"` both trip it.

The error is fatal:

```
ethrpc.go:757      CreateMempool -> errors.Errorf(...)
blockchain.go:193  NewBlockChain propagates
blockbook.go:445   getBlockChainWithRetry(..., seconds=120)
blockbook.go:197   -> exitCodeFatal
```

Two aggravating details:

- **It retries for 120 s first.** `getBlockChainWithRetry` is called with `seconds=120`, so a static config error that can never resolve is retried once per second for two minutes before the process gives up.
- **It is reported as an RPC fault.** Each retry logs `glog.Error("rpc: ", err, " Retrying...")` (blockbook.go:446), so the operator sees 120 lines prefixed `rpc:` for what is a config-validation failure.

The comment above explains why the check is not in `Validate` — the effective retention depends on the env-configured provider, which `Validate` cannot see. That reasoning holds, but `Validate` failures skip the retry loop and this one does not. Worth either short-circuiting the retry for config errors or hoisting the check somewhere that fails fast.

No shipped `configs/coins/*.json` sets `mempoolTxTimeout`, so CI will not catch this — the exposure is entirely in deployed configs. Please call it out in the release notes, and correct docs/config.md, which still says `"0s"` "preserves the legacy zero-retention setting" in the same bullet that now makes it fatal.

cranycrane (commented):
Filing the retention change as anchored comments, since it was only a note in my previous review summary.

This is the most operationally consequential change in the PR: the private-tx pending window moves 5 min -> 3 h, and an explicit `mempoolTxTimeout` at or below the cache retention becomes a startup failure. Neither appears in the PR description. Details inline.

cranycrane (approved):
LGTM after fixing the findings

bchain/coins/eth/ethrpc.go
```diff
@@ -708,6 +778,15 @@ func (b *EthereumRPC) CreateMempool(chain bchain.BlockChain) (bchain.Mempool, er
...
```
cranycrane: This guard keys on `b.alternativeSendTxProvider != nil` alone, but the cache it protects only exists when `fetchMempoolTx` is on:

- `cacheMempoolTransaction` is reachable only under `p.onlyAlternative && p.fetchMempoolTx` (send path), or via `exposeAcceptedSend` behind that same gate
- `GetTransaction` and `removeTransaction` both early-return on `!p.fetchMempoolTx`

So with `*_ALTERNATIVE_SENDTX_URLS` set and `*_ALTERNATIVE_FETCH_MEMPOOL_TX` unset, `mempoolTxs` is never written and never read — there is no address index to lose and nothing to invert. But a deployment in that shape carrying an explicit `mempoolTxTimeout: "10m"` now hard-fails startup, and because it is wrapped in `ErrConfiguration` it deliberately does not retry out of it.

Gating on the flag that makes the inversion possible in the first place:

```go
if b.alternativeSendTxProvider != nil && b.alternativeSendTxProvider.fetchMempoolTx &&
    mempoolRetentionInverted(b.alternativeSendTxProvider.mempoolTxsTimeout, mempoolTxTimeout) {
```

(Separately, and not a code issue: `configs/coins/bsc.json` sets neither `mempoolTxTimeout` nor `disableMempoolSync`, so on a relay-enabled BSC the derived default takes the wrapped *public* mempool from 10 min to 3 h 30 m retention. `ethereum.json` and every `*_archive.json` pin `mempoolTxTimeout` explicitly, and `arbitrum.json`/`base.json` disable mempool sync — bsc looks like the one relay network that inherits the new default. Might be worth pinning it there too.)

bchain/coins/eth/ethrpc.go
```diff
@@ -1977,19 +2055,67 @@ func (b *EthereumRPC) EthereumTypeEstimateGas(params map[string]interface{}) (ui
...
 }
 
+// minEstimateFallbackTimeout is what the primary estimate keeps even when the relay leg consumed the
+// whole request budget. Without a floor a slow relay would leave the fallback pre-expired, which is the
+// failure the deadline placement above exists to avoid.
+const minEstimateFallbackTimeout = 5 * time.Second
+
+// remainingEstimateTimeout returns what is left of the request's budget since started, floored at
+// minEstimateFallbackTimeout.
+func (b *EthereumRPC) remainingEstimateTimeout(started time.Time) time.Duration {
```
cranycrane, line 2139: `minEstimateFallbackTimeout` is returned unconditionally once the budget is thin, so the floor can exceed the budget it is meant to live inside. With the default `rpc_timeout: 15`, a blackholing relay burns 15 s on the provider leg and the fallback then gets a full fresh 5 s → ~20 s total, on the endpoint the wallet calls per keystroke.

That is a clear improvement on the 30 s it replaces, and the comment above says exactly that. But it still overshoots `b.Timeout`, and the overshoot grows in relative terms the lower `rpc_timeout` is set — at `rpc_timeout: 2` the request can run ~7 s, 3.5x its own budget.

If `b.Timeout` is meant to be a real ceiling:

```go
func (b *EthereumRPC) remainingEstimateTimeout(started time.Time) time.Duration {
	floor := minEstimateFallbackTimeout
	if b.Timeout < floor {
		floor = b.Timeout
	}
	if remaining := b.Timeout - time.Since(started); remaining > floor {
		return remaining
	}

	return floor
}
```

`TestEthereumTypeEstimateGasBoundsFallbackToRequestBudget` asserts `elapsed <= b.Timeout + minEstimateFallbackTimeout`, so this reads as deliberate — flagging it in case the overshoot wasn't the intended reading of "what it gets is the rest of the request's budget, not a fresh one".

cranycrane (commented):
Three notes from a read of the send / estimate path. Two are inline; this one sits on a line the diff doesn't touch.

### The provider cache lookup isn't hash-normalized (`bchain/coins/eth/ethrpc.go:1897`)

`GetTransaction` normalizes the txid for the RPC leg but hands the caller's raw string to the cache:

```go
hash := ethcommon.HexToHash(txid)                                  // normalized
if b.alternativeSendTxProvider != nil {
    tx, txFound = b.alternativeSendTxProvider.GetTransaction(txid)  // not normalized
}
```

The cache is now keyed on `tx.Hash().Hex()` from `decodeAlternativeSendTx` — always lower-case and `0x`-prefixed. A txid arriving in any other form (upper/mixed case, missing `0x`) misses the cache and falls through to the primary RPC, which by design does not have the private transaction, so it reads as not-found.

This is pre-existing behaviour, but it matters more after this PR: the cache is now the authoritative store for a relay-accepted send rather than a best-effort mirror of what the relay surfaced, so a missed lookup is exactly the "exposed nowhere" outcome the PR sets out to eliminate. `removeTransactionFromMempool` → `RemoveTransaction(txid)` has the same shape.

Passing `hash.Hex()` would close it, or normalizing inside the provider's map accessors so every entry point is covered.

---

None of the three is blocking. One caveat on the depth of this review: I could not execute `go test ./bchain/coins/eth/...` in my environment (loopback binding unavailable, and the provider tests all use `httptest.NewServer`), so the new coverage is reviewed by reading only. `go build`, `go vet` and `go test ./common/...` are clean, and the Grafana renderer accepts the new panels.

bchain/coins/eth/ethrpc.go
pragmaxim: Fixed in ff4c5ed3 exactly as suggested — both the guard **and** the derived default now key on `cacheEnabled` (provider + `fetchMempoolTx`), since the URLS-only over-reach applied to the default too (it silently took the mempool from legacy hours to window+30m). Two tests pin the URLS-only contract: an explicit short `mempoolTxTimeout` is accepted, and the derivation provably never reads the alternative window. Docs (`config.md`, `evm-send-mempools.md`) updated to say "cache enabled" where they said "provider configured", and `bsc.json` is pinned to 12h/3h like its archive sibling — good catch on it being the one relay network inheriting the derived default.

One deliberate stop short: `cacheEnabled` keys on `fetchMempoolTx` alone, not `onlyAlternative && fetchMempoolTx` (the actual cache-write gate). A FETCH-without-ONLY deployment is an inert configuration either way, and there the guard errs on the safe side.

bchain/coins/eth/ethrpc.go
pragmaxim: Fixed in 2ebfd648 with your exact clamp — the overshoot wasn't the intended reading, thanks. `remainingEstimateTimeout` now floors at `min(minEstimateFallbackTimeout, b.Timeout)`, so a coin with `rpc_timeout` under 5s worst-cases at two budgets instead of budget+5s. The bounds test now asserts the clamped two-budget ceiling (it previously allowed the unclamped bound and couldn't have caught a regression), and a unit subtest pins the clamp arithmetic directly at a 2s budget.

pragmaxim:
The review-body point (cache lookup not hash-normalized) is fixed in 71b7a26e. Rather than normalizing at the two call sites, `normalizeTxid` (`ethcommon.HexToHash(txid).Hex()`) is applied inside the provider's boundary methods — `GetTransaction`, `removeTransaction` (which covers `RemoveTransaction` from block sync and the read path), and `cacheMempoolTransaction` (which also covers the relay's echoed txid on the undecodable-hex insert path) — so every present and future entry point is covered, as you suggested. A table test pins upper-case, prefix-less and `0X`-prefixed spellings for both lookup and removal.

All three points addressed (see the inline replies for the other two); full unit suite green.

PR #1639, opened 2026-07-21
eth: wallet-declared privatePending hint for nonce + gas routing + 2 replicas handling 

pragmaxim:
**Backs the design note:** https://gist.github.com/pragmaxim/f71b03a4ed4c0030b6cba7fb0a302e96

### The problem: two replicas, one heuristic that only one of them has

Blockbook can run as load-balanced replicas behind a single DNS name with no request affinity: Trezor Suite's "preferred backend" pinning is per-*URL*, so every reconnect is a fresh load-balancer decision. Today the private-send state a nonce lookup depends on is **per-instance**: `recentSenders` is populated only on the replica that accepted the `eth_sendRawTransaction`. When the wallet's next WebSocket session lands on the *other* replica — or the accepting replica restarts — that instance sees no reason to route to the relay, answers `eth_getTransactionCount` from the primary RPC (which cannot see a private tx), and hands the wallet back the nonce of a transaction it has in flight. Gas estimates have the same blind spot for a pending private predecessor (e.g. an `approve` a following swap's gas depends on).

The wallet knows its in-flight set authoritatively, so this PR lets it **declare** it: an optional `privatePending: {nonces, txids}` field on the read requests themselves. Routing then follows the declaration on *whichever replica* serves the request — the heuristic stays as the fallback for hint-less clients, and both consumers of the `recentSenders`/`useForNonces`/`nonceURL` machinery are covered, which is the prerequisite for retiring that machinery once wallet adoption is confirmed (not done here; this PR is additive).

### Mechanism

- **`getAccountInfo` → top-level `privatePending`** routes the nonce lookup to the relay and folds each declared nonce into the pending-nonce floor as an **occupied slot**: the answer walks the backend's reply across the contiguous run of occupied slots (declared and cached alike), so it never hands out an in-flight nonce and never advertises past a hole (#1675) — a declared nonce above an unfilled slot strands instead of lifting the answer. The list is capped at 64; past the cap the **lowest** entries are kept, since only the slots just above the backend's answer can ever be consumed.
- **`estimateFee` → `specific.privatePending`** is a **routing signal only** — the wallet cannot simulate gas, so Blockbook still calls `eth_estimateGas`, just against the relay's pending-private state; the field is stripped from the forwarded call object.
- Only `nonces` drive behavior; `txids` are accepted for forward compatibility. With no relay configured the hint is a no-op. Absent the field, behavior is exactly today's; an older Blockbook ignores it.

### Interaction with the aligned relay (base branch #1638)

The relay now answers `eth_getTransactionByHash` and the pending `eth_getTransactionCount` over its full ~3 h window, which makes the hint *more* effective: a hint-routed lookup on the replica that never saw the send gets an authoritative answer directly from the relay, with the declared floor as the safety net for a lagging node or a tx the answering node never saw. One asymmetry is documented explicitly: a declaration is **never expired server-side** (unlike a cached tx, which the missing eviction retires) — pruning it is the wallet's side of the contract, and a wallet deriving it from Blockbook's own pending answers self-heals within minutes of a drop.

### Suite adoption

The producer side is **merged in trezor-suite `develop`** (`754da40740`): Suite declares its own pending txs on every account sync and on estimate composes, with the wire format matching this PR (top-level on `getAccountInfo`, under `specific` on `estimateFee`). Note it declares *all* own pending txs, not only private ones, and at keystroke rate while anything is pending — the metric/panel texts and the trust-boundary rationale here are written against that reality: the routed population is bounded by senders with something pending, not by wallets declaring "rarely".

### Trust boundary (accepted)

`privatePending` is an unauthenticated, per-request client hint. It is never written into `recentSenders` or the pending-tx cache, so a hostile client can distort only its **own** request's answer — never shared state. Its one outward effect, routing the read to the relay, adds no capability an unauthenticated caller doesn't already have: `sendTransaction` spends the same relay quota on every call, fanned out to every relay URL, at no cost for a rejected transaction — so relay-quota protection, if ever wanted, belongs in the server's per-IP rate limiting, not in this path. Routing the nonce lookup also rests on a stated relay property: Blink's pending count is documented as max(node pending count, relay's latest accepted nonce + 1) — a public-mempool superset — so a routed answer never falls below the primary's. Full contract in [`docs/evm-send.md`](docs/evm-send.md).

### Shape of the series (12 commits)

1. Wire plumbing: `WsPrivatePending` on `getAccountInfo` → `AddressFilter.PrivatePendingNonces` (no behavior change).
2. Nonce routing short-circuit + declared slots joining the floor walk.
3. Cap keeps the lowest nonces (order-independent answers).
4.–5. `estimateFee` contract + routing short-circuit.
6.–9. Contract/docs hardening from adversarial review: the walk (not `max+1`) stated everywhere a client or operator reads it, best-effort estimate URL selection, the hint-less trade-off, the never-expired-server-side declaration.
10. `openapi.yaml` parity: `WsPrivatePending` schema + parity assertion (the parity typecheck is not run by PR CI — it would have failed at deploy).
11. Operator texts: the request counters/panels now say routing is declared-or-recent-sender and what their volume tracks.
12. Tests: the declared-floor matrix (declared-only, union with cache, nonce 0, holes/stranding, `MaxUint64`, cap), estimate wire-decode through a real envelope, and an end-to-end seam test — the fakechain now honors declared nonces, so silently dropping either variadic forward fails a server-level `getAccountInfo` test.

### Known, out of scope

The committed `blockbook-api.ts` is globally stale vs the generator (indent + a couple of Tron fields) — pre-existing on `master`; these additions are hand-mirrored in the file's committed style. A full regeneration is a separate cleanup.


bchain/coins/eth/ethrpc.go
```diff
@@ -2433,29 +2481,40 @@ func (b *EthereumRPC) EthereumTypeGetBalance(addrDesc bchain.AddressDescriptor)
 // lookup fails, the pending nonce is still returned with confirmedOK=false so the caller
 // can omit it rather than failing the whole request. When confirmedOK is false the returned
 // confirmed value is 0 and must be ignored.
-func (b *EthereumRPC) EthereumTypeGetNonces(addrDesc bchain.AddressDescriptor, withConfirmed bool) (uint64, uint64, bool, error) {
+func (b *EthereumRPC) EthereumTypeGetNonces(addrDesc bchain.AddressDescriptor, withConfirmed bool, privatePendingNonces ...uint64) (uint64, uint64, bool, error) {
```
cranycrane, line 2546: **Routing on the hint rests on an undocumented relay property — and can *lower* the answer**

This gate routes to the relay whenever a declaration is present. The declared floor then compensates for the *declaring wallet's own* txs — but nothing compensates for txs the wallet doesn't know about, and the relay's answer replaces the primary's entirely.

Concrete failure: confirmed nonce 5; another device on the same account has public pending txs at 5 and 6; this client declares its own private tx at 7. The primary would answer 7. If the relay's `eth_getTransactionCount("pending")` reflects only its private window and not the public mempool, it answers 5 → the floor walk finds 5 unoccupied → **answer 5, stranded=true** → the wallet is handed a nonce another pending tx already owns. Without the hint this request would have been answered correctly by the primary.

`docs/evm-send.md` and `docs/config.md` state the relay answers over its whole 3 h window, but neither states whether that count is a *superset* of the public mempool. The design hinges on it: if it is a superset, routing is strictly better and this is fine; if it is private-only, the hint converts a correct primary answer into a collision for any multi-device account. The pre-PR exposure was the same but gated to senders who had just sent privately — the Suite producer widens it to every account with anything pending.

Two asks:

1. **State the relay's pending-count semantics explicitly in `docs/evm-send.md`** (does it include the public mempool, or only its own private window?). Everything below follows from that answer.
2. If it is *not* a public-mempool superset, **consider making the nonce path floor-only (no routing)**. The primary + declared floor already produces the correct answer for the declaring wallet in every test in this PR — `TestEthereumTypeGetNonces_PrivatePendingHint_RaisesPrimaryFallback` demonstrates exactly that — at zero relay cost and with guaranteed public-mempool coverage. Routing buys only *undeclared* private txs (another device sending privately on the same account), which is a real case, so this is a trade-off rather than an obvious win — but it should be an explicit documented choice rather than a side effect of reusing the existing routing gate.

docs/evm-send.md
```diff
@@ -159,6 +159,77 @@ not surface accepted transactions over its window at all, set `alternativeMissin
...
+without the private predecessor. Declaring `privatePending` closes that gap deterministically;
+widening routing for hint-less senders is avoided because it is indistinguishable from the #1629
+hot-path drain.
+
+**Trust boundary (accepted).** `privatePending` is an *unauthenticated client hint* — Blockbook does
+not verify the caller owns the address or that a private tx actually exists. This is safe because the
+declaration is per-request only: it is never written into `recentSenders` or the pending-tx cache, so
+a hostile client can distort only its **own** request's answer and cannot poison another client's
+view or any shared state. Its one outward effect is forcing the read to route to the relay; that is
+bounded by the relay's own rate-limit quota and the per-connection pending-requests limit, and — by
```
cranycrane: **The trust boundary understates the outward effect**

Two gaps in "bounded by the relay's own rate-limit quota and the per-connection pending-requests limit":

- **The per-connection limit is not the ceiling.** The practical bound is per-IP connections x pending requests: `defaultWsMaxConnectionsPerIP = 128` x `defaultWsPendingRequestsLimit = 48` is roughly 6k concurrent hint-routed relay calls **per IP**, before considering multiple source IPs.
- **The relay quota is the same resource `eth_sendRawTransaction` depends on**, and with `ALTERNATIVE_SENDTX_ONLY=true` (`UseOnlyAlternativeProvider`) the send path has no primary fallback. So the hint's outward effect is not purely per-caller: an unauthenticated *read* can degrade *broadcast* for everyone. Reads degrade gracefully; sends do not. That is a different claim from "can distort only its own request's answer", which does hold — I verified nothing writes the declaration into `recentSenders` or the pending-tx cache, and there is no `getAccountInfo` response cache keyed on `AddressFilter`.

Suggested mitigation, cheap and aligned with the existing design: a budget for hint-routed relay reads (token bucket per connection or per IP, or a global ceiling) that **degrades to floor-only** when exhausted. On the nonce path that costs nothing in answer quality (see the other comment); on the estimate path the fallback is the primary RPC, i.e. exactly pre-hint behavior. That turns the doc's bound into a structural property rather than a deployment assumption.

cranycrane (commented):
Two points I'd want settled before merge — both about the routing gate rather than the plumbing, which looks solid (interface fan-out is complete, the floor walk semantics and the cap-keeps-lowest reasoning are right, and the fakechain change genuinely makes the variadic seam observable). Full review notes elsewhere; these are the two blocking-ish ones.

cranycrane (approved):
LGTM

bchain/coins/eth/ethrpc.go
pragmaxim: You were right to demand this be stated rather than assumed — the property turns out to be documented, and the design now says so. Blink's `eth_getTransactionCount` API reference: for the `pending` tag it "returns the greater of: the node's pending transaction count, and Blink's latest recorded nonce for that address + 1" — i.e. a **public-mempool superset**, so a routed answer can never fall below what the primary would have said, and the multi-device collision in your scenario can't occur with a conforming relay (your scenario's relay answer would be ≥ 7, not 5).

Addressed in fbad99b3: ask 1 is done — `docs/evm-send.md` states the superset property as what routing rests on, and the gate's comment in `ethrpc.go` carries it too. Ask 2's conditional therefore doesn't trigger, but the docs now say explicitly what a relay *without* the property would require — a floor-only nonce path, which the fallback branch already computes — so the routing-vs-floor-only choice is a documented decision with its premise named, not a side effect of reusing the gate.

docs/evm-send.md
pragmaxim: Both gaps confirmed (the 128 × 48 arithmetic checks out against `defaultWsMaxConnectionsPerIP`/`defaultWsPendingRequestsLimit`, and the shared-quota-with-broadcast point is the sharper one — reads degrade gracefully, sends under `ALTERNATIVE_SENDTX_ONLY` don't). Implemented your suggestion in c969e449 as a global ceiling: relay reads that exist *only* because of a declaration draw on a token bucket (`ALTERNATIVE_HINT_READS_PER_SECOND`, 50/s sustained, 2× burst, `0` disables; env-configurable since quota sizes are deployment knowledge). Past the budget the hint degrades exactly as you outlined — nonce → primary + declared floor (same answer for the declaring wallet), estimate → primary RPC (pre-hint). The `recentSenders` heuristic is consulted first and never draws on the budget, so real private senders are unaffected even under a flood. `eth_alternative_hint_reads_total{result=routed|degraded}` plus a panel make an under-sized budget — or someone probing the boundary — visible, and the trust-boundary section now states the real arithmetic and the budget as the structural bound.

pragmaxim: Follow-up: after a second look we **removed the budget again** and answered your observation with documentation instead — because the capability it guarded turns out not to be marginal to the hint. `sendTransaction` is itself unauthenticated and spends the same relay quota on every call, **fanned out to every relay URL**, unbudgeted, at no cost for a rejected transaction. A `privatePending` declaration therefore buys an attacker strictly *less* quota spend per request than the endpoint that has always been there; budgeting the side door while the front door is open added machinery (which already needed a hardening round for its own defects, plus a metric to watch for its own misconfiguration) without adding protection — on a series whose point is to retire routing machinery.

Your two factual points stand: the per-connection limit is not a meaningful bound, and the relay quota is shared with broadcasts. The conclusion drawn from them changed: relay-quota protection, if ever wanted, belongs in the server's per-IP rate limiting uniformly across methods (sends included), not inside this feature.

The add→harden→revert churn was squashed out of the branch for clean history — the surviving commit is **bf79510d**, which carries the rewritten trust boundary and the prose trims (the interim budget commits referenced earlier in this thread no longer exist on the branch). If you disagree that the sendTransaction vector makes the budget moot, happy to reinstate it — it's one commit either way.

PR #1640, opened 2026-07-21
ci: BB_STAGING escape hatch for work-in-progress coins

pragmaxim:
## Problem

A new coin's `BB_RUNNER_<coin>` / `BB_*_RPC_URL_*_<coin>` **repository variables are visible on every branch at once**, but its `configs/coins/<coin>.json` lives only on the feature branch that adds it. On `master` and every sibling branch that orphaned variable trips two global fail-closed validations and breaks builds/deploys/tests **for every coin**:

- `.github/scripts/runner.py` `validate_runner_map_configs` → `BB_RUNNER_* entries without matching configs/coins/<coin>.json: robinhood_archive`
- `build/tools/templates.go` `validateRPCEnvVars` (run on every per-coin `LoadConfig`) → `RPC env vars reference unknown coin aliases: robinhood_archive ...`

GitHub has no per-branch variables (only org/repo/environment scopes), so the fix keys off the artifact that *is* already per-branch: the committed config.

## Solution

A new `BB_STAGING` repository variable lists work-in-progress coins (comma/space separated). Because it's a repo variable it is visible on every branch, which is exactly what's needed to tell all branches "tolerate this coin." It means **"tolerate the *absence* of this coin's config"**, not "disable checks":

| Config on this branch? | In `BB_STAGING`? | Result |
|---|---|---|
| Yes (feature branch) | Yes/No | coin fully built / tested / deployed (`BB_STAGING` is a no-op) |
| No (master / others) | Yes | coin skipped, no error |
| No (master / others) | No | still fails closed — typo/stale detection preserved |

The discriminator is config-file presence, not branch name, so the feature branch that carries the config exercises the coin normally while sibling branches skip it.

- **`runner.py`**: drop a `BB_STAGING` coin from the runner map only where its config is absent (also keeps it out of `all_coins` / `build ALL`).
- **`templates.go`**: add `BB_STAGING` coins (with `-`/`_` variants) to the valid-alias set so their orphan RPC vars are tolerated.
- **`Makefile`**: `BB_STAGING` is a standalone control var (like `BB_BUILD_ENV`), forwarded into the build/test container via `BB_RPC_ENV` — name-only `-e BB_STAGING` so a multi-coin list value survives intact. Covers the deb build path (`generate.go` → `LoadConfig`).

## Known limitation

Tolerance is opt-in: a coin mistyped in *both* the coin variable and `BB_STAGING` (no config on any branch) is silently skipped rather than erroring. It's caught where it matters — on the feature branch the config is present, so the coin fails to build and the author notices — and the skip is logged. Documented in `docs/ci_cd.md`.

## Testing

- `python3 -m unittest discover -s .github/scripts -p '*_test.py'` — 69 pass
- `go test ./build/tools/` — pass
- Covers: prune-without-config, keep-with-config, `-`/`_` variant matching, all-pruned diagnostic, and RPC-alias tolerance.

## Review

Reviewed at xhigh effort; addressed in the final commit: fixed a misleading `no BB_RUNNER_* variables found` diagnostic when pruning empties the map, and made the Go test robust to ambient orphan RPC overrides.

### DIFF
--- a/.github/scripts/runner.py
+++ b/.github/scripts/runner.py
@@ -113,6 +113,34 @@ def load_runner_map(vars_map: dict) -> dict[str, str]:
     return mapping
 
 
+def load_staging_coins(vars_map: dict) -> set[str]:
+    # BB_STAGING lists work-in-progress coins (comma/space separated); include the
+    # '-'/'_' variants so it matches either coin-name spelling.
+    staging: set[str] = set()
+    for token in re.split(r"[\s,]+", str(vars_map.get("BB_STAGING") or "")):
+        name = token.strip().lower()
+        if name:
+            staging.update({name, name.replace("-", "_"), name.replace("_", "-")})
+    return staging
+
+
+def prune_staging_coins(
+    workspace: Path, runner_map: dict[str, str], staging: set[str]
+) -> dict[str, str]:
+    # Drop a staging coin only where its config is absent, so an unmerged coin's
+    # repo-wide variables don't break sibling branches while the feature branch that
+    # carries the config still builds and deploys it normally.
+    if not staging:
+        return runner_map
+    kept = {}
+    for coin, runner in runner_map.items():
+        if coin in staging and not coin_config_path(workspace, coin).exists():
+            log(f"BB_STAGING: skipping '{coin}' (no configs/coins/{coin}.json on this branch)")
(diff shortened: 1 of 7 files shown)


cranycrane (approved):
LGTM

PR #1386, opened 2026-01-11
Use effectiveGasPrice for fee calculation on L2 networks

h45hc47:
## Problem
On L2 networks (Arbitrum, Optimism, Base, etc.), the `fees` field in API responses returns incorrect values.

### Example from issue #1227:
Transaction: `0xc53e311af533dc185732c2299f7c585a8a65f8b5319ab82a1b01f03c86c3e97b`
- Blockbook returns: `66373200000000` (incorrect)
- Arbiscan shows: `553110000000` (correct)

## Root Cause
The fee was calculated as `gasPrice * gasUsed`, but `gasPrice` from the transaction is the "Gas Price Bid" (what user was willing to pay), not the actual price paid.
On L2 networks, the actual gas price (`effectiveGasPrice`) is returned in the transaction receipt and can differ significantly from the bid price.

## Solution
1. Added `EffectiveGasPrice` field to `RpcReceipt` struct
2. Added field to protobuf schema for persistent storage
3. Updated `PackTx`/`UnpackTx` to serialize/deserialize the new field
4. Modified fee calculation in `api/worker.go` to use `effectiveGasPrice` when available, with fallback to `gasPrice` for older transactions

## Files Changed
- `bchain/types_ethereum_type.go` - Added EffectiveGasPrice to RpcReceipt
- `bchain/coins/eth/ethtx.proto` - Added field to protobuf schema
- `bchain/coins/eth/ethtx.pb.go` - Regenerated protobuf
- `bchain/coins/eth/ethparser.go` - Updated Pack/Unpack andGetEthereumTxData
- `api/worker.go` - Updated fee calculation logic

## Testing
Verified on Arbitrum transactions from issue #1227 - fees now match Arbiscan values.

PR #1645, opened 2026-07-22
fix(eth): use effectiveGasPrice for L2 transaction fees

### LINKED ISSUE #1227: An incorrect amount of fees is returned in the "fees" field
First example:
Trezor Arbitrum Blockbook (https://arb1.trezor.io/api/v2/tx/0xc53e311af533dc185732c2299f7c585a8a65f8b5319ab82a1b01f03c86c3e97b) returns 66373200000000 as the commission amount. The correct amount of commissions is 553110000000 (https://arbiscan.io/tx/0xc53e311af533dc185732c2299f7c585a8a65f8b5319ab82a1b01f03c86c3e97b).

Second example:
Trezor Arbitrum Blockbook (hhttps://arb1.trezor.io/api/v2/tx/0x267c4c6c8d084e1c1a3bd977ee560137153eafccbb81442e1b65c545e6c25841) returns 304217352000 as the commission amount. The correct amount of commissions is 276691352000 (https://arbiscan.io/tx/0x267c4c6c8d084e1c1a3bd977ee560137153eafccbb81442e1b65c545e6c25841).

Presumably the incorrect calculation of commissions is due to the fact that the gasPrice field contains “Gas Price Bid” and not “Gas Price Paid”, which takes part in the formula for calculating the final amount of commissions.

pragmaxim:
## Problem

On L2 networks (Arbitrum, Optimism, Base, …) the `fees` field returned by the API is wrong. Reported in #1227:

- tx `0xc53e311a…c3e97b` — Blockbook returns `66373200000000`, Arbiscan shows `553110000000`.
- tx `0x267c4c6c…6c25841` — Blockbook returns `304217352000`, Arbiscan shows `276691352000`.

## Root cause

The fee was computed as `gasPrice * gasUsed`, but a transaction's `gasPrice` on L2s is only the **gas price bid**, not the price actually paid. The price actually charged is the receipt's **`effectiveGasPrice`**, which can differ substantially from the bid.

## Fix

Compute the fee from `effectiveGasPrice` when available, falling back to `gasPrice` for older/legacy transactions, and add the separate `l1Fee` where a rollup reports one (OP stack; Arbitrum bakes the L1 cost into `gasUsed` and reports no `l1Fee`):

fee = (effectiveGasPrice if > 0 else gasPrice) × gasUsed + l1Fee

- **Persistence** — `EffectiveGasPrice` added to the eth receipt protobuf (`optional bytes … = 8`) and Pack/Unpack, so it is stored per transaction. `ethtx.pb.go` was regenerated but kept in the repo's existing `string([]byte{…})` `rawDesc` format (bytes produced with `protoc` and verified to reproduce the committed descriptor byte-for-byte on the unmodified proto), so the generated diff stays minimal.
- **Fee calculation** — a shared `getEthereumFeesSat` helper is now used by both the transaction path and the balance-history path. Balance history additionally gains `l1Fee`, which it previously omitted (addresses the earlier review comment on #1386).
- **API surface** — `effectiveGasPrice` is exposed in `EthereumSpecific` (`api/types.go`), in `openapi.yaml`, and in `blockbook-api.ts` (hand-edited, not regenerated).
- **Web explorer** — the tx templates now show *Effective Gas Price* and base the `Gwei/gas` fee annotation on it, so the displayed per-gas price matches the fee on L2s.
- **Tests** — unit tests for the fee calc (issue scenario, OP-stack `l1Fee`, legacy fallback, zero/nil, mempool guard), for the balance-history accounting, and for the pack/unpack round trip.

## Note on already-indexed data

Like every receipt-field addition, this is backward compatible: old records unpack cleanly and fall back to `gasPrice` (unchanged historical behavior, no crash). New blocks store `effectiveGasPrice` and get correct fees automatically; **already-indexed L2 blocks need a resync to retroactively correct their fees.**

Note that this PR reimplements #1386 by @h45hc47 on a fresh branch off current `master` (the original PR conflicts too much since the code progressed in this area), so the credit goes to him and I appreciate his help.

Fixes #1227.


cranycrane (approved):
LGTM

PR #1646, opened 2026-07-22
feat(server): add request timeouts and resource bounds for REST, WebSocket, and address lookups

cranycrane:
## Summary

Adds explicit resource limits across the HTTP, WebSocket, and database-lookup paths so that a single slow or misbehaving client cannot tie up server resources indefinitely.

### HTTP server timeouts (`server/timeouts.go`, `public.go`, `internal.go`)

Go's `http.Server` leaves all timeouts disabled by default, so connections that send the request line and headers very slowly (or never finish them) can accumulate and pin goroutines and file descriptors. Both the public (REST/UI) and internal (admin/metrics) servers now set explicit `ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, `IdleTimeout`, and `MaxHeaderBytes`. The values are deliberately generous so large paginated API and explorer responses over slow links are never truncated.

### WebSocket (`server/websocket.go`)

- Bound the decompressed message payload to the existing 4 MB message limit via a safe read (`readLimitedMessage`), so a small compressed frame cannot expand into an oversized buffer.
- Add a handshake timeout to the upgrader.
- Clear the inherited HTTP read/write deadlines on upgraded connections so long-lived subscription streams are unaffected; `outputLoop` continues to set its own per-message write deadlines.

### Address-descriptor lookups (`db/rocksdb.go`, `api/worker.go`)

`GetAddrDescTransactions` could visit an unbounded number of keys when given a truncated address descriptor (e.g. `ad:76a9`), which is a prefix of a large slice of the addresses column family. The scan now aborts after skipping a bounded number of non-matching keys, and the API rejects descriptors that do not resolve to a real address (account-based chains still rely on the scan bound). Added `TestGetAddrDescTransactions_ScanBound` covering this.

## Test plan

- `go test ./db/ -run TestGetAddrDescTransactions_ScanBound`
- Existing `server/` and `db/` tests

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -1483,23 +1483,11 @@ func (w *Worker) getAddrDescAndNormalizeAddress(address string) (bchain.AddressD
 	addrDesc, err := w.chainParser.GetAddrDescFromAddress(address)
 	if err != nil {
 		// try if the address is not address descriptor converted to string
-		ad, errAd := bchain.AddressDescriptorFromString(address)
+		var errAd error
+		addrDesc, errAd = bchain.AddressDescriptorFromString(address)
 		if errAd != nil {
 			return nil, "", NewAPIError(fmt.Sprintf("Invalid address, %v", err), true)
 		}
-		// The descriptor is arbitrary user-supplied hex of any length. A short
-		// descriptor such as ad:76a9 is a prefix of a large slice of the
-		// addresses column family and would otherwise make
-		// GetAddrDescTransactions scan a large part of it. Reject descriptors
-		// that do not resolve to a real address; complete descriptors (standard
-		// scripts, multisig, and account-based addresses) always resolve to at
-		// least one. On account-based chains every hex string resolves to an
-		// address, so those rely on the scan bound in
-		// db.GetAddrDescTransactions instead.
-		if addrs, _, e := w.chainParser.GetAddressesFromAddrDesc(ad); e != nil || len(addrs) == 0 {
-			return nil, "", NewAPIError(fmt.Sprintf("Invalid address, %v", err), true)
-		}
-		addrDesc = ad
 	}
 	// convert the address to the format defined by the parser
 	addresses, _, err := w.chainParser.GetAddressesFromAddrDesc(addrDesc)
--- a/api/worker.go
+++ b/api/worker.go
@@ -1482,12 +1482,23 @@ func (w *Worker) txFromTxid(txid string, bestHeight uint32, option AccountDetail
 func (w *Worker) getAddrDescAndNormalizeAddress(address string) (bchain.AddressDescriptor, string, error) {
 	addrDesc, err := w.chainParser.GetAddrDescFromAddress(address)
 	if err != nil {
-		var errAd error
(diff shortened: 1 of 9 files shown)


api/worker.go
```diff
@@ -1487,12 +1487,24 @@ func (w *Worker) txFromTxid(txid string, bestHeight uint32, option AccountDetail
...
+		// The descriptor is arbitrary user-supplied hex of any length. A short
+		// descriptor such as ad:76a9 is a prefix of a large slice of the
+		// addresses column family and would otherwise make
+		// GetAddrDescTransactions scan a large part of it. Reject descriptors
+		// that do not resolve to a real address; complete descriptors (standard
+		// scripts, multisig, and account-based addresses) always resolve to at
+		// least one. On account-based chains every hex string resolves to an
+		// address, so those rely on the scan bound in
+		// db.GetAddrDescTransactions instead.
+		if addrs, _, e := w.chainParser.GetAddressesFromAddrDesc(ad); e != nil || len(addrs) == 0 {
```
pragmaxim: **Behavioral regression: non-standard-but-indexed scripts become unqueryable by full descriptor.**

`IsAddrDescIndexable` indexes *any* script except empty/`OP_RETURN`, including `NonStandardTy` scripts, but `outputScriptToAddresses` returns `len==0, nil` for those. So this `len(addrs) == 0` guard will reject full, well-formed descriptors for non-standard scripts that are actually indexed and previously returned transactions. Bare pubkey/multisig still resolve (non-empty), so the blast radius is narrow, but this is a real behavior change for a genuine descriptor — not just truncated garbage.

Suggest either calling this out in the PR description / changelog as an intentional tradeoff, or tightening the guard (e.g. reject only when the descriptor is shorter than a plausible minimum) if that lookup path matters to consumers.

server/timeouts.go
```diff
@@ -0,0 +1,19 @@
...
+// HTTP server timeouts for the REST/UI (public) and admin/metrics (internal) interfaces;
+// Go's http.Server disables them all by default, so without these a connection that stalls before reaching
+// a handler holds a goroutine and file descriptor indefinitely.
+const (
+	// httpReadHeaderTimeout bounds the time a client may take to send the request line and headers
+	httpReadHeaderTimeout = 10 * time.Second
+	// httpReadTimeout bounds the whole request read
+	httpReadTimeout = 30 * time.Second
+	// httpWriteTimeout bounds response writes
+	httpWriteTimeout = 120 * time.Second
```
pragmaxim: **`httpWriteTimeout = 120s` is a hard cap on total handler+write time.**

Go's `WriteTimeout` covers handler execution *plus* response write. Heavy but legitimate requests — e.g. large xpub queries with full tx detail against a busy backend — can occasionally exceed 120s and would now be cut off mid-response. This is the one timeout here that can truncate a *valid* slow response rather than only killing an abusive/idle connection. Worth confirming 120s comfortably covers the worst-case xpub/blockbook queries under load.

pragmaxim (approved):
LGTM

PR #1649, opened 2026-07-22
fix: make --repair honest about what it can and cannot fix (#1413)

### LINKED ISSUE #1413: --repair flag doesn't fix database inconsistencies
The `--repair` flag often leaves the database in an inconsistent state after running. Users report needing to run it multiple times or it not resolving issues at all.

### Root Cause

The current `RepairRocksDB()` function only calls RocksDB's low-level repair API with default options. This has several limitations:

1. **Wrong database options** - Uses `NewDefaultOptions()` instead of the same options used during normal operation (bloom filters, column families, compression settings, etc.)
2. **Doesn't reset inconsistent state** - If database is marked as `DbStateInconsistent` (from interrupted bulk import), the flag remains set and Blockbook refuses to start
3. **Only fixes physical corruption** - Handles corrupted SST files/manifests, but not application-level issues like UTXO balance mismatches or missing transaction references

### Current Code

func RepairRocksDB(name string) error {
    glog.Infof("rocksdb: repair")
    opts := grocksdb.NewDefaultOptions()
    return grocksdb.RepairDb(name, opts)
}### Proposed Solution

1. Use proper database options matching `createAndSetDBOptions()` during repair
2. After repair, check and reset `DbStateInconsistent` flag if present
3. Better document what `--repair` actually fixes vs `--fixutxo` for UTXO inconsistencies

### Workaround

- **For `DbStateInconsistent` errors:** Delete database and re-import
- **For UTXO balance issues (Bitcoin-type only):** Use `--fixutxo` flag instead
- **For physical corruption:** Current `--repair` may work, but results vary

pragmaxim:
Fixes #1413.

## Problem

`RepairRocksDB` called RocksDB's low-level `RepairDb` with default options and nothing else. The attempt to *also* reset `DbStateInconsistent` after repair reopened the database with a single-default-column-family `grocksdb.OpenDb`. Blockbook's database is **always multi-column-family**, and RocksDB refuses a single-CF open of a multi-CF database (*"You have to open all column families…"*). Consequently:

- The state-reset block was **dead code** on any real database — it never ran.
- Its error was **swallowed** (`glog.Warning` + `return nil`), so `blockbook --repair` exited `0` (success) while the inconsistent flag stayed set; the next normal start still aborted with *"database is in inconsistent state and cannot be used."*
- `RepairDb` takes a single `Options` (no per-CF descriptors) and drops column families with no live SST files, so in the narrow case the reset *did* run it could certify a gutted index as healthy.

There is also a deeper semantic issue: `DbStateInconsistent` is set **only** by `InitBulkConnect` and cleared only on a successful `CloseBulkConnect`. It means *an initial/bulk import was interrupted* — the index is **incomplete** (writes were never issued, and bulk flushes don't align to block boundaries, so there is no consistent height to resume from). Physical SST repair cannot reconstruct never-written data, so blindly clearing the flag serves a permanently-corrupt index.

## What this PR does

Reworks `RepairRocksDB(name string, forceRepair bool)`:

- **Physical repair with production-matching options.** Uses `createAndSetDBOptions` (bloom filter table factory) instead of bare defaults. Honest about grocksdb's single-`Options`/no-per-CF limitation.
- **Correct multi-CF reopen.** Discovers the column families from the database itself via `ListColumnFamilies` (the `cfNames` package global is empty during `--repair`, which runs before the chain/parser is loaded) and reopens **all** of them with the same per-CF options as normal operation (the `addresses` CF gets no bloom filter, mirroring `openDB`). This eliminates the dead-code bug.
- **Data-loss detection.** Compares the CF set before vs. after `RepairDb`; a dropped CF that held rows (per the tracked internal-state column stats) is reported as an error rather than silently accepted. (Dropping *empty* CFs — commonly `fiatRates`/`blockFilter`/`transactions` and several Ethereum CFs — is normal and benign; they're recreated empty on next open.)
- **Honest state handling.** An interrupted bulk import (`DbStateInconsistent`) is **refused by default** with an actionable message telling the operator the database must be re-imported. The new `-forcerepair` flag forces it into `DbStateOpen` (not `DbStateClosed`, matching `SetInconsistentState(false)` semantics) with loud warnings that the index may be incomplete and a full resync is strongly advised.
- **Real error propagation.** Every genuine failure returns an error, so `blockbook --repair` exits non-zero instead of a misleading `0`.
- **Resource cleanup.** All cgo objects (Options/Cache/ReadOptions/WriteOptions/CF handles/DB) are freed with correct LIFO teardown.
- **Docs & tests.** Honest doc comment; `changelog.md` and `docs/rocksdb.md` now spell out `--repair` (physical SST/MANIFEST corruption) vs `-fixutxo` (UTXO/balance recompute, Bitcoin-type) vs re-import (`DbStateInconsistent`). Adds 5 tests covering healthy round-trip, inconsistent-refuse, `-forcerepair`, the CF-diff helper, and the nonexistent-path error.

## Design note

Issue #1413 proposed simply "reset the `DbStateInconsistent` flag." This PR intentionally does **not** do that by default, because an interrupted bulk import is genuinely unrecoverable by repair — resetting it would boot a truncated index as if healthy. The `-forcerepair` escape hatch preserves that capability for operators who knowingly accept an incomplete index over a full resync, but it is off by default and warns loudly.

## Testing

```
go build ./db/... .                                        # ok
go test -tags unittest -run 'TestRepair|TestMissingColumnFamilies' ./db/   # 5/5 pass
go test -tags unittest ./db/                               # full suite, no regressions
```

The tests were validated to actually guard the regression: re-introducing the single-CF-open + swallow-error bug makes `TestRepairRocksDB_Inconsistent_Refuses` and `TestRepairRocksDB_Inconsistent_ResetForcesOpen` fail.


pragmaxim:
Recreating this PR directly from origin/trezor instead of from the pragmaxim fork

PR #1650, opened 2026-07-22
fix: make --repair honest about what it can and cannot fix (#1413)

### LINKED ISSUE #1413: --repair flag doesn't fix database inconsistencies

pragmaxim:
Fixes #1413.

## Problem

`RepairRocksDB` called RocksDB's low-level `RepairDb` with default options and nothing else. The attempt to *also* reset `DbStateInconsistent` after repair reopened the database with a single-default-column-family `grocksdb.OpenDb`. Blockbook's database is **always multi-column-family**, and RocksDB refuses a single-CF open of a multi-CF database (*"You have to open all column families…"*). Consequently:

- The state-reset block was **dead code** on any real database — it never ran.
- Its error was **swallowed** (`glog.Warning` + `return nil`), so `blockbook --repair` exited `0` (success) while the inconsistent flag stayed set; the next normal start still aborted with *"database is in inconsistent state and cannot be used."*
- `RepairDb` takes a single `Options` (no per-CF descriptors) and drops column families with no live SST files, so in the narrow case the reset *did* run it could certify a gutted index as healthy.

There is also a deeper semantic issue: `DbStateInconsistent` is set **only** by `InitBulkConnect` and cleared only on a successful `CloseBulkConnect`. It means *an initial/bulk import was interrupted* — the index is **incomplete** (writes were never issued, and bulk flushes don't align to block boundaries, so there is no consistent height to resume from). Physical SST repair cannot reconstruct never-written data, so blindly clearing the flag serves a permanently-corrupt index.

## What this PR does

Reworks `RepairRocksDB(name string, forceRepair bool)`:

- **Physical repair with production-matching options.** Uses `createAndSetDBOptions` (bloom filter table factory) instead of bare defaults. Honest about grocksdb's single-`Options`/no-per-CF limitation.
- **Correct multi-CF reopen.** Discovers the column families from the database itself via `ListColumnFamilies` (the `cfNames` package global is empty during `--repair`, which runs before the chain/parser is loaded) and reopens **all** of them with the same per-CF options as normal operation (the `addresses` CF gets no bloom filter, mirroring `openDB`). This eliminates the dead-code bug.
- **Data-loss detection.** Compares the CF set before vs. after `RepairDb`; a dropped CF that held rows (per the tracked internal-state column stats) is reported as an error rather than silently accepted. (Dropping *empty* CFs — commonly `fiatRates`/`blockFilter`/`transactions` and several Ethereum CFs — is normal and benign; they're recreated empty on next open.)
- **Honest state handling.** An interrupted bulk import (`DbStateInconsistent`) is **refused by default** with an actionable message telling the operator the database must be re-imported. The new `-forcerepair` flag forces it into `DbStateOpen` (not `DbStateClosed`, matching `SetInconsistentState(false)` semantics) with loud warnings that the index may be incomplete and a full resync is strongly advised.
- **Real error propagation.** Every genuine failure returns an error, so `blockbook --repair` exits non-zero instead of a misleading `0`.
- **Resource cleanup.** All cgo objects (Options/Cache/ReadOptions/WriteOptions/CF handles/DB) are freed with correct LIFO teardown.
- **Docs & tests.** Honest doc comment; `changelog.md` and `docs/rocksdb.md` now spell out `--repair` (physical SST/MANIFEST corruption) vs `-fixutxo` (UTXO/balance recompute, Bitcoin-type) vs re-import (`DbStateInconsistent`). Adds 5 tests covering healthy round-trip, inconsistent-refuse, `-forcerepair`, the CF-diff helper, and the nonexistent-path error.

## Design note

Issue #1413 proposed simply "reset the `DbStateInconsistent` flag." This PR intentionally does **not** do that by default, because an interrupted bulk import is genuinely unrecoverable by repair — resetting it would boot a truncated index as if healthy. The `-forcerepair` escape hatch preserves that capability for operators who knowingly accept an incomplete index over a full resync, but it is off by default and warns loudly.

## Testing

```
go build ./db/... .                                        # ok
go test -tags unittest -run 'TestRepair|TestMissingColumnFamilies' ./db/   # 5/5 pass
go test -tags unittest ./db/                               # full suite, no regressions
```

The tests were validated to actually guard the regression: re-introducing the single-CF-open + swallow-error bug makes `TestRepairRocksDB_Inconsistent_Refuses` and `TestRepairRocksDB_Inconsistent_ResetForcesOpen` fail.


db/rocksdb.go
```diff
@@ -32,11 +32,214 @@ const maxAddrDescLen = 1024
...
+
+	// Data-loss cross-check on dropped CFs. is.DbColumns row counts are a reliable POSITIVE
+	// signal only (Rows>0 => the CF definitely held data at last stats computation), but an
+	// unreliable negative (only cfTransactions is tracked during normal sync; full per-CF
+	// stats need a prior --computedbstats run). So this catches genuine physical loss of a
+	// populated CF when stats exist and warns otherwise; the interrupted-bulk-import gutting
+	// scenario is caught independently by the DbStateInconsistent gate below.
+	var lost []string
+	for _, cf := range dropped {
+		if rows, ok := columnRows(is, cf); ok && rows > 0 {
```
cranycrane, line 178: **Possible false-positive data-loss refusal from stale `--computedbstats` counts.**

This check treats a dropped CF as unrecoverable loss whenever `is.DbColumns` reports `Rows > 0`, but that count is only kept fresh for `cfTransactions` during normal sync — every other CF's count is set only by an explicit `--computedbstats` run and is never decremented afterward.

_Scenario:_ operator runs `--computedbstats` (records e.g. `blockFilter`/`transactions` `Rows=N`), a later rollback/pruning empties that CF and its SSTs get compacted away, but stats are never recomputed. A subsequent `--repair` on a still-readable MANIFEST lists the CF in `beforeCFs`, `RepairDb` legitimately drops the now-empty CF, and this loop reports it as `lost` and aborts with *"must be re-imported/resynced"* — forcing an unnecessary full resync of a perfectly repairable DB.

Consider reconciling the stale count against the CF's actual current contents rather than trusting `Rows` alone.

db/rocksdb.go
```diff
@@ -32,11 +32,214 @@ const maxAddrDescLen = 1024
...
+
+	switch is.DbState {
+	case common.DbStateInconsistent:
+		if !resetInconsistent {
+			return errors.Errorf("database is in inconsistent state: an initial/bulk import was interrupted, the index is incomplete and cannot be recovered by --repair; the database must be re-imported (resynced from scratch). Re-run with -forcerepair to force it into open state (data may be incomplete; a full resync is strongly advised)")
+		}
+		glog.Warning("rocksdb: -forcerepair: forcing inconsistent database into open state; the index may be INCOMPLETE, a full resync is strongly advised")
+		// DbStateOpen, not DbStateClosed — matches SetInconsistentState(false) semantics;
+		// Closed would falsely claim a graceful shutdown.
+		is.DbState = common.DbStateOpen
```
cranycrane, line 201: **`-forcerepair` resumes forward on an incomplete index; the resync warning is advisory only.**

Flipping `DbStateInconsistent → DbStateOpen` makes the next normal start treat the DB as an ordinary ungraceful shutdown and resume from the persisted `BestHeight`. But an interrupted bulk import can leave `BestHeight` ahead of the block data actually written (bulk flushes don't align to block boundaries, as the doc comment notes), so ranges below `BestHeight` are never re-fetched and the index keeps permanent gaps — while the API serves incomplete results.

Nothing at runtime marks the index incomplete; the *"full resync is strongly advised"* text is only a repair-time log line and is trivially ignored. Worth considering whether `-forcerepair` should force a state that guarantees a full resync (or persist a durable 'index-incomplete' marker) rather than plain `DbStateOpen`.

blockbook.go
```diff
@@ -169,7 +170,7 @@ func mainWithExitCode() int {
 	}
 
 	if *repair {
-		if err := db.RepairRocksDB(*dbPath); err != nil {
+		if err := db.RepairRocksDB(*dbPath, *forceRepair); err != nil {
```
cranycrane, line 178: **`-forcerepair` without `-repair` is a silent no-op.**

`*forceRepair` is only read inside this `if *repair` block, so `blockbook -forcerepair <path>` (without `-repair`) does nothing, prints no warning, and falls through to normal startup — which then aborts on the inconsistent DB. The flag help even says *"with -repair"*, but nothing enforces or warns about the missing `-repair`. Consider erroring/warning when `-forcerepair` is set without `-repair`.

cranycrane (commented):
Automated code-review pass — three correctness edge cases in the `--repair` rework. No crash/compile issues found; these are latent edge cases and an operational footgun.

db/rocksdb.go
pragmaxim: Fixed in b717c203. Agreed — the `Rows>0` heuristic can't distinguish real loss from a CF legitimately emptied by a rollback/prune with stale `--computedbstats` counts, and a dropped CF is empty *now* by definition (RepairDb only drops CFs with no live SSTs). Rather than reconcile against contents we no longer have post-drop, the check no longer aborts: a drop of a previously-populated CF now logs a prominent **warning** (not a hard error), and benign empty-CF drops log at Info. The authoritative unrecoverable case (interrupted bulk import) is still caught by the `DbStateInconsistent` gate, so no false-positive resync is forced.

db/rocksdb.go
pragmaxim: Fixed in b717c203 (messaging/semantics). You're right that `DbStateOpen` resumes forward from the persisted best height and never backfills blocks below it, so gaps from an interrupted bulk import are permanent. Rather than pretend otherwise, `-forcerepair` is now documented and logged as a **startability escape hatch, not a recovery**: the refuse message, the runtime warnings, the doc comment and the changelog all now state the index may remain PERMANENTLY incomplete and that a full **recreate** (delete + re-import) is the only guaranteed fix — replacing the misleading "a full resync is advised". A durable machine-readable 'index-incomplete' marker surfaced through the API would be a stronger guarantee, but that's a larger change (new InternalState field + API wiring); flagging it as a possible follow-up rather than scope-creeping this PR. Open to adding it if you'd prefer.

blockbook.go
pragmaxim: Fixed in b717c203. `blockbook -forcerepair` without `-repair` now fails fast with `exitCodeFatal` and a clear message ("-forcerepair has no effect without -repair; re-run with both flags...") instead of silently falling through to normal startup.

cranycrane (approved):
LGTM

PR #1652, opened 2026-07-23
fix(eth): restrict ENS aliases to trusted registrars; index modern ENS controllers

### LINKED ISSUE #1622: ENS address aliases are unvalidated and not reverted on reorg
### Summary

The ENS / address-alias handling has several correctness and robustness bugs. Alias records parsed from `NameRegistered` logs are stored with insufficient validation and are not kept consistent across chain reorganizations.

### Affected code

- `bchain/coins/eth/dataparser.go` — `getEnsRecord`
- `bchain/coins/eth/ethrpc.go` — `processEventsForBlock`
- `db/rocksdb.go` — `storeAddressAliasRecords` / `GetAddressAlias`
- `db/rocksdb_ethereumtype.go` — `storeBlockSpecificDataEthereumType` / `DisconnectBlockRangeEthereumType`

### Details

1. **No emitter check.** `getEnsRecord` runs for every log in a block and matches only on the event signature (`Topics[0]`). It does not verify the log was emitted by the ENS registrar/controller contract, so any contract emitting a matching event can register an alias for an arbitrary address.
2. **No name validation.** The alias name is decoded straight from log data into a string with no UTF-8 validation, no length limit, and no control-character / normalization checks.
3. **Unverified, last-write-wins storage.** `storeAddressAliasRecords` overwrites `cfAddressAliases[address]` unconditionally — no ownership check and no comparison against the existing value, so a later write can silently replace a correct label.
4. **Not reorg-safe.** `DisconnectBlockRangeEthereumType` does not revert alias records written by disconnected blocks, and the in-memory `cachedAddressAliasRecords` LRU is never invalidated. Aliases from orphaned blocks persist across reorgs and restarts.
5. **Harvested where ENS can't be validated.** `address_aliases` is enabled on several non-mainnet / L2 configs, while ENS resolution is mainnet-only — so records are harvested and stored but can never be validated on those chains.

### Impact

Incorrect or misleading address labels can be stored and served through the API/UI, including overwriting previously correct labels, and stale labels can persist after a reorg.

### Suggested fix

- Only accept `NameRegistered` events emitted by the known ENS registrar/controller contract(s).
- Validate alias names: enforce UTF-8, a byte-length cap, and reject/normalize control and bidirectional characters.
- Track alias writes per block height and revert them on disconnect; invalidate the LRU cache accordingly.
- Consider disabling `address_aliases` on chains where ENS cannot be validated.

cranycrane:
## Summary

Makes ENS address-alias handling correct and comprehensive.

`getEnsRecord` recorded an alias from any `NameRegistered` event, matching only the event signature and never the emitting contract. Any contract that emitted a look-alike event — with `topics[2]` set to an arbitrary address — had its alias written to the DB and later served by the API and explorer as if it were a genuine ENS name, so the stored aliases were unreliable.

## Changes

- **Emitter allowlist** — `getEnsRecord` now only accepts events from contracts in a per-chain trusted set (new `ens_registrars` config key). Semantics are **trust-none by default**: absent/empty records nothing, an address list restricts to those emitters, and `"*"` accepts any (explicit opt-in). Ethereum pins all four ENS `ETHRegistrarController` addresses (2019, 2020, 6-arg, current 7-arg).
- **Full controller coverage** — match the original 5-arg `NameRegistered`, the 6-arg (baseCost + premium, `0x69e37f15…`) and the **current 7-arg** event (referrer, `0xc224…`) emitted by the live mainnet controller. The name is parsed via its ABI offset word, so all layouts work uniformly. Without the 7-arg match, every registration through the live controller was silently dropped.
- **Name validation** — decoded alias names are rejected if empty, over-long, non-UTF-8, or carry control characters, bidirectional overrides, invisible format characters (`Cf`, except ZWJ/ZWNJ), or line/paragraph separators, so a stored/displayed alias can't hide or misrepresent its true content.
- **Config validation** — malformed `ens_registrars` entries are logged and dropped so a typo doesn't silently disable a controller.
- **BSC `.bnb`** — the Space ID `.bnb` name service is non-functional in practice, so no new aliases are recorded (trust-none); the `.bnb` suffix is kept so any legacy records render with their original label rather than a misleading `.eth` one. `address_aliases` stays on (it also drives contract/token-name display).
- **Internal** — the trusted-emitter set lives on `EthereumRPC` (not the parser), keeping it out of the `EthereumLikeParser` interface.

## Client compatibility

The API response shape (`AddressAliasesMap` / `AddressAlias{Type, Alias}`) is **unchanged** — trezor-suite keeps receiving the same JSON; only which aliases are recorded changes.

## Operational notes / known limitations

- **Applies to newly synced data — refresh existing data with `-rebuildensaliases` (no full reindex needed).** The emitter allowlist gates parsing during sync and does not retroactively touch aliases the old accept-any code wrote. Because `cfAddressAliases` is a pure projection of `NameRegistered` logs (nothing else writes it; contract/token names live in a separate CF), the one-shot `-rebuildensaliases` flag rebuilds just that column family: it runs at startup holding the DB exclusively, wipes `cfAddressAliases` + the alias cache, then rescans `NameRegistered` logs from the trusted registrars via chunked `eth_getLogs` (filtered by `address` + the three `NameRegistered` topics) and re-derives aliases through the same `getEnsRecord` path used in sync. This both **purges** spoofed aliases the old code stored and **backfills** the modern 6-/7-arg registrations it never matched. Scans block 0 → backend tip by default (`-blockheight`/`-blockuntil` to bound); tune the per-request span with `-ensrebuildchunk` (default 100000) for nodes that cap `getLogs` ranges. A trust-none chain (empty `ens_registrars`, e.g. BSC) is wiped but records nothing. A full reindex is no longer required for alias correctness.
- **Non-mainnet chains index no ENS aliases** unless their `ens_registrars` is populated (e.g. Sepolia has a real ENS deployment but is intentionally left empty here). This is by design — `address_aliases` on those chains still serves contract/token names — so there is deliberately no startup warning for the `address_aliases && empty ens_registrars` combination, which is a valid "contract-names-only" configuration.
- **Alias name length is capped at 256 bytes.** Real ENS names are far shorter; a longer name (unconfirmed to exist on mainnet) would be dropped on a fresh sync.

## Testing

Unit tests cover the 5-/6-/7-arg event layouts, untrusted-emitter rejection, trust-none and wildcard sets, and the name-validation matrix (including invisible-character cases). `go build` + ENS tests pass; configs valid JSON; `gofmt` clean.

> Note: branch is based on `981d3d65` and is ~8 commits behind `master`; two touched files (`ethparser.go`, `ethrpc.go`) also changed in master's recent L2-fee work, so a rebase may be needed before merge.

### DIFF
--- a/bchain/coins/eth/dataparser.go
+++ b/bchain/coins/eth/dataparser.go
@@ -303,9 +303,47 @@ func (p *EthereumParser) ParseInputData(signatures *[]bchain.FourByteSignature, 
 	return &parsed
 }
 
-// getEnsRecord processes transaction log entry and tries to parse ENS record from it
-func getEnsRecord(l *rpcLogWithTxHash) *bchain.AddressAliasRecord {
+// maxEnsAliasNameLen caps the decoded ENS name length (in bytes) accepted as an
+// address alias, bounding storage and display abuse from an oversized log.
+const maxEnsAliasNameLen = 256
+
+// validEnsAliasName rejects names unsafe to store or render as an address alias:
+// empty, over the length cap, not valid UTF-8, or containing control characters
+// (NUL, newline, ANSI escapes) or bidirectional overrides (e.g. U+202E) that could
+// spoof or corrupt the displayed alias. Bidi_Control is used rather than the whole
+// Cf category so legitimate emoji names using the zero-width joiner still pass.
+func validEnsAliasName(name string) bool {
+	if len(name) == 0 || len(name) > maxEnsAliasNameLen {
+		return false
+	}
+	if !utf8.ValidString(name) {
+		return false
+	}
+	for _, r := range name {
+		if unicode.IsControl(r) || unicode.Is(unicode.Bidi_Control, r) {
+			return false
+		}
+	}
+	return true
+}
+
+// getEnsRecord processes transaction log entry and tries to parse ENS record from it.
+// registrars is the set of contract addresses (lower-cased, 0x-prefixed) trusted to
+// emit NameRegistered events: a log whose emitter (l.Address) is not in the set is
+// rejected, so an arbitrary contract cannot forge an ENS alias for any address. An
+// empty set therefore trusts no one and records nothing; the special entry "*"
+// accepts any emitter (legacy behavior, for chains with a different name service).
+func getEnsRecord(l *rpcLogWithTxHash, registrars map[string]struct{}) *bchain.AddressAliasRecord {
 	if len(l.Topics) == 3 && l.Topics[0] == nameRegisteredEventSignature && len(l.Data) >= 322 {
+		if _, acceptAny := registrars[ensRegistrarWildcard]; !acceptAny {
+			emitter := strings.ToLower(l.Address)
+			if !strings.HasPrefix(emitter, "0x") {
+				emitter = "0x" + emitter
+			}
+			if _, ok := registrars[emitter]; !ok {
+				return nil
+			}
+		}
 		address, err := addressFromPaddedHex(l.Topics[2])
 		if err != nil {
 			return nil
@@ -329,7 +367,11 @@ func getEnsRecord(l *rpcLogWithTxHash) *bchain.AddressAliasRecord {
 		if err != nil {
 			return nil
 		}
-		return &bchain.AddressAliasRecord{Address: address, Name: string(b)}
+		name := string(b)
+		if !validEnsAliasName(name) {
+			return nil
+		}
+		return &bchain.AddressAliasRecord{Address: address, Name: name}
 	}
 	return nil
 }
--- a/bchain/coins/eth/dataparser.go
+++ b/bchain/coins/eth/dataparser.go
@@ -327,51 +327,83 @@ func validEnsAliasName(name string) bool {
 	return true
 }
 
-// getEnsRecord processes transaction log entry and tries to parse ENS record from it.
-// registrars is the set of contract addresses (lower-cased, 0x-prefixed) trusted to
-// emit NameRegistered events: a log whose emitter (l.Address) is not in the set is
-// rejected, so an arbitrary contract cannot forge an ENS alias for any address. An
-// empty set therefore trusts no one and records nothing; the special entry "*"
-// accepts any emitter (legacy behavior, for chains with a different name service).
+// getEnsRecord processes a transaction log entry and tries to parse an ENS
+// address alias (a NameRegistered event) from it. Both the original 5-argument
+// event (old ETHRegistrarController) and the newer 6-argument event that split
+// cost into baseCost + premium are recognized. registrars is the set of contract
+// addresses (lower-cased, 0x-prefixed) trusted to emit these events: a log whose
+// emitter (l.Address) is not in the set is rejected, so an arbitrary contract
+// cannot forge an ENS alias for any address. An empty set therefore trusts no
+// one and records nothing; the special entry "*" accepts any emitter (legacy
+// behavior, for chains with a different name service).
 func getEnsRecord(l *rpcLogWithTxHash, registrars map[string]struct{}) *bchain.AddressAliasRecord {
-	if len(l.Topics) == 3 && l.Topics[0] == nameRegisteredEventSignature && len(l.Data) >= 322 {
-		if _, acceptAny := registrars[ensRegistrarWildcard]; !acceptAny {
-			emitter := strings.ToLower(l.Address)
-			if !strings.HasPrefix(emitter, "0x") {
-				emitter = "0x" + emitter
-			}
-			if _, ok := registrars[emitter]; !ok {
-				return nil
-			}
-		}
-		address, err := addressFromPaddedHex(l.Topics[2])
-		if err != nil {
-			return nil
-		}
-		c, err := strconv.ParseInt(l.Data[194:194+64], 16, 64)
-		if err != nil {
-			return nil
-		}
-		const nameStart = 194 + 64
-		// int(c)<<1 can overflow: c is attacker-controlled (up to 2^63-1) and Go's
-		// signed left shift wraps silently, so de can land anywhere in [0, 257]
-		// below nameStart. The lower bound must be the slice
-		// start index, not 0 — a de below the start makes l.Data[nameStart:de] a
-		// low>high slice expression that panics. Checking de < nameStart guarantees
-		// nameStart <= de <= len(l.Data), so the slice below can never panic.
-		de := nameStart + (int(c) << 1)
-		if de > len(l.Data) || de < nameStart {
-			return nil
-		}
-		b, err := hex.DecodeString(l.Data[nameStart:de])
-		if err != nil {
-			return nil
-		}
-		name := string(b)
-		if !validEnsAliasName(name) {
-			return nil
-		}
-		return &bchain.AddressAliasRecord{Address: address, Name: name}
+	if len(l.Topics) != 3 {
+		return nil
 	}
-	return nil
+	if l.Topics[0] != nameRegisteredEventSignature && l.Topics[0] != nameRegisteredWithPremiumEventSignature {
+		return nil
+	}
+	if _, acceptAny := registrars[ensRegistrarWildcard]; !acceptAny {
+		emitter := strings.ToLower(l.Address)
+		if !strings.HasPrefix(emitter, "0x") {
+			emitter = "0x" + emitter
+		}
+		if _, ok := registrars[emitter]; !ok {
+			return nil
+		}
+	}
+	address, err := addressFromPaddedHex(l.Topics[2])
+	if err != nil {
+		return nil
+	}
+	name, ok := parseEnsNameFromLogData(l.Data)
+	if !ok || !validEnsAliasName(name) {
+		return nil
+	}
+	return &bchain.AddressAliasRecord{Address: address, Name: name}
+}
+
+// readHexWordAt reads one 32-byte EVM word starting at hexStart (an index into
+// the 0x-prefixed hex data string) as an int64, reporting false if the word is
+// out of range or does not fit in an int64.
+func readHexWordAt(data string, hexStart int) (int64, bool) {
+	if hexStart < 2 || hexStart+evmWordHex > len(data) {
+		return 0, false
+	}
+	v, err := strconv.ParseInt(data[hexStart:hexStart+evmWordHex], 16, 64)
+	if err != nil {
+		return 0, false
+	}
+	return v, true
+}
+
+// parseEnsNameFromLogData extracts the ABI-encoded `string name` (the first
+// non-indexed parameter of every NameRegistered variant) from a log's data. The
+// leading word holds the byte offset to the name's length word; the name bytes
+// follow. Reading via the offset - rather than a hardcoded position - keeps this
+// layout-agnostic, so the extra baseCost/premium word of the newer event is
+// handled without special-casing. Every index derives from attacker-controlled
+// input, so all bounds are checked and no slice expression can panic.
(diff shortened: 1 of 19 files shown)


pragmaxim:
### Alternative to a full reindex: alias-only backfill

The "Operational notes" say a full reindex is required to purge/refresh aliases. That's overkill — address aliases are a **pure projection of `NameRegistered` logs** isolated in their own column family (`cfAddressAliases`, written only by `storeAddressAliasRecords` ← `getEnsRecord` ← `eth_getLogs`). Nothing else depends on it and it depends on nothing but logs, so we can rebuild just that one CF instead of replaying every block/tx/trace.

There are really two problems the reindex solves, and a backfill covers both:
1. **Purge** spoofed aliases already written by the old accept-any code (I confirmed an active look-alike emitter `0xeeb9b6bf5fb68fb726005f7ba549c2f4b32f2dad` writing spam names like `ransomware`/`infostealer` to a single owner on mainnet — currently in the DB as aliases).
2. **Backfill** aliases missing because the old code matched only the 5-arg event — i.e. essentially all modern registrations via the 6-/7-arg controllers.

**one-shot `-rebuildensaliases` maintenance flag** (modeled on `-fixutxo`/`-rollback`: runs after the backend connects, does its work, exits):

1. `DeleteRangeCF(cfAddressAliases, …)` to wipe the CF, and clear the `cachedAddressAliasRecords` LRU. (`DeleteRangeCF` is already used in the v5→v6 / v6→v7 migrations.)
2. For each trusted registrar, `eth_getLogs` from its deployment block to head, **filtered by `address` + the 3 `NameRegistered` topics**, in the node's max range chunks (dev Erigon caps `getLogs` at 1000 blocks).
3. Feed each log through the existing `getEnsRecord(l, ensRegistrars)` → `storeAddressAliasRecords` — no second parser to keep in sync.

**Cost:** the node serves these straight from its log index and returns *only* ENS registrations (a few hundred k events across all of ENS history), so this is minutes-to-hours vs. days for a full ETH-archive reindex, and it touches a single CF.

Notes:
- Run it at startup (DB held exclusively, like `fixutxo`) to avoid alias-cache coherence issues.

Either way the "Operational notes" should point to this alias-only procedure rather than "full reindex required", so operators don't pay for a reindex they don't need.

cranycrane:
Thanks — agreed, and done. Added a one-shot `-rebuildensaliases` maintenance flag along the lines you sketched, so operators don't pay for a full reindex.

Runs at startup with the DB held exclusively, then exits (like `-fixutxo`):
1. Wipes `cfAddressAliases` via `DeleteRangeCF` and purges the `cachedAddressAliasRecords` LRU — a stale cached alias would otherwise outlive the wipe.
2. Rescans `NameRegistered` logs from the trusted registrars with chunked `eth_getLogs`, filtered server-side by `address` + `topics[0] ∈ {5-arg, 6-arg, 7-arg}` — one combined filter (`address[]` + `topic0[]` OR) rather than a per-registrar loop.
3. Feeds every returned log through the existing `getEnsRecord(l, ensRegistrars)` → batched `storeAddressAliasRecords`, so there's no second parser to keep in sync (and the emitter check is re-applied as defense-in-depth).

This purges the spoofed aliases (e.g. the `0xeeb9b6bf…` `ransomware`/`infostealer` spam you found) and backfills the 6-/7-arg registrations the 5-arg-only code dropped.

Deviations worth flagging:
- **Start block** defaults to 0 → backend tip rather than per-registrar deployment blocks — the config carries addresses, not deployment heights, and a filtered `getLogs` seeks straight to matching logs so pre-deployment ranges return empty fast. `-blockheight`/`-blockuntil` bound it.
- **Chunk size** is `-ensrebuildchunk` (default 100000); drop to 1000 for dev Erigon / capped nodes.
- **Trust-none chains** (BSC): wiped, nothing re-recorded — exactly the purge you'd want for the legacy Space ID `.bnb` spam.
- **Layering:** CF wipe + LRU purge in `db`, log scan on `EthereumRPC` (kept out of the `EthereumLikeParser` interface), wired by `db.RocksDB.RebuildEnsAliases` behind a small `ensAliasRebuilder` interface so embedders like BSC qualify.

Unit tests cover chunk stepping, the emitter filter (trusted / untrusted-dropped / wildcard), the trust-none no-op, and interruption, plus a `db` test that the wipe clears both the CF and the cache. Updated the Operational notes accordingly.

One thing I couldn't self-verify: an end-to-end run against a live archive node, so the full-history wall-clock is still your estimate — if you can point me at a dev node I'll do a real run.

db/rocksdb_ethereumtype.go
```diff
@@ -1953,3 +1953,46 @@ func (d *RocksDB) periodicStoreAddrContractsCache() {
...
+func (d *RocksDB) RebuildEnsAliases(chain bchain.BlockChain, fromBlock, toBlock uint32, chunkBlocks int, stop chan os.Signal) error {
+	if !d.chainParser.UseAddressAliases() {
+		return errors.New("RebuildEnsAliases: address_aliases is disabled for this coin")
+	}
+	rebuilder, ok := chain.(ensAliasRebuilder)
+	if !ok {
+		return errors.New("RebuildEnsAliases: coin does not support ENS alias rebuild")
+	}
+	glog.Info("RebuildEnsAliases: wiping existing address aliases")
+	if err := d.deleteAllAddressAliases(); err != nil {
```
pragmaxim: **🔴 Data loss: the alias CF is wiped before checking there is anything to rebuild.**

`db.RebuildEnsAliases` gates only on `UseAddressAliases()`, then calls `deleteAllAddressAliases()` here (DeleteRangeCF + cache purge). Only *afterward* does the eth-layer `RebuildEnsAliases` early-return `nil` when `len(ensRegistrars)==0` (`ethrpc.go` "nothing to backfill").

So on any EthereumType coin with `address_aliases: true` **and no `ens_registrars`** — currently `base_archive`, `optimism_archive`, `bsc_archive`, `avalanche_archive`, `arbitrum_archive` — running `-rebuildensaliases` **deletes every alias and re-derives none**. BSC is the sharpest case: `bscrpc.go` explicitly keeps the `.bnb` suffix to preserve legacy records, which this wipes. Because accept-any recording was removed, not even a full reindex brings them back → irrecoverable.

*Fix:* check for an empty trusted set **before** the wipe. The db layer can't see `ensRegistrars`, so add e.g. `HasEnsRegistrars() bool` to the `ensAliasRebuilder` interface and skip the wipe (return early) when it's false.

db/rocksdb_ethereumtype.go
```diff
@@ -1953,3 +1953,46 @@ func (d *RocksDB) periodicStoreAddrContractsCache() {
...
+			return false
+		}
+		select {
+		case <-stop:
+			return true
+		default:
+			return false
+		}
+	}
+	err := rebuilder.RebuildEnsAliases(fromBlock, toBlock, chunkBlocks, interrupted, d.storeAddressAliasRecordsBatch)
```
pragmaxim: **🟠 Wipe-first rebuild is non-atomic — a transient error or interrupt degrades live alias data.**

The CF is emptied up front (line 1979), then this call repopulates it chunk-by-chunk. A single `eth_getLogs` failure (RPC timeout, node hiccup) aborts with the CF already wiped and only pre-failure chunks rewritten, so address lookups serve **no** aliases until a full successful re-run. The default chunk (100 000 blocks) reuses `b.Timeout`, so a wide chunk on mainnet can hit that ceiling — exactly this trigger.

Related: an interrupt exits `exitCodeOK` (`performEnsAliasRebuild` maps `ErrOperationInterrupted`→OK in `blockbook.go`), so a half-finished rebuild looks like a clean success.

*Fix:* build into a shadow CF / temp keyspace and atomically swap, or upsert-then-delete-stale instead of wipe-first; at minimum, exit non-zero (loud warning) on interrupt so operators know the aliases are incomplete.

bchain/coins/eth/dataparser.go
```diff
@@ -303,33 +303,92 @@ func (p *EthereumParser) ParseInputData(signatures *[]bchain.FourByteSignature,
...
+}
+
+// getEnsRecord parses an ENS alias from a NameRegistered log (5-, 6- and 7-arg
+// events). Only emitters in registrars are accepted; empty records nothing, "*"
+// accepts any.
+func getEnsRecord(l *rpcLogWithTxHash, registrars map[string]struct{}) *bchain.AddressAliasRecord {
+	if len(l.Topics) != 3 || !isTrustedNameRegisteredEvent(l.Topics[0]) {
+		return nil
+	}
+	if _, acceptAny := registrars[ensRegistrarWildcard]; !acceptAny {
```
pragmaxim, line 352: **🟠 Trust-none silently turns off alias recording during normal sync on non-ethereum EVM coins.**

This emitter check flips behavior from accept-any to trust-none for **all** EthereumType coins, but only `ethereum`/`ethereum_archive` got `ens_registrars`. On `base_archive`, `optimism_archive`, `bsc_archive`, `avalanche_archive`, `arbitrum_archive` (all `address_aliases: true`), sync now records **no** aliases, with no warning.

This is partly by-design, but it's material for **Base (Basenames)** and **Optimism**, which have real, popular ENS-style name services that accept-any *was* indexing. Please decide per-coin: populate their registrar addresses, or consciously accept dropping them — and consider a one-line startup log when `address_aliases && empty ens_registrars` so the "feature is off" state is visible.

bchain/coins/eth/ethrpc.go
```diff
@@ -1421,17 +1434,121 @@ func (b *EthereumRPC) processEventsForBlock(blockNumber string) (map[string][]*b
...
+		glog.Info("RebuildEnsAliases: ens_registrars is empty (trust none); nothing to backfill")
+		return nil
+	}
+	if chunkBlocks <= 0 {
+		chunkBlocks = ensRebuildDefaultChunkBlocks
+	}
+	// topic0 filter: the NameRegistered signatures getEnsRecord can parse (OR).
+	filter := map[string]interface{}{
+		"topics": []interface{}{[]string{
+			nameRegisteredEventSignature,
```
pragmaxim: **🟡 Duplicated NameRegistered signature list — drift will silently break the backfill or live sync.**

These three topic0 signatures are hard-coded here in the getLogs filter and again in `isTrustedNameRegisteredEvent` (`dataparser.go`). When a future controller version is added to one site but not the other, the two diverge: add it only to `isTrustedNameRegisteredEvent` and the backfill's `eth_getLogs` never fetches those logs (silently misses every alias from the new controller); add it only to the filter and live `getEnsRecord` drops them. A single shared `[]string` used by both makes drift impossible.

blockbook.go
```diff
@@ -526,6 +539,30 @@ func performRollback() error {
...
+	bestHeight, err := chain.GetBestBlockHeight()
+	if err != nil {
+		return errors.Annotatef(err, "GetBestBlockHeight")
+	}
+	from := uint32(0)
+	if *blockFrom >= 0 {
+		from = uint32(*blockFrom)
+	}
+	to := bestHeight
+	if *blockUntil >= 0 && uint32(*blockUntil) < to {
```
pragmaxim, line 602: **🟡 Unchecked `int`→`uint32` truncation of `-blockheight`/`-blockuntil`.**

`uint32(*blockUntil)` (and `uint32(*blockFrom)` above) wraps silently for values above 2³²-1, scanning a wrong (small) range — or tripping the `from > to` guard — with no error surfaced. Low likelihood at current chain heights, but a cheap explicit bounds-check would avoid a silent misconfiguration.

cranycrane:
This is postponed as indexing and keeping the current state for address->ENS resolution is complex and actually not required now, also the current ENS code is deprecated and not working as expected, therefore this #1695  is a chosen follow-up

cranycrane:
not intended anymore as described in https://github.com/trezor/blockbook/issues/1783

PR #1653, opened 2026-07-23
fix(internal-server): protect against CSRF

cranycrane:
## What

Adds CSRF protection to the internal server's state-changing `/admin` endpoints.

The `/admin` surface is gated by HTTP Basic auth, which authenticates the operator but does not defend against CSRF: browsers replay cached Basic credentials on cross-origin requests, so a malicious page an authenticated admin visits could drive forged `POST`/`PUT`/`DELETE` requests against the internal port (runtime-settings changes, contract-info forgery, anti-abuse state resets).

## How

A same-origin check in `requireAdminAuth`, so it covers every admin endpoint (JSON and form) in one place:

- State-changing methods (`POST`/`PUT`/`DELETE`/`PATCH`) are rejected with `403` when the `Origin` header — or, absent that, `Referer` — names a host different from the request's `Host`. Opaque (`null`) and unparsable origins never match.
- Safe methods (`GET`/`HEAD`/`OPTIONS`) are exempt.
- Requests with neither `Origin` nor `Referer` are allowed, so scripted clients (`curl`, automation) keep working — a browser cannot be made to omit `Origin` on a cross-origin mutation, so this does not reopen the hole.

## Tests

`TestRequireAdminAuthCSRF` covers safe-method exemption, same-origin pass, cross-origin/`null`/unparsable rejection, Origin-over-Referer precedence, the Referer fallback, and the header-less automation case.

### DIFF
--- a/server/internal.go
+++ b/server/internal.go
@@ -9,6 +9,7 @@ import (
 	"html/template"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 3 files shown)


pragmaxim (approved):
LGTM

PR #1654, opened 2026-07-23
Fix/increase get block default page size

cranycrane:
This increases the default `getBlock` tx size per page to fix broken Trezor-Suite clients.


pragmaxim (approved):
LGTM

PR #1548, opened 2026-06-07
fix(zcash): migrate testnet backend to Zebra 5.0.0 for NU6.2

sneurlax:
NU6.2 (consensus branch id 0x5437f330, testnet activation height 4052000) needs a NU6.2-aware node so Blockbook forwards the live consensus.chaintip instead of a stale value. zcashd reached end of support on 2025-05-31 and is being replaced by zebrad/zallet; the only NU6.2-capable zcashd build (6.20.0) ships no debian-bullseye tarball and publishes no sha256, so it cannot pass the existing sha256 build verification unattended. Migrate testnet to Zebra 5.0.0, mirroring the mainnet Zebra backend.

Mirror the mainnet docker backend (zfnd/zebra:5.0.0, same registry-verified repo digest) with testnet values via a new zcash_testnet.conf that sets network = "Testnet" and the testnet p2p listen_addr 0.0.0.0:18233. Zebra has no ZeroMQ, so drop backend_message_queue and message_queue_binding_template; the empty binding makes Blockbook fall back to polling. Also drop the zcashd-only zcash-fetch-params postinst and addnode params, and switch the service to simple. Reindent to 4 spaces to match configs/coins/zcash.json.

See:
https://zfnd.org/zebra-4-5-3-and-5-0-0-emergency-soft-fork-and-nu6-2-activation/ https://github.com/ZcashFoundation/zebra/blob/616fa36ecc06e730f1e2accc9a45bca4eb08b561/CHANGELOG.md?plain=1#L22 https://github.com/ZcashFoundation/zebra/blob/616fa36ecc06e730f1e2accc9a45bca4eb08b561/zebra-chain/src/parameters/network_upgrade.rs#L234 https://github.com/zcash/zcash/blob/v6.20.0/src/chainparams.cpp#L536

PR #1655, opened 2026-07-23
Update Zcash zebrad to 6.0.0 for NU6.3's new branch_id

sneurlax:
This is a follow-on of https://github.com/trezor/blockbook/pull/1548, which updated zebrad for NU6.2 and testnet: this updates zebrad for NU6.3, which is supported by zebrad 6.0.0.

The NU6.3 upgrade will occur around July 28, in about 5 days.  This should be merged before or around that time :) See the related https://github.com/trezor/trezor-suite/pull/30239


cranycrane:
Hey, thank you! We have already updated to 6.2.2 in https://github.com/trezor/blockbook/pull/1659

PR #1656, opened 2026-07-24
feat(eth): poll alternative fee providers eagerly (10s) with a configurable stale window

pragmaxim:
## Summary

- Reduce the EVM 1inch/Infura EIP-1559 fee refresh cadence from **60s → 10s** for every EVM provider config (arbitrum/base/optimism/polygon/ethereum archive + avalanche[_archive]) so fee estimates track the chain more closely.
- Make the cache staleness tolerance a **first-class, absolute config value** (`staleSeconds`) instead of a poll-period multiplier. It defaults to **600s (10 minutes)** and is independent of the poll cadence, via a shared `defaultFeeStaleSeconds` / `feeStaleDuration` helper living in the shared `alternativefeeprovider.go`.

## Why

The previous design coupled the stale window to the poll cadence (`window = periodSeconds × stalePeriods`), so changing one silently moved the other. Decoupling means:
- the 10s-poll coins keep a clean 10-minute window;
- `bsc_archive` (still 60s poll, intentionally not switched to 10s) now also gets the same 10-minute window via the default, instead of the accidental 60 minutes the old shared multiplier would have produced.

## Notes

- 10s polling is ~6× the request volume against 1inch/Infura; the API plan must tolerate it. Fees still fall back to node estimation past the stale window.
- Data lifecycle is unchanged: fetched by a background goroutine into an in-memory cache (not persisted to the DB); consumers read the cache.
- `docs/config.md` updated to document `staleSeconds`; fee-provider tests cover the default and an explicit override.

### DIFF
--- a/bchain/coins/eth/alternativefeeprovider.go
+++ b/bchain/coins/eth/alternativefeeprovider.go
@@ -1,12 +1,32 @@
 package eth
 
 import (
+	"net/http"
 	"sync"
 	"time"
 
 	"github.com/trezor/blockbook/bchain"
 	"github.com/trezor/blockbook/common"
 )
+
+// feeProviderParams is the shared config for the EVM alternative fee providers
+// (Infura and 1inch). The two were field-for-field identical and had already
+// drifted once, so they share one struct to keep parsing and validation in sync.
+type feeProviderParams struct {
+	URL           string `json:"url"`
+	PeriodSeconds int    `json:"periodSeconds"`
+	// StaleSeconds is how long a cached estimate stays usable after the last
+	// successful refresh before it is considered stale (falls back to node
+	// estimation). It is independent of the poll cadence. Optional; defaults to
+	// defaultFeeStaleSeconds when unset.
+	StaleSeconds int `json:"staleSeconds"`
+}
+
+// feeHTTPClient is shared by the fee pollers. The timeout bounds each poll well
+// under the poll cadence so a blackholed connection (accepted, headers never
+// sent) can't block the single poller goroutine forever and silently freeze
+// fee refreshes; http.DefaultClient has no timeout.
+var feeHTTPClient = &http.Client{Timeout: 5 * time.Second}
 
 type alternativeFeeProvider struct {
 	eip1559Fees       *bchain.Eip1559Fees
@@ -42,10 +62,37 @@ type alternativeFeeProviderInterface interface {
 	GetEip1559Fees() (*bchain.Eip1559Fees, error)
 }
 
+// defaultFeeStaleSeconds is the cached-estimate stale window (10 minutes) used
+// by the EVM alternative fee providers when the config omits "staleSeconds".
+const defaultFeeStaleSeconds = 600
+
+// feeStaleDuration returns the stale window for cached estimates: the configured
+// staleSeconds (falling back to the 10-minute default when unset or non-positive),
+// clamped up to periodSeconds. The clamp guarantees the window can never be
+// shorter than the poll cadence — otherwise a config like {periodSeconds:60,
+// staleSeconds:30} would flap provider→node once per poll on a healthy provider,
+// a trap the old periodSeconds*30 formula made structurally impossible.
+func feeStaleDuration(periodSeconds, staleSeconds int) time.Duration {
+	if staleSeconds <= 0 {
+		staleSeconds = defaultFeeStaleSeconds
+	}
+	if staleSeconds < periodSeconds {
+		staleSeconds = periodSeconds
+	}
+	return time.Duration(staleSeconds) * time.Second
+}
+
 func (p *alternativeFeeProvider) GetEip1559Fees() (*bchain.Eip1559Fees, error) {
 	p.mux.Lock()
 	defer p.mux.Unlock()
-	if p.lastSync.Add(p.staleSyncDuration).Before(time.Now()) {
+	// Treat an unset staleSyncDuration as the default rather than 0, so a future
+	// provider that forgets to initialize it degrades to the 10-minute window
+	// instead of a silent no-op (lastSync.Add(0) is always in the past).
+	stale := p.staleSyncDuration
+	if stale <= 0 {
+		stale = defaultFeeStaleSeconds * time.Second
+	}
+	if p.lastSync.Add(stale).Before(time.Now()) {
 		return nil, nil
 	}
 	return p.eip1559Fees, nil
--- a/bchain/coins/eth/infurafees.go
+++ b/bchain/coins/eth/infurafees.go
@@ -75,18 +75,11 @@ type infuraFeesResult struct {
 	BaseFeeTrend               string          `json:"baseFeeTrend"`
 }
 
-type infuraFeeParams struct {
-	URL           string `json:"url"`
-	PeriodSeconds int    `json:"periodSeconds"`
-}
-
 type infuraFeeProvider struct {
 	*alternativeFeeProvider
-	params infuraFeeParams
+	params feeProviderParams
 	apiKey string
 }
-
-const infuraFeeStalePeriods = 30
 
 // NewInfuraFeesProvider initializes https://gas.api.infura.io provider
 func NewInfuraFeesProvider(chain bchain.BlockChain, params string, metrics *common.Metrics) (alternativeFeeProviderInterface, error) {
@@ -104,15 +97,11 @@ func NewInfuraFeesProvider(chain bchain.BlockChain, params string, metrics *comm
 	}
 	p.params.URL = strings.Replace(p.params.URL, "${api_key}", p.apiKey, -1)
 	p.chain = chain
-	// Keep cached Infura fees through throttling bursts.
-	// Current archive configs poll every 60s, which gives a 30-minute window.
-	p.staleSyncDuration = infuraFeeStaleDuration(p.params.PeriodSeconds)
+	// Keep cached Infura fees through throttling bursts for the configured stale
+	// window (defaults to 10 minutes), independent of the poll cadence.
+	p.staleSyncDuration = feeStaleDuration(p.params.PeriodSeconds, p.params.StaleSeconds)
 	go p.FeeDownloader()
 	return p, nil
-}
-
-func infuraFeeStaleDuration(periodSeconds int) time.Duration {
-	return time.Duration(periodSeconds*infuraFeeStalePeriods) * time.Second
 }
 
 func (p *infuraFeeProvider) FeeDownloader() {
@@ -185,7 +174,7 @@ func (p *infuraFeeProvider) getData(res interface{}) error {
 		return err
 	}
 	httpReq.Header.Set("Content-Type", "application/json")
-	httpRes, err := http.DefaultClient.Do(httpReq)
+	httpRes, err := feeHTTPClient.Do(httpReq)
 	if httpRes != nil {
 		defer httpRes.Body.Close()
 	}
--- a/bchain/coins/eth/infurafees_test.go
+++ b/bchain/coins/eth/infurafees_test.go
@@ -5,18 +5,34 @@ import (
 	"time"
 )
 
-func TestInfuraFeeStaleDuration(t *testing.T) {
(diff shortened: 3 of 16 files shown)


configs/coins/ethereum_archive.json
```diff
@@ -70,7 +70,7 @@
                 "address_aliases": true,
                 "eip1559Fees": true,
                 "alternative_estimate_fee": "1inch",
-                "alternative_estimate_fee_params": "{\"url\": \"https://api.1inch.dev/gas-price/v1.5/1\", \"periodSeconds\": 60}",
+                "alternative_estimate_fee_params": "{\"url\": \"https://api.1inch.dev/gas-price/v1.5/1\", \"periodSeconds\": 10}",
```
cranycrane, line 73: **[ops / plausible]** Dropping `periodSeconds` 60→10 multiplies provider API usage 6× (~8,640 req/day per config). The six Infura configs share one `INFURA_API_KEY` (~51,840 Gas API calls/day against that key's credit quota), and this 1inch config alone makes ~259k calls/month. `FeeDownloader` has no backoff on 429, polls unconditionally even when nothing queries fees, and `lastSync` only advances on success — so once a quota is exhausted, every chain on the key silently falls back to node estimation after the 600 s stale window for the rest of the quota period, the opposite of the freshness this PR targets. Worth confirming the account tiers absorb 6× before merging, and/or adding backoff on repeated failures.

bchain/coins/eth/infurafees.go
```diff
@@ -104,17 +107,13 @@ func NewInfuraFeesProvider(chain bchain.BlockChain, params string, metrics *comm
 	}
 	p.params.URL = strings.Replace(p.params.URL, "${api_key}", p.apiKey, -1)
 	p.chain = chain
-	// Keep cached Infura fees through throttling bursts.
-	// Current archive configs poll every 60s, which gives a 30-minute window.
-	p.staleSyncDuration = infuraFeeStaleDuration(p.params.PeriodSeconds)
+	// Keep cached Infura fees through throttling bursts for the configured stale
+	// window (defaults to 10 minutes), independent of the poll cadence.
+	p.staleSyncDuration = feeStaleDuration(p.params.StaleSeconds)
 	go p.FeeDownloader()
```
cranycrane, line 103: **[correctness / pre-existing, made worse]** `getData` uses `http.DefaultClient` — no timeout, no request context (same in `oneinchfees.go`) — and this goroutine is the only poller. A blackholed connection (accepted, headers never sent; `DefaultTransport` has no `ResponseHeaderTimeout`) blocks `Do()` forever: polling halts permanently and fees silently revert to node estimation once the stale window lapses, with no watchdog or retry. This PR shrinks that window from 30 min to 10 min for existing deployments while 6× polling raises the odds of ever hitting such a connection. The repo already has the pattern at `ethrpc.go:638` (`http.Client{Timeout: 2 * time.Second}`); a ~5 s timeout here bounds each poll well under the 10 s period.

bchain/coins/eth/alternativefeeprovider.go
```diff
@@ -42,6 +42,19 @@ type alternativeFeeProviderInterface interface {
...
 }
 
+// defaultFeeStaleSeconds is the cached-estimate stale window (10 minutes) used
+// by the EVM alternative fee providers when the config omits "staleSeconds".
+const defaultFeeStaleSeconds = 600
+
+// feeStaleDuration returns the configured stale window, falling back to the
+// 10-minute default when staleSeconds is unset or non-positive.
+func feeStaleDuration(staleSeconds int) time.Duration {
+	if staleSeconds <= 0 {
```
cranycrane, line 76: **[correctness]** Nothing validates or clamps `staleSeconds >= periodSeconds` — a configuration the removed `periodSeconds*30` formula made structurally impossible. With e.g. `{"periodSeconds": 60, "staleSeconds": 30}` on a perfectly healthy provider, cached fees serve for 30 s after each refresh, then `GetEip1559Fees` returns `(nil, nil)` for the remaining ~30 s of every cycle — the fee source flaps provider→node once per poll. docs/config.md advertises the knob as "independent of the poll cadence" with no floor warning. A clamp (`max(staleSeconds, periodSeconds)`) or a constructor error would close the trap.

bchain/coins/eth/oneinchfees.go
```diff
@@ -81,7 +80,7 @@ func NewOneInchFeesProvider(chain bchain.BlockChain, params string, metrics *com
 		return nil, errors.New("NewOneInchFeesProvider: missing ONE_INCH_API_KEY env variable.")
 	}
 	p.chain = chain
-	p.staleSyncDuration = oneInchFeeStaleDuration(p.params.PeriodSeconds)
+	p.staleSyncDuration = feeStaleDuration(p.params.StaleSeconds)
```
cranycrane: **[correctness / pre-existing]** L75 above validates `p.params.PeriodSeconds == 0`, while the Infura constructor rejects `<= 0` (`infurafees.go:101`) — so a negative `periodSeconds` passes only this provider's validation. `time.NewTimer`/`timer.Reset` with a negative duration fire immediately, turning `FeeDownloader` into a zero-delay busy loop against api.1inch.dev, paced only by HTTP round-trip time. One-character fix: `== 0` → `<= 0`, which also aligns the two providers' fail-fast behavior.

docs/config.md
```diff
@@ -103,8 +103,8 @@ Good examples of coin configuration are
           * Tron-specific endpoint configuration is documented in [Tron Config](/docs/tron-config.md).
           * Infura alternative EIP-1559 fee provider configuration:
             * `alternative_estimate_fee` – Set to `infura` to use Infura Gas API fee suggestions instead of native node fee estimation.
```
cranycrane: **[docs]** The line below was generalized to "polls the provider", but this section still documents only `infura` as a valid `alternative_estimate_fee` value. The `1inch` provider (shipped in `ethereum_archive.json` and governed by these same params, including the new `staleSeconds`) and its hard `ONE_INCH_API_KEY` requirement (`NewOneInchFeesProvider` fails startup without it) are documented nowhere operator-facing — `docs/env.md` lists only `INFURA_API_KEY`. Since exactly these lines were edited, worth completing the generalization.

bchain/coins/eth/alternativefeeprovider.go
```diff
@@ -42,6 +42,19 @@ type alternativeFeeProviderInterface interface {
 	GetEip1559Fees() (*bchain.Eip1559Fees, error)
 }
 
+// defaultFeeStaleSeconds is the cached-estimate stale window (10 minutes) used
+// by the EVM alternative fee providers when the config omits "staleSeconds".
+const defaultFeeStaleSeconds = 600
+
+// feeStaleDuration returns the configured stale window, falling back to the
+// 10-minute default when staleSeconds is unset or non-positive.
+func feeStaleDuration(staleSeconds int) time.Duration {
```
cranycrane: **[altitude]** Each constructor must remember `p.staleSyncDuration = feeStaleDuration(...)` — the base never defaults it. A future provider that forgets that one line compiles, starts, and polls forever, yet `GetEip1559Fees` returns `(nil, nil)` on every call (`lastSync.Add(0)` is always in the past): a silent no-op visible only as unexplained node fallback in production. Defaulting inside the base — treating a zero `staleSyncDuration` as `defaultFeeStaleSeconds` in the stale check, or setting it in a base initializer — would make it unforgettable.

bchain/coins/eth/oneinchfees.go
```diff
@@ -52,6 +52,11 @@ type oneInchFeeFeesResult struct {
 type oneInchFeeParams struct {
 	URL           string `json:"url"`
 	PeriodSeconds int    `json:"periodSeconds"`
+	// StaleSeconds is how long a cached estimate stays usable after the last
```
cranycrane: **[cleanup]** This field + 4-line comment is copy-pasted verbatim into `infuraFeeParams` (`infurafees.go:81-85`), and the two param structs are now field-for-field identical (`url`, `periodSeconds`, `staleSeconds`). A single shared params struct next to `feeStaleDuration` would remove the duplication — the pair has already drifted once (`<= 0` vs `== 0` period validation), and nothing enforces parity for the next divergence.

bchain/coins/eth/infurafees.go
```diff
@@ -78,6 +78,11 @@ type infuraFeesResult struct {
 type infuraFeeParams struct {
 	URL           string `json:"url"`
 	PeriodSeconds int    `json:"periodSeconds"`
```
cranycrane: **[altitude]** The 10 s cadence is hardcoded into 7 coin JSONs while `PeriodSeconds` gets no code default — the exact pattern this PR just introduced for `StaleSeconds`. Every future EVM config must remember `"periodSeconds": 10` or silently get whatever its author copies from (`bsc_archive.json:72` already sits at 60 s behind `"infura-disabled"`; whoever re-enables it resurrects the old cadence unnoticed). A code default with config override would make the cadence one constant, and would also unify the three current behaviors for non-positive values (infura errors on `<= 0`, 1inch only on `== 0`, `staleSeconds` silently defaults).

bchain/coins/eth/infurafees_test.go
```diff
@@ -37,7 +45,7 @@ func TestInfuraFeeProviderUsesCachedFeesDuringStaleWindow(t *testing.T) {
 	})
 
 	provider.mux.Lock()
-	provider.lastSync = time.Now().Add(-29 * time.Minute)
+	provider.lastSync = time.Now().Add(-9 * time.Minute)
```
cranycrane, line 56: **[test]** `-9 * time.Minute` here and `-11` below are only valid because `staleSyncDuration` was set from `feeStaleDuration(0)` = the 10-minute default — the constant is encoded twice, once indirectly. Deriving the offsets from the duration itself (`staleSyncDuration ∓ time.Minute`) keeps the test correct if `defaultFeeStaleSeconds` ever changes.

bchain/coins/eth/infurafees_test.go
```diff
@@ -5,18 +5,26 @@ import (
 	"time"
 )
 
-func TestInfuraFeeStaleDuration(t *testing.T) {
-	got := infuraFeeStaleDuration(60)
-	want := 30 * time.Minute
+func TestFeeStaleDurationDefaultsToTenMinutes(t *testing.T) {
```
cranycrane, line 8: **[test]** These two tests exercise `feeStaleDuration`, which lives in `alternativefeeprovider.go` — and `alternativefeeprovider_test.go` exists and is touched by this same PR. Moving them there keeps the coverage next to the code it covers; a maintainer changing the helper won't find these in the infura-named file.

cranycrane (commented):
Code review of 845f72d9 (multi-angle scan, each finding adversarially verified, then a gap sweep). 11 findings survived verification — 10 as inline comments below, plus this one that can't be attached inline because the files aren't part of the diff:

**[ops] Monitoring & docs still encode the removed 30×-period stale window**
- `configs/grafana/panels.yaml:735` — cache-age panel cutoff is `vector(1800)` with legend `default stale cutoff (30x 60s = 30m)` (description at `:729` repeats the 30x rule)
- `configs/metrics.yaml:384` — help text of `blockbook_alternative_fee_provider_last_sync_timestamp_seconds` still says "30x the poll period, e.g. 30m at a 60s poll"
- `docs/env.md:76` — still says Infura configs "poll once per minute and keep serving the last successful fee data for up to 30 failed polls"

The real threshold after this PR is 600 s (or configured `staleSeconds`). During a provider wedge Blockbook switches to `onchain_fallback` at cache age 601 s while the dashboard shows the age line comfortably under the 1800 s dashed cutoff — operators would read the provider as healthy for ~20 extra minutes during exactly the incident the panel exists for.

Also verified and dismissed, for the record: `bsc_archive.json`'s leftover 60 s is dead config (`"infura-disabled"` never initializes a provider); `staleSeconds: 0` silently taking the default is documented+tested intent; the params JSON ignoring unknown keys matches repo-wide config practice.

cranycrane (approved):
LGTM

PR #1677, opened 2026-07-29
fix(eth): treat balanceOf reverts as benign in single-call path

cranycrane:
## Summary

The ERC20 balance batch path already classified deterministic `balanceOf` reverts (`execution reverted`, `division or modulo by zero`, `out of gas`, `invalid opcode`) as benign via `isNonRetriableEthCallError` and left them nil without warning. The standalone single-call path did not — it returned the raw error, which `worker.go` then logged at **warning** level, flooding logs for holders of dead/non-conforming/rebasing tokens (e.g. `EthereumTypeGetErc20ContractBalance ... execution reverted: division or modulo by zero`).

This change makes the single-call path consistent: deterministic reverts are classified as `ErrInvalidErc20Balance` and logged at `V(2)`. Genuine RPC/transport errors still surface unchanged and remain visible at warning level.

The balance result is unaffected in all cases (these tokens already yielded no balance) — only the log level changes.

## Changes
- `EthereumTypeGetErc20ContractBalanceAtBlock` runs RPC errors through `isNonRetriableEthCallError`; reverts return `ErrInvalidErc20Balance` and increment a new `observeEthCallError("single", "reverted")` metric.
- Broadened the `ErrInvalidErc20Balance` doc comment.
- Added tests for both the benign-revert and genuine-RPC-error cases.

### DIFF
--- a/bchain/coins/eth/contract.go
+++ b/bchain/coins/eth/contract.go
@@ -495,11 +495,10 @@ func (b *EthereumRPC) GetContractInfo(contractDesc bchain.AddressDescriptor) (*b
 	return b.fetchContractInfo(address)
 }
 
-// ErrInvalidErc20Balance is returned when a balanceOf eth_call succeeds but returns data that
-// cannot be parsed as a 32-byte integer (empty "0x" or non-conforming output). It is benign and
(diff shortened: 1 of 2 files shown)


pragmaxim (approved):
LGTM

PR #1679, opened 2026-07-29
feat(server): make per-IP websocket connection limits configurable

cranycrane:
## What

The per-IP websocket connection cap (128) and connection-attempt cap (64/min) were hard-coded constants. This exposes them as network-prefixed env vars, following the existing `<net>_WS_*` limit pattern:

- `<net>_WS_MAX_CONNECTIONS_PER_IP` — max concurrently open connections per client key (default `128`, `0` disables)
- `<net>_WS_MAX_CONNECTION_ATTEMPTS_PER_IP` — max new handshakes per client key within the trailing 1-minute window (default `64`, `0` disables)

## Notes

- The limiter now carries the configured caps and skips a check when its cap is `0`, so `0` cleanly disables rather than rejecting every connection.
- Defaults are unchanged, so existing deployments behave identically unless the new vars are set.
- Docs added in `docs/env.md`; a new `TestWebsocketConnectionLimiterDisabled` covers the `0`-disables path. All limiter tests pass.


pragmaxim (approved):
LGTM

PR #1682, opened 2026-07-29
fix(tron): count internal transaction inflows in balance history

### LINKED ISSUE #1660: Tron balance history goes negative: internal-transaction inflows never counted (processInternalTransactions off for Tron)
## Summary

For Tron, `getBalanceHistory` accounts for only **one direction** of smart-contract TRX flows. TRX sent **into** a contract (the top-level call value) is booked as `sent`, but TRX that comes **back** from a contract (unwrapping WTRX, removing liquidity, swap payouts, refunds) arrives via **internal transactions**, which are never counted as `received`. The result is a one-directional leak that drags the cumulative balance below zero for any DeFi-active address — Trezor Suite renders this as a balance graph that goes deeply **negative**, which is impossible for a spendable balance.

Root cause: `processInternalTransactions` is **not enabled** in `configs/coins/tron.json`, so `bchain.ProcessInternalTransactions` defaults to `false`. Internal data is therefore neither indexed at sync (`bchain/coins/tron/tronrpc.go:883`) nor read in balance-history accounting (`api/worker.go:1792` early-returns).

## Reproduction

- Address: `TT2T17KZhoDu47i2E4FWxfG79zdkEWkU9N`
- Endpoint: WS/REST `getBalanceHistory`
- Observed: cumulative `received - sent` bottoms at **≈ −4.96B TRX**; totals `received 10,591,741,179.20` / `sent 15,295,334,567.64` (net **−4.7B**).
- Reality (from `/api/v2/address/...`): the account is fully solvent — **289.5M liquid + 196.6M staked + 344.7M unstaking ≈ 830.7M TRX**. It is impossible to have net-sent 4.7B more than ever received; the `received` side is being undercounted.

The current `balance` field is correct (read from account state). Only the *reconstructed history* is wrong.

## Evidence — one transaction, dissected

Tx `d54a1abc1c21795c2969881d2a7fb4318c3e24df68879100682dd5f791f9a547` (block 83,534,426) — a SunSwap V3 add-liquidity call.

`walletsolidity/gettransactioninfobyid` shows two native-TRX internal transfers (all other entries have empty `callValueInfo`):

| internal tx | caller → to | native TRX |
|---|---|---|
| `600aa8…` | position mgr → WTRX contract | 177,460,640.803707 (our TRX wrapped into the pool) |
| `dd6b90…` | position mgr → **`TT2T17KZ…` (owner)** | **13,245.012561 ← refund back to the address** |

Arithmetic confirms the top-level call value is **not** duplicated in `internal_transactions` (so enabling processing does not double-count the send):

```
wrap 177,460,640.803707  +  refund 13,245.012561  =  177,473,885.816268  ==  top-level call value  ✅
```

- **Today (internal OFF):** `sent 177,473,885.82`, `received 0` → net **−177,473,885.82**.
- **Correct (internal ON):** `sent 177,473,885.82`, `received 13,245.01` → net **−177,460,640.80**, exactly the WTRX actually deposited.

This single tx is 100% of the `sent` in the `time: 1781269200` (2026-06-12 13:00 UTC) balance-history bucket. The multi-billion drift comes from the *mirror* transactions (closing positions / unwrapping WTRX / swap-to-TRX), where call value is `0` (booked `sent 0`) and the **entire principal** returns as a native-TRX internal transfer that currently books `received 0`.

## Verification of the existing parser

`buildInternalDataFromTronInfos` (`bchain/coins/tron/tronInternalDataProvider.go`) already extracts these transfers correctly (skips empty `callValueInfo` and TRC-10; `ToTronAddressFromAddress` handles both the `41…` hex form from `gettransactioninfobyblocknum` and base58 passthrough). `processInternalTransactionsForBalanceHistory` (`api/worker.go:1804-1824`) would then credit `received += 13,245.01`. So the accounting is correct once the flag is on.

## Secondary bug exposed by the same data — CREATE misclassification

`bchain/coins/tron/tronInternalDataProvider.go:157`:

```go
if topType == bchain.CALL && info.ContractAddress != "" {
    topType = bchain.CREATE
    contractAddr = ToTronAddressFromAddress(info.ContractAddress)
}
```

Tron populates `contract_address` for ordinary `TriggerSmartContract` **calls**, not just deployments (confirmed in the `d54a…` response, `contract_address = TLSWrv…`). So every contract *call* is reclassified as a **contract creation** and the called contract is registered as "created in this block" with `UnhandledTokenStandard`. This does **not** corrupt balance-history amounts (that path reads only `d.Transfers`), but it pollutes the contract registry / first-seen + token-standard metadata. It should be gated on a real creation signal (tx being `CreateSmartContract`, or a `create`/`suicide` internal note) rather than merely `contract_address != ""`.

⚠️ A full-history backfill would run this heuristic over every block, so **this must be fixed before backfilling**.

## Proposed fix — no full reindex required

1. Enable `processInternalTransactions: true` in `configs/coins/tron.json` (fixes all new blocks). Note: `gettransactioninfobyblocknum` is already fetched per block for fees/receipts and already contains `internal_transactions`, so there is **no extra sync-time RPC call** — the data is currently parsed and discarded.
2. Backfill history using the existing primitive rather than a reindex: `ReconnectInternalDataToBlockEthereumType` (`db/rocksdb_ethereumtype.go:743`) patches internal data into an already-synced block and, via `existingBlock=true` → `setAddressTxIndexesToAddressMap`, also inserts the address→tx index entries for internal-only counterparties. It's driven by `RefetchInternalDataRoutine` (`api/ethereumtype.go:50`), which drains `cfBlockInternalDataErrors`.
   - Since the flag was always off, that error table is empty, so a small **range-driver** is needed: loop heights → `GetBlock(h)` → `ReconnectInternalDataToBlockEthereumType(block)` (or seed `cfBlockInternalDataErrors` for the range and trigger `RefetchInternalData()`).
   - This re-fetches each block from the node (RPC load ≈ resync of the range) but writes only internal transfers + affected address indexes — no rebuild of tx data / token transfers / balances / UTXO indexes.

## Caveats to validate

- Fix the `:157` CREATE misclassification **first**.
- `ReconnectInternalDataToBlockEthereumType` was designed for a handful of transient-failure blocks, not full-history backfill. Confirm on a small range that re-adding an address already indexed via its top-level vin/vout does **not** double its tx count; diff `getBalanceHistory` + address tx count before/after, then run the full range.

## Affected code

- `configs/coins/tron.json` — `processInternalTransactions` absent → defaults false (`bchain/coins/eth/ethrpc.go:274`)
- `bchain/coins/tron/tronrpc.go:883` — internal data built only when flag on
- `api/worker.go:1792` — balance-history internal accounting gated off
- `api/worker_balance_history_tron.go` — Tron balance-history logic
- `bchain/coins/tron/tronInternalDataProvider.go:157` — CREATE misclassification
- `db/rocksdb_ethereumtype.go:743` / `api/ethereumtype.go:50` — existing backfill path

cranycrane:
Fixes #1660. **#1703 has landed** (merged 2026-08-12) and this branch is rebased on top of it.

Tron `getBalanceHistory` drifted billions of TRX negative on active addresses, because two independent classes of inflow were never counted as `received`.

**TRX returning from contracts** arrives via internal transactions, and `processInternalTransactions` was off — while TRX sent *into* a contract was booked as `sent`. That is the bulk of the drift on DeFi-active addresses.

**TRX returning from staking** was partly dropped too: Stake 1.0 unfreeze was booked as no movement although it credits the principal immediately, and java-tron's implicit sweep of matured unstake tranches was only read on an explicit withdraw. This second class was found while verifying the first against the deployed index.

## Internal transactions

1. **fix(tron): classify contract creations by missing recipient, not `contract_address`** — java-tron fills the tx-info `contract_address` for ordinary `TriggerSmartContract` calls too, so every contract call was registered as a contract creation. A creation is now recognized by the missing eth-style recipient combined with a reported `contract_address` (neither signal alone discriminates: native non-VM operations such as `FreezeBalance`/`WithdrawBalance` also have a null `to`, but never a `contract_address`). Had to land before anything re-processes history.
2. **feat(tron): enable `processInternalTransactions`** — zero extra sync-time RPC: `gettransactioninfobyblocknum` is already fetched per block for fees/receipts; the internal transfers were parsed and discarded. Fixes all newly synced blocks.
3. **chore(tron): enable `vm.saveInternalTx` in packaged backend config** — upstream's `main_net_config.conf` ships the flag off, under which java-tron records no internal transactions at all. The deployed backend already runs with it on (verified: it serves internal transactions for the whole chain history); this keeps a package rebuild from silently regressing that.

**Review fixes** (commits after the review rounds; the findings were verified against a live mainnet node and java-tron `GreatVoyage-v4.7.7` sources):

4. **fix(tron): skip rejected internal txs, register full contract lifecycle** — rejected frames (e.g. `OUT_OF_ENERGY`) no longer book phantom transfers; every non-rejected create/suicide note emits its registry event in execution order.
5. **fix(tron): emit zero-value transfers for create and suicide frames** — eth `processCallTrace` parity: without a transfer, a factory-deployed or ephemeral child's own address history never contains the tx that created/destroyed it, and the reorg rollback from #1703 has no row to key off. Plain zero-value calls stay skipped.
6. **fix(tron): skip featured frames, drop bogus error and selfdestruct inference** — featured internal transactions (`delegateResourceOfEnergy`, `freezeBalanceV2For*`, …) report the staked amount in `callValue` although no TRX moves; unknown notes are now skipped explicitly instead of booked as CALL transfers, and the packaged nile config turns `vm.saveFeaturedInternalTx` off to match mainnet's data shape. The per-frame `rejected` flag no longer stamps an error on successful transactions (java-tron rejects single frames inside txs that succeed), and the top-level SELFDESTRUCT inference is gone — it could not survive packing (one type bit; contract packed only for CREATE) and would have surfaced a destroyed contract as `createdContract`. `GetAddrDescFromAddress` also returns `ErrAddressMissing` for the empty address instead of falling through to a log line — empty is a normal shape for creation frames.
7. **fix(tron): do not store internal data rows for plain transactions** — eth parity: empty internal data no longer writes a `cfInternalData` row for every Tron transaction.
8. **fix(tron): do not register contracts from failed deployments** — java-tron precomputes and reports `contract_address` even when the deployment failed; the CREATE classification and the registry entry are now gated on the receipt result.
9. **chore(tron): bump backend `package_revision` for the config changes** — the extract_command of both packaged backends changed, but a rebuilt package kept the identical `-latest` version string and would never install as an upgrade; both configs now use the repo-wide `satoshilabs-1` revision, which also sorts above `latest` in dpkg comparison.
10. **fix(tron): gate the receipt contract fallback on deployment success** — the same precomputed `contract_address` also reached the fallback in `buildTxFromHTTPData`, which stamped CREATE on the stored internal data and indexed the never-created address although the registry (correctly) got no entry. Verified against GreatVoyage-v4.7.7: `VMActuator` sets the address on the `ProgramResult` before `VM.play` and commits state only on success, and `TransactionInfo` result is FAILED exactly when the commit was skipped, which maps to receipt status `0x0`.
11. **fix(tron): unify the VM success predicate, treat `0x` as missing recipient** — the CREATE classification, the error stamping and the receipt contract fallback now share one success predicate reading `receipt.result`, so classification and fallback can no longer disagree on a deployment's outcome. The classification also follows `EthTxToTx` recipient semantics (`""` and `"0x"` both missing) — a `"0x"` shape would have silently lost the registry entry, which only this path emits. The valued-transfer loop documents why it has no root-frame guard: the rebuilt deployment tx carries value 0, so skipping would lose the value.

## Staking

12. **fix(tron): credit staking inflows missing from balance history** — with internal transactions credited, `TT2T17KZ…` still failed `received >= sent` by exactly 600,000,000 TRX. Two further leaks in the operation-driven balance-history path:

    - **Stake 1.0 `UnfreezeBalanceContract` booked as no movement.** V1 unfreeze returns the principal to the spendable balance immediately — no unlock period, no `WithdrawExpireUnfreeze` — but operation `unfreeze` mapped to `DirectionNone` for V1 and V2 alike, while V1 *freeze* was still booked outgoing. Asymmetric, so it drifted permanently negative. Now branches on `ContractType`.
    - **The implicit expired-unfreeze sweep was dropped.** java-tron's `unfreezeExpire()` credits matured unstake tranches straight to the spendable balance and reports the total in `withdraw_expire_amount` — not only on `WithdrawExpireUnfreeze` but also on `UnfreezeBalanceV2` and `CancelAllUnfreezeV2`. Verified in the GreatVoyage sources: the sweep runs in `UnfreezeBalanceV2Actuator`, `CancelAllUnfreezeV2Actuator` and `WithdrawExpireUnfreezeActuator`, and in none of `FreezeBalanceV2Actuator`, `DelegateResourceActuator`, `UnDelegateResourceActuator`. Blockbook read the field only for the explicit withdraw, so every implicit sweep vanished.

    `withdraw_expire_amount` is now parsed for every contract type into a new `TronChainExtraData.WithdrawnUnfreeze` and credited independently of the operation's own principal movement, with `withdraw` and the newly mapped `cancelUnfreeze` switched to `DirectionNone` so nothing is counted twice. A separate field is unavoidable: `UnfreezeBalanceV2Contract` carries two distinct amounts at once — `v.UnfreezeBalance` (locks for 14 days, no spendable change) and `withdraw_expire_amount` (immediately spendable). It is net-zero on the struct, because it replaces the redundant `UnstakeAmount` write on `WithdrawExpireUnfreezeContract`, which held the same value and rendered a duplicate row on the tx page.

**API change:** `unstakeAmount` no longer appears on `WithdrawExpireUnfreezeContract` transactions — the amount moved to `withdrawnUnfreeze`. Trezor Suite is unaffected and needs no change: `TransactionHeader.tsx` and suite-native `TransactionName.tsx` read `unstakeAmount` only when the contract type is `UnfreezeBalanceContract`/`UnfreezeBalanceV2Contract`, and `trx` is in neither `LOCAL_BALANCE_HISTORY_COINS` nor `IGNORED_BALANCE_HISTORY_COINS`, so Suite consumes the server's `getBalanceHistory` numbers verbatim.

**Split out of this PR:** the contract registry consistency fixes moved to #1703, now merged. They repaired two pre-existing bugs in the shared eth-type DB layer (a same-block create→selfdestruct lost its destruction; reorgs left stale `cfContracts` rows that permanently disable `GetBalanceHistory`) that affect all eth-type chains, not just Tron — Tron simply becomes the second producer of registry events.

**Healing plan for existing indexes:** this PR makes newly synced blocks correct. The two halves heal differently:

- The **internal data** rows (`cfInternalData`) are index output, so already-indexed history needs a **full reindex** (the online backfill routine proposed in an earlier revision of this PR was dropped from it).
- The **staking amounts** live in `ChainExtraData` on the stored tx record, which is a read cache rather than index output — `PutTx` has a single caller (`db/txcache.go`, on an API-read miss) and the indexer never writes tx records, which is why DB migrations drop the whole `transactions` column family. A reindex heals this too, but dropping and recreating that column family is sufficient on its own.

**Operator notes:** after deploying, regenerate `blockchaincfg.json`, restart, and reindex. The backend must run with `vm.saveInternalTx = true` (the packaged config now enforces it). Until the reindex completes, expect benign `Contract … not found in DB` warnings as pre-existing TRC-20 contracts self-heal on first read — they signal missing internal data in old blocks, not corruption. If the `transactions` column family is cleared separately, run `ldb` as the blockbook system user, not root, or the recreated MANIFEST/OPTIONS/LOG end up root-owned and Blockbook cannot reopen the DB; expect one slow, node-heavy warm-up afterwards while the cache refills.

**Testing:** unit tests cover the CREATE classification (the #1660 call-with-`contract_address` case fails on master), failed deployments, featured-frame skipping, TRC-10 filtering, rejected frames, the ephemeral create→selfdestruct registry event order, and on the staking side the V1-vs-V2 unfreeze branch, `cancelUnfreeze`, the sweep parsed for every contract type, and that the withdraw path no longer duplicates its amount into `unstakeAmount`. The DB-side merge and disconnect rollback are tested in #1703. `go test -tags unittest` passes for `bchain/coins/tron`, `db`, `api`.

Verified against a live backend: running the block builder over block 83,534,426 books the reference tx `d54a1abc…`'s 13,245.012561 TRX refund to `TT2T17KZ…` as an internal transfer.

Verified end to end on the deployed dev index at this branch's HEAD with the tx cache cleared:

- `TT2T17KZ…` reconciles exactly — `received - sent = 291,213,156.988331` equals the live balance to the sun (residual 0, was 600,000,000), `received >= sent` holds, and the forward cumulative bottoms at 95.22 TRX with zero negative days.
- Its full staking identity closes: V2 unfreeze 3,435,808,377.400336 = swept 3,091,132,253.692002 + still-locked 344,676,123.708334, the last matching `stakingInfo.unstakingBatches` to the sun.
- The sweeping tx `1b6764ae5a99acc75fb9d9cd02e4ecd7f33fd0b547d6b969e3a8375260cb33f6` (block 56,295,136) now carries both `unstakeAmount: 400000000000000` and `withdrawnUnfreeze: 600000000000000`, confirming from node data that the two amounts genuinely co-occur. Of that account's 11 sweeps, 10 were explicit withdraws (already correct before this PR) and exactly one was the implicit `UnfreezeBalanceV2` sweep. No withdraw tx still carries a duplicate `unstakeAmount`.
- `TUAAqYySyBJDLJDxnQKqtcPBW1JXJuqrSS` (Stake 1.0) improved from a 167,570 TRX residual and now satisfies `received >= sent`.
- A sample of 9 further staking and resource-delegation accounts reconciles at residual 0.

**Known remaining gap, not fixed here:** some internal TRX inflows are still absent from the index, which keeps `getBalanceHistory` genuinely negative for a minority of contract-fed accounts — e.g. `TChPByGjbBDsL2rPpwd51xmQGWJNi1U68C`, residual 86,478.32 TRX with the forward cumulative bottoming at −85,982. The accounting logic is not at fault: rebuilding it independently from all 6,364 of that account's transactions, `sent` matches Blockbook to the sun on every day and `received` matches once indexed internal transfers are included, so the shortfall is inflow that was never indexed. `TUAAqYy…` shows the same shape with a 20,001.40 TRX residual. Worth tracking separately from this PR.


cranycrane:
### Code review

Found 1 issue:

1. `internalDataEqual` breaks backfill idempotency for CREATE transactions on non-Tron Ethereum-type chains (bug due to `api/internaldata_backfill.go` comparing `stored.Contract != computed.Contract` with asymmetric normalization). The computed side comes raw from the trace (`d.Contract = r.To` in [ethrpc.go#L1554-L1558](https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/bchain/coins/eth/ethrpc.go#L1554-L1558), lowercase hex) and the `fixEIP55` block only normalizes `Transfers[i].From/To`, never `internalData.Contract` ([ethparser.go#L158-L166](https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/bchain/coins/eth/ethparser.go#L158-L166)), while the stored side is unpacked via `GetAddressesFromAddrDesc` → `EIP55Address`, which always returns checksummed mixed-case ([rocksdb_ethereumtype.go#L835-L843](https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/db/rocksdb_ethereumtype.go#L835-L843)). So on a standard eth-type chain every contract-creation tx compares unequal, `filterAlreadyReconnectedInternalData` never skips it, and re-running a range re-runs `ReconnectInternalDataToBlockEthereumType` on it — re-incrementing the contract transaction counters that the filter's own comment says it exists to prevent, and contradicting the "Re-running a range is safe" contract of `StartInternalDataBackfill`. Tron is unaffected (base58 is case-canonical both ways), which is why the tests in `api/internaldata_backfill_test.go` don't catch it — consider comparing address descriptors or normalizing both sides before comparing.

https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/api/internaldata_backfill.go#L168-L174

cranycrane:
### Code review — follow-up (verified against live backend)

Tested the PR's assumptions against java-tron (`backend15.sldev.cz:8091` HTTP + `:8545` JSON-RPC) over ~720 recent mainnet blocks. The core classification logic checks out on live data: `to` is null only for `CreateSmartContract` (verified via both `eth_getTransactionByHash` and bulk `eth_getBlockByNumber`, which is what `GetBlock` parses), `contract_address` is filled for ordinary `TriggerSmartContract` calls (2,642 of 12,589 txs vs. 4 actual creates in one sample), factory creates carry the factory — not the child — in `contract_address`, the only notes observed were `call`/`create`/`suicide`, and the issue #1660 example tx decomposes exactly as documented. Found 2 issues in how the enabled pipeline handles real chain data:

1. Rejected internal transactions are booked as real value flows. The transfer loop appends a transfer for every internal tx with `callValue > 0` without checking `itx.Rejected`, and neither `processInternalData` ([rocksdb_ethereumtype.go#L601-L607](https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/db/rocksdb_ethereumtype.go#L601-L607)) nor `processInternalTransactionsForBalanceHistory` ([worker.go#L1803-L1806](https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/api/worker.go#L1803-L1806)) gates on the stored error. Live specimen: `ab58f2059d7fd71350d2ae7e862bea3dd428040b6c87cf8ee98ca70a04af65fa` failed with `OUT_OF_ENERGY`; all 12 internal txs are `rejected: true`, two carrying `callValue: 457584` — both would be indexed and booked into balance history although no TRX moved. A failed high-value DeFi call would inject its entire internal flow into the history this PR is fixing. Ethereum's tracer path has the same weakness, but Tron hands over an explicit `rejected` flag, so skipping is a one-line guard.

https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/bchain/coins/tron/tronInternalDataProvider.go#L200-L214

2. The ephemeral create→use→selfdestruct pattern corrupts the contract registry. `detectSelfDestructedContract` runs before the internal-tx loop and sets `d.Contract` to the first suicided contract; the nested-create guard `to != d.Contract` then skips registering that same contract's creation, producing a destroyed-but-never-created entry — and for txs with multiple suicides only the first destruction is ever registered, so the remaining children (registered as created by the new nested-create block) become permanent zombies. Live specimens: `76a2228970dd1f9198b8c5edefab46fe7bfe3b1fed81fabd5b398d23c1fa916c` (SUCCESS factory call: 3 creates + 3 suicides → recorded as 1 destroyed-never-created + 2 created-never-destroyed; eth's `processCallTrace` would record all 6 registry events), plus `4391da24…`, `be72904c…`, `ea2ce22c…` (single ephemeral). This pattern is common: 36 of 600 scanned blocks contained suicide notes and every sampled one was ephemeral (MEV-style), i.e. roughly 2,500 tx/day. Note the same txs also illustrate that zero-value nested creates/suicides emit no transfer at all (every one of the 81 observed create notes had empty `callValueInfo`), so these ephemeral contracts' own address histories stay empty — unlike eth, which appends CREATE/SELFDESTRUCT transfers unconditionally.

https://github.com/trezor/blockbook/blob/fb3b34edf94cd7717416f13c846f17a7022e2edc/bchain/coins/tron/tronInternalDataProvider.go#L165-L199

pragmaxim (approved):
LGTM

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -149,32 +149,21 @@ func buildInternalDataFromTronInfos(
...
-		}
-
-		if topType == bchain.CREATE && contractAddr != "" {
+		// A transaction deploys a contract only when its eth-style representation
+		// has no recipient (java-tron's JSON-RPC leaves `to` null solely for
+		// CreateSmartContract) AND the node reported the deployed address in the
+		// transaction info. contract_address alone is not a creation signal:
+		// java-tron fills it for ordinary TriggerSmartContract calls as well.
+		deployedContract := ""
+		if tx.To == "" && info.ContractAddress != "" {
```
cranycrane: **Medium — the comment's stated reason is factually wrong (the code is right).**

`to` is *not* null "solely for `CreateSmartContract`". Of 220 null-`to` txs in a 134,239-tx mainnet sample, **216 are native non-EVM contracts**: `WithdrawBalance` 95, `FreezeBalance` 60, `UnfreezeBalance` 31, `FreezeBalanceV2` 14, `UnfreezeBalanceV2` 10, `WithdrawExpireUnfreeze` 3, `ExchangeTransaction` 1, `CancelAllUnfreezeV2` 1, `AccountPermissionUpdate` 1. Verified individually, e.g.

```
FreezeBalanceV2Contract    b02358bb…ccd9b → to: null
WithdrawBalanceContract    8b88c532…81a87 → to: null
CreateSmartContract        d3ba408f…b1d7  → to: null
TriggerSmartContract       d54a1abc…a547  → to: "0x72db65b2e023e4783d46023e7135c692e527f6cb"
```

All the discriminating power comes from the `info.ContractAddress != ""` half of the `&&` — `contract_address` is empty for every non-VM tx. The conjunction is a perfect discriminator (0 FP / 0 FN on the sample), so the behaviour is correct; only the justification is not.

Two suggestions:
1. Reword to say the pair is what discriminates, and that `contract_address` is empty for native non-EVM contracts.
2. Consider using the authoritative signal instead — `GetBlock` already has `RawData.Contract[0].Type == "CreateSmartContract"` available (both maps are awaited before `tronrpc.go:883`). Relying on `getTo()` returning null is a java-tron JSON-RPC artifact; any proxy or alternate node that normalises `to` to the deployed address would silently turn every deployment into a `CALL` and register no contract at all.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -184,14 +173,53 @@ func buildInternalDataFromTronInfos(
 				return data, contracts, err
 			}
 
+			// a rejected internal transaction did not execute - it moved no
+			// value, created no contract and destroyed none; it only flags
+			// the transaction error below
+			if itx.Rejected {
+				continue
+			}
```
cranycrane, line 188: **High — `rejected` is per-frame, not per-tx, so the tail loop at lines 250-259 stamps an error on successful transactions.**

Skipping the transfers here is right. The problem is the companion loop below: java-tron rejects a *single* frame when a nested call reverts (`Program.java:955/1161` → `internalTx.reject()` then `stackPushZero()`) and the outer transaction keeps running and can succeed. Real mainnet case:

```
tx 4ae4a08e829dca98c8279adf18bfe8ed8eeeca2d83b3697f1a3b24c8c3e862fc   block 82,000,016
receipt.result = SUCCESS      eth receipt status = 0x1      log count = 1
internal_transactions: 8 frames, ALL rejected:true, all note "call"
```

That yields `ethereumSpecific.error = "Internal transaction rejected"` alongside `status: 1`. The web templates gate the error text on `Status == 0` (`tx_tron.html:39`, `txdetail_tron.html:8`) so the HTML UI is unaffected, but any JSON consumer (Suite) sees an error on a transaction that succeeded. Frequency in the sample: 1 in 16,915 SUCCESS txs carrying internal frames — rare per-tx, thousands chain-wide.

Suggest only appending the note when the receipt itself is non-SUCCESS:

```go
if d.Error != "" {
    d.Error += "; internal transaction rejected"
}
```

Related nit: for the 607 failed txs in the sample, *every* frame is rejected, so the suffix is unconditional noise on `"REVERT"` / `"OUT_OF_ENERGY"`. And unlike eth, this string bypasses `PackInternalTransactionError`'s 1-byte compression (`ethparser.go:655-670`) and is stored raw.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -184,14 +173,53 @@ func buildInternalDataFromTronInfos(
...
+
 			from := ToTronAddressFromAddress(itx.CallerAddress)
 			to := ToTronAddressFromAddress(itx.TransferToAddress)
 
+			// registry events are emitted in note order (parity with eth
+			// processCallTrace), so that an ephemeral contract's creation is
+			// stored before its destruction merges into it
```
cranycrane: **High — the in-block merge this comment relies on does not exist; the destruction is silently dropped.**

`storeContractInfo` resolves a destruction entry by reading *committed* state:

```go
// db/rocksdb_contracts.go:173-183
if contractInfo.CreatedInBlock == 0 && contractInfo.DestructedInBlock != 0 {
    storedCI, err := d.GetContractInfo(key, "")   // LRU + d.db.GetCF only
    if storedCI == nil { return nil }             // <- destruction dropped
```

`GetContractInfo` reads `d.db.GetCF` (`db/rocksdb_contracts.go:96`); the creation `PutCF` emitted moments earlier sits in an uncommitted `grocksdb.WriteBatch` that is flushed only after the whole block (`db/rocksdb.go:734` then `:746`), and there is no `WriteBatchWithIndex` anywhere in the repo. The creation also `cachedContracts.delete`s its key, so the cache misses too.

So for a *fresh* ephemeral contract — exactly the create→use→selfdestruct case this commit targets — the destruction returns early and the row ends as `{CreatedInBlock: H, DestructedInBlock: 0}`, i.e. recorded as alive. Emission order is inert here; it only matters for a contract created in an *earlier* block. The same applies when the creator and destroyer are two different txs in one block.

The unit test "Ephemeral contract created and destroyed in one call" passes because it asserts the returned slice, not what the DB retains — and its comment ("creation first, so that the destruction merges into the stored creation") repeats the same claim.

Either dedupe/merge the `[]bchain.ContractInfo` for the block before storing, or give `storeContractInfo` an in-batch map to consult — and until then, drop the ordering rationale from both comments so it does not read as a solved problem.

A second, smaller issue in the same helper: line 181 mutates the object returned by `GetContractInfo`, which may be the shared LRU value (`db/contract_info_cache.go:61`), before the batch commits — a concurrent reader can observe `DestructedInBlock` early, and permanently if the batch write then fails.

cranycrane (commented):
Review follow-up: the **High** and **Medium** findings from my analysis, as inline comments. (The three blocking items — failed-deploy contract registration, the PR-body/commit-message mismatch with the absent backfill, and the unbumped `package_revision` — are tracked separately, as are the test-hardening notes.)

Data claims below were verified against a live mainnet node (`backend15`, 1,080 blocks / 134,239 txs / 90,088 internal frames) and against java-tron `GreatVoyage-v4.7.7` sources.

For the record, the core fix checks out: `contract_address` is set for 41,700 of 41,704 sampled txs that are ordinary `TriggerSmartContract` calls (only 4 are deploys), so the old classifier misread essentially every contract call as a deployment; and `tx.To == "" && contract_address != ""` is a perfect discriminator on that sample (0 false positives, 0 false negatives). The reference tx `d54a1abc…a547` does carry the 13,245.012561 TRX refund that master drops.

### Medium findings that have no anchor in this diff

- **`cfContracts` is never rolled back, and this PR amplifies the drift.** Nothing in `disconnectBlockTxsEthereumType` deletes from `cfContracts` (`db/rocksdb_contracts.go:114/184/278` are the only writers; 278 is the admin DELETE). Commit 389d3b9d changes Tron from one registry event per *transaction* to one per *frame*, so a reorged block — routine on Tron — now leaves proportionally more stale rows, and a stray row permanently disables `GetBalanceHistory` for that address (`api/worker.go:2085`). Either mirror `disconnectErcProtocols` over the disconnected range, or document the acceptance.
- **Post-deploy log wave.** `api/worker.go:709-712` emits `Contract %v not found in DB` only when `ProcessInternalTransactions` is on. On a Tron DB whose history was indexed without it, the first read of every pre-existing TRC-20/721 logs this before self-healing. It is the same log shape that signals a genuine index gap on other chains — worth a line in the PR so it is not misread as corruption.
- **`internalTxs` counts transfer legs, not transactions.** `ac.InternalTxs++` (`db/rocksdb_ethereumtype.go:489`) fires once per leg, outside the per-tx dedupe at `:517-520`, so `?filter=1` page counts can overshoot what `getAddressTxids` returns. Pre-existing eth behaviour, newly reachable on Tron.
- **An empty `cfInternalData` row per Tron transaction.** `EthTxToTx` nils out empty internal data (`bchain/coins/eth/ethparser.go:154-158`), but `buildTxFromHTTPData` then overwrites `tx.CoinSpecificData` (`bchain/coins/tron/tronrpc.go:716`), restoring a non-nil pointer, so `db/rocksdb_ethereumtype.go:730` always runs. Eth writes no row at all. Pure storage waste at Tron's tx volume; consider passing `nil` when the entry is empty.
- **`GetInternalDataForBlock` (lines 52-78) is unreachable in production** — `TronRPC.GetBlock` shadows `EthereumRPC.GetBlock` and calls `buildInternalDataFromTronInfos` directly (`tronrpc.go:883`). Worth a comment, since it is the method the provider unit test exercises while production takes another path. Separately, its `return nil, nil, err` at `:73` diverges from eth's contract of returning a correctly-sized `data` slice on error (`ethrpc.go:1525`), which the caller relies on for `&internalData[i]` — harmless while the method is dead, a latent nil-slice index panic if it is ever wired up.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -184,14 +173,53 @@ func buildInternalDataFromTronInfos(
...
+			switch t {
```
cranycrane, line 204: **High — unrecognised notes fall through to `CALL`, turning staking operations into phantom multi-million-TRX transfers.**

`tronNoteHexToInternalType` default-maps any unknown note to `CALL` (`bchain/coins/tron/tronparser.go:348-350`), so the frame's `callValue` is booked as a TRX transfer at line 225. But featured notes carry the **staked/delegated amount**, which never moves. Observed live on `backend15`:

```
note delegateResourceOfEnergy     caller 41c64e69… → 417b098c…  callValue 8333333000000   (8,333,333 TRX)
note unDelegateResourceOfEnergy   caller 41c64e69… → 41c9b586…  callValue 1255110000000   (1,255,110 TRX)
```

Across the sample: 43 such frames, **54,298,294 TRX of transfers that did not happen**. Proof the value stays put — the delegator `TU3kjFuhtEo42tsCBtfYUAZxoqQ4yuSLQ5` is not drained; `/walletsolidity/getaccount` shows the amount under `account_resource.delegated_frozenV2_balance_for_energy`, and a same-block delegate/undelegate pair (`1ff0e6d9…bb4c` / `8cd47fcd…66f8`, block 65,000,007) moves an identical 1,255,110 TRX in both directions while the txs' own `call_value` is 1,649 TRX / null.

Whether these reach Blockbook is a node-config question — java-tron filters the list to `call`/`create`/`suicide` unless `vm.saveFeaturedInternalTx` is on (`chainbase/.../capsule/utils/TransactionUtil.java:127-140`, v4.7.7). Which means:

- **Mainnet as packaged is safe** — the pinned `main_net_config.conf` leaves `saveFeaturedInternalTx` commented out at line 701.
- **Nile as packaged is not.** `config-nile.conf` ships `saveInternalTx = true` **and** `saveFeaturedInternalTx = true` (lines 530-531 of the pinned `3f6f53ca…` revision), and this PR adds no sed to turn it off. Nile will fabricate these transfers, so it is not a faithful pre-flight for mainnet.
- **`backend15` clearly runs with the flag on** — the frames quoted above came from it — so the node used to verify this change has a different data shape from the packaged config.

I would fix this in the parser rather than depend on node config: return an explicit "ignore this frame" outcome for anything that is not `call`/`create`/`suicide`. There are ~28 such notes in v4.7.7 (`freezeFor*`, `unfreezeFor*`, `freezeBalanceV2For*`, `unfreezeBalanceV2For*`, `delegateResourceOf*`, `unDelegateResourceOf*`, `withdrawReward`, `voteWitness`, `withdrawExpireUnfreeze*`, `cancelAllUnfreezeV2`, each × resource type), and all the value-bearing ones behave identically. Optionally also add the nile sed for symmetry with mainnet.

Note this used to be visible: the removed `detectTopType` at least logged `glog.Warningf("Unknown Tron internal transaction type %v", t)`. After this PR an unhandled note type is completely silent.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -184,14 +173,53 @@ func buildInternalDataFromTronInfos(
...
+			case bchain.CREATE:
+				// nested create frames register the child contract but do not
+				// change the top-level type - factory calls remain CALLs
+				if to != "" && to != deployedContract {
```
cranycrane, line 208: **Medium — the `deployedContract` guard covers the registry entry but not the transfer emitted at line 237.**

The guard exists to stop the root create being registered twice should java-tron ever list it among `internal_transactions`. But the zero-value fallback below checks only `to != ""`, so in that same scenario you would still emit a duplicate `CREATE` transfer for the deployed contract. Either both sites should consult `deployedContract` or neither should.

For what it is worth, the guard cannot suppress a legitimate nested CREATE: `deployedContract` is non-empty only when `tx.To == ""` (a `CreateSmartContract`), a CREATE2 redeploy always goes through a factory (`tx.To != ""`), and a nested create cannot land on the root contract's own tx-hash-derived address. It is also empirically untested — all 4 `CreateSmartContract` txs in my sample have no `internal_transactions` at all, so the skip arm never runs today, and the "Deployment with constructor-created child contract" test uses a *different* child address. A regression that removed the guard would double-register every deployment and still pass the suite.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -184,14 +173,53 @@ func buildInternalDataFromTronInfos(
...
+						CreatedInBlock: blockHeight,
+						Standard:       bchain.UnhandledTokenStandard,
+					})
+				}
+			case bchain.SELFDESTRUCT:
+				if from != "" {
+					contracts = append(contracts, bchain.ContractInfo{
+						Contract:          from,
+						DestructedInBlock: blockHeight,
+					})
```
cranycrane, line 220: **Medium — moving from one registry event per transaction to one per frame amplifies reorg drift.**

This is the right semantics (eth registers every nested create and suicide — `bchain/coins/eth/ethrpc.go:1477` and `:1485` — so this is genuine parity), but it interacts badly with a pre-existing gap: **nothing ever removes `cfContracts` rows on disconnect.** `DisconnectBlockRangeEthereumType` clears `cfTransactions`, `cfInternalData`, `cfAddresses`, `cfBlockTxs`, `cfHeight`, `cfBlockInternalDataErrors` and reverts `cfErcProtocols` (`db/rocksdb_ethereumtype.go:1382-1426`), but `cfContracts` is untouched — its only writers are `db/rocksdb_contracts.go:114/184` plus the admin DELETE at `:278`.

On a chain with routine 1-block reorgs, each orphaned block now leaves proportionally more rows whose `CreatedInBlock`/`DestructedInBlock` point at a height that no longer contains the tx. The sharp edge is `api/worker.go:2085`: any address with a `cfContracts` row gets `"GetBalanceHistory for a contract not allowed"` permanently, repairable only via `DELETE /admin/contract-info/<address>` — which is itself lossy (see the comment at `db/rocksdb_contracts.go:248-251`).

Suggest either deleting `cfContracts` rows whose `CreatedInBlock` falls in the disconnected range (mirroring `disconnectErcProtocols`), or explicitly noting the accepted drift in the PR.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -184,14 +173,53 @@ func buildInternalDataFromTronInfos(
...
+					// Tron internal transactions describe only nested frames,
+					// so unlike eth the top-level SELFDESTRUCT type is
+					// inferred from the first destroyed contract
+					if d.Type == bchain.CALL {
+						d.Type = bchain.SELFDESTRUCT
+						d.Contract = from
+					}
+				}
```
cranycrane: **High — this inference cannot survive packing, and the field it writes means something else.**

Two independent problems.

**1. It is dead code.** `packEthInternalData` stores the type in a *single bit* and only packs the contract for `CREATE`:

```go
// db/rocksdb_ethereumtype.go:812-817
// internalType is one bit (CALL|CREATE), it is joined with count of internal transfers*2
l := packVaruint(uint(data.internalType)&1+uint(len(data.transfers))<<1, varBuf)
...
if data.internalType == bchain.CREATE {
```

`SELFDESTRUCT` is 2 (`bchain/types_ethereum_type.go:59-62`), so `2&1 == 0` → the row unpacks as `CALL` with no contract (`unpackEthInternalData:835`), and `processInternalData` indexes `id.Contract` only for `CREATE` (`:608`). Nothing set here can reach `/api/v2/tx`. The tests asserting `wantType: bchain.SELFDESTRUCT` (test file lines 294 / 331 / 374) therefore pin behaviour no client can observe.

**2. If it ever were persisted, it would be wrong.** `EthereumInternalData.Contract` surfaces as `EthereumSpecific.CreatedContract` — documented as *"Address of contract created by this transaction"* (`api/types.go:303`, set at `api/worker.go:515`) — and the template renders it only under `{{if eq $tx.EthereumSpecific.Type 1}}` labelled "Contract creation" (`txdetail_tron.html:64`). Putting a **destroyed** contract there is a category error. Concrete case:

```
tx 23bffcf1831a75c18dc318faf68e5faa3a2eb9734902979ea8078a4066d03016   block 85,056,933
top-level: TriggerSmartContract → 0x3fea1575…, SUCCESS, eth to non-null
frames: create(413fea15…→41e065e6…), 4× call, suicide(41e065e6…→4118c792…)
```

This code yields `type = 2` and `createdContract = TWRiSFE9UFw9mufv6uEto5ZZyTfUrQ9Mks` — an ephemeral contract that was destroyed, not created. Geth on the same trace shape reports `type = CALL`, `createdContract = null`: `processCallTrace` never touches `d.Type`/`d.Contract`, and eth sets the root type from the root frame only (`ethrpc.go:1554-1560`). All 56 suicide-bearing txs in my sample are top-level `TriggerSmartContract`, so 56/56 would be relabelled.

It is also inconsistent with the sibling rule eight lines up, which deliberately leaves a nested CREATE from changing the top-level type ("factory calls remain CALLs") — the same nesting level, opposite treatment.

Suggest deleting the `d.Type`/`d.Contract` mutation entirely (and the assertions that depend on it). The per-transfer `Type: SELFDESTRUCT` already round-trips correctly — `t.internalType` is packed as a full byte at `db/rocksdb_ethereumtype.go:820` — and that is what the UI actually renders (`txdetail_tron.html:85`).

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -184,14 +173,53 @@ func buildInternalDataFromTronInfos(
...
+			transferEmitted := false
 			for _, cv := range itx.CallValueInfo {
 				// skip TRC-10
 				if cv.CallValue <= 0 || cv.TokenID != "" {
 					continue
 				}
+				transferEmitted = true
 
 				val := *big.NewInt(cv.CallValue)
 				d.Transfers = append(d.Transfers, bchain.EthereumInternalTransfer{
```
cranycrane, line 238: **Medium — this loop can emit several transfers for one frame, where eth emits exactly one.**

eth's `processCallTrace` appends a single transfer per frame (`bchain/coins/eth/ethrpc.go:1470-1495`); here every positive non-TRC-10 `callValueInfo` entry produces its own. Low practical risk — java-tron puts the sole TRX entry at `callValueInfo[0]` and any extras are TRC-10 (`buildInternalTransaction`, L155-167) — and I saw no counterexample in 90,088 frames, so this is a robustness note rather than a live bug. Worth either a comment recording the invariant, or breaking after the first accepted entry.

The TRC-10 filter itself is correct and load-bearing: 30 frames in the sample have the two-entry shape `[{}, {"tokenId":"1002000","callValue":57753367}]` and the `TokenID != ""` test drops only the token leg. The `CallValue <= 0` skip is also required — `callValueInfo: [{}]` is by far the most common shape. Neither is covered by a test case, though, and the interaction with the fallback below is subtle: a TRC-10-only **create** frame still emits a zero-value `CREATE` transfer (because `transferEmitted` stays false), while a TRC-10-only **call** frame emits nothing. Deliberate, I assume, but unpinned.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -201,6 +229,18 @@ func buildInternalDataFromTronInfos(
...
+			// transfer even when no TRX moved, so that the created/destroyed
+			// contract's own address history contains this transaction -
+			// plain zero-value calls stay skipped
+			if !transferEmitted && ((t == bchain.CREATE && to != "") || (t == bchain.SELFDESTRUCT && from != "")) {
+				d.Transfers = append(d.Transfers, bchain.EthereumInternalTransfer{
+					Type: t,
+					From: from,
+					To:   to,
+				})
+			}
```
cranycrane, line 257: **Medium — the emptiness guards never fire in practice, so the stated rationale rests on cases that do not occur.**

The eth-parity claim itself is accurate — `processCallTrace` does emit CREATE and SELFDESTRUCT transfers regardless of value while skipping zero-value plain calls (`bchain/coins/eth/ethrpc.go:1470-1489`) — so the mechanism is right. But across 90,088 mainnet frames: **0 have an empty or missing `transferTo_address`, and 0 have an empty `caller_address`.** All 83 `create` frames and all 56 `suicide` frames carry both endpoints, and 83/83 creates are zero-value, so the fallback fires for every real create frame. The cause is structural: java-tron builds both addresses via `DataWord.toTronAddress()`, which cannot yield empty.

So `to != ""` / `from != ""` is dead defensive code, and the test case asserting a suicide with `to: ""` ("no beneficiary recorded") does not correspond to a shape the node emits. Not worth removing — it is cheap and correct — but the comment reads as if the empty case is the motivating scenario, and it is not.

If such a frame ever did appear, there is a small one-way drift: connect skips the unparseable address (`db/rocksdb_ethereumtype.go:635-639`, `ErrAddressMissing` handled quietly), while `appendAddress` stores 20 zero bytes (`:796-805`) that unpack to the valid Tron zero address `T9yD14Nj9j7xAB4dbGeiX9h8unkKHxuWwb`, so `disconnectInternalData` (`:1329-1344`) decrements `TotalTxs`/`InternalTxs` that were never incremented. Also note `TronParser.GetAddrDescFromAddress` logs at `glog.Infof` for every such call (`tronparser.go:71`), defeating the caller's deliberate silence on `ErrAddressMissing` — eth's parser logs nothing.

The record format is otherwise safe here: fixed-width addresses mean a zero-length descriptor can never shorten a record or desynchronise the transfers that follow, and a zero `Value` packs to a single byte and unpacks back to 0 (`db/rocksdb.go:2904-2951`).

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -149,49 +154,82 @@ func buildInternalDataFromTronInfos(
...
+		// A transaction deploys a contract only when its eth-style representation
+		// has no recipient AND the node reported the deployed address. Neither
+		// signal alone discriminates: java-tron fills contract_address for
+		// ordinary TriggerSmartContract calls too, and `to` is null also for
+		// native non-VM operations (FreezeBalance, WithdrawBalance, ...), which
+		// have an empty contract_address. A failed deployment created nothing -
+		// java-tron precomputes contract_address even for those.
+		deployedContract := ""
+		deploySucceeded := info.Receipt.Result == "" || info.Receipt.Result == "SUCCESS"
+		if tx.To == "" && info.ContractAddress != "" && deploySucceeded {
```
pragmaxim: `tx.To == ""` uses stricter semantics than `EthTxToTx`, which treats a recipient as present only when `len(tx.To) > 2` — i.e. `"0x"` counts as *missing* there but as *present* here. If java-tron's eth-compat endpoint ever reported `"to": "0x"` for a deployment, the receipt fallback in `buildTxFromHTTPData` would still fix the vout and internal data, but the `cfContracts` registry entry would be silently lost (it is only produced here). Live verification says `to` is null today, so this is purely defensive — but `len(tx.To) <= 2` would match the `EthTxToTx` semantics at no cost.

bchain/coins/tron/tronrpc.go
```diff
@@ -687,9 +692,12 @@ func (b *TronRPC) buildTxFromHTTPData(txByID *tronGetTransactionByIDResponse, tx
 		return nil, errors.Annotatef(err, "txid %v", txByID.TxID)
 	}
 
+	// java-tron precomputes the receipt ContractAddress even for failed
+	// deployments - only a successful receipt proves the contract exists
 	if len(tx.Vout) > 0 &&
 		tx.Vout[0].ScriptPubKey.Addresses == nil &&
 		csd.Receipt != nil &&
+		csd.Receipt.Status == "0x1" &&
```
pragmaxim: This gate and `deploySucceeded` in `buildInternalDataFromTronInfos` read two different success signals: `Status` is derived from the top-level `result` field (`"FAILED"`/empty) in `tronBuildRpcReceipt`, while `deploySucceeded` reads `Receipt.Result` (`"SUCCESS"`/failure code). java-tron sets them consistently, but if a shape with `receipt.result = OUT_OF_ENERGY` and an empty top-level `result` ever appeared, the two paths would disagree — internal data would say CALL+error while this fallback upgrades to CREATE. Deriving both from the same field would remove the possibility.

bchain/coins/tron/tronInternalDataProvider.go
```diff
@@ -149,49 +154,82 @@ func buildInternalDataFromTronInfos(
...
+						Contract:          from,
+						DestructedInBlock: blockHeight,
+					})
+				}
+			}
+
+			// java-tron puts at most one TRX entry in callValueInfo (extras
+			// are TRC-10), so like eth this emits one transfer per frame
+			transferEmitted := false
 			for _, cv := range itx.CallValueInfo {
```
pragmaxim, line 230: The `to != deployedContract` guard protects the registry entry and the zero-value transfer below against a hypothetical root create frame, but this valued-transfer loop books all TRX entries unguarded. In that same hypothetical — java-tron emitting a root frame carrying the deployment's callValue — the value would double-count with the tx-level value already booked on the vout. java-tron records only nested frames, so this is dead-shape territory, but the asymmetry deserves either a comment or the same guard for CREATE frames here.

pragmaxim (commented):
Three minor robustness observations from a detailed pass — none blocking, all defensive against node shapes that today's java-tron does not emit.

pragmaxim (approved):
LGTM

cranycrane:
### Code review

Reviewed head `3be75f84` (12 commits since the previous review rounds). All earlier inline threads are addressed at this head (rejected frames, featured notes, per-frame `rejected` error stamping, top-level SELFDESTRUCT inference, `deployedContract` guard symmetry, `0x` recipient, shared success predicate), and the same-block create-then-selfdestruct merge plus `cfContracts` rollback now exist via #1703. The balance-history `withdrawnUnfreeze` credit is applied once, for the owner vin only, and the eth-parity internal-data drop is safe against the receipt CREATE fallback.

Found 1 issue:

1. The mainnet `extract_command` rewrite drops `constantCallTimeoutMs = 5000` from `main_net_config.conf`. Master commit [b3763ded](https://github.com/trezor/blockbook/commit/b3763dedc55e530738200f14a522b93aad7d0e03) ("chore(tron-backend): increase timeout to allow 100 multicall batch sizes", merged in #1700) changed the `supportConstant` sed to a two-line replacement that also injects that setting. The PR's rebased commit 9b0ddfae rewrote the same string and restored the old single-line `supportConstant = true` form, so the diff against the base removes the timeout bump while keeping the `package_revision` bump that will trigger a backend rebuild with the regressed config (bug due to a rebase conflict resolution in `configs/coins/tron.json`; `grep -r constantCallTimeoutMs` at this head returns nothing). Full-size Multicall3 `aggregate3` batches on Tron then fall back to the per-call RPC path that #1700 was built to avoid, or time out under load. Re-add `\\n  constantCallTimeoutMs = 5000` to the `supportConstant` sed.

https://github.com/trezor/blockbook/blob/3be75f8421880e3faa33e6779b36b460f9777dc1/configs/coins/tron.json#L27-L29

53gur0 (approved):
utACK

PR #1687, opened 2026-07-30
feat(grafana): chart websocket IP blocklist metrics

cranycrane:
## What

Adds two Grafana panels for the websocket IP-blocklist metrics that were registered in `metrics.yaml` when the per-connection message rate limiter shipped, but never got charted:

- **Websocket IPs currently blocked** — `blockbook_websocket_blocked_ips`, the "users locked out right now" gauge
- **Websocket blocked-IP rejections per minute** — `rate(blockbook_websocket_blocked_connections)`, banned clients hammering reconnects

Both sit next to the existing connection-rejections panel so the enforcement panels are grouped together.

## Why

The message-rate limiter's 12 h IP bans currently have zero dashboard visibility: on the Tron backends the blocklist held 24 (backend16) / 2 (backend15) IPs with ~33k rejected reconnect attempts, and the first signal was user reports of "Account discovery error", not a chart. The `message_rate_limit` channel closes are technically visible in the closes panel but are drowned out by ~90k `read_error` closes.

With this change every metric in `metrics.yaml` is referenced by at least one panel (verified both directions: Go `metric:` tags ↔ `metrics.yaml` ↔ `panels.yaml`).

`render_grafana.py --check` passes (106 panels, 87 metrics).

### DIFF
--- a/configs/grafana/panels.yaml
+++ b/configs/grafana/panels.yaml
@@ -400,11 +400,11 @@ panels:
         legend: '{{coin}} - {{reason}}/min'
   websocket.blocked_ips:
     title: Websocket IPs currently blocked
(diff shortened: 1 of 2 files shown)


pragmaxim (approved):
LGTM

PR #1688, opened 2026-07-30
feat: increase WS rate limit default to 5000 and make max connections per IP configurable

pragmaxim:
## Changes

Two commits on this branch:

### 1. Increase `_WS_MESSAGE_RATE_LIMIT` default from 2500 to 5000 (`3981804`)

Bumps the per-connection websocket message rate limit to 5000 messages per 10-minute window, providing more headroom for deployment patterns with higher burst traffic.

### 2. Make `maxWebsocketConnectionsPerIP` configurable (`cc1af5c`)

Replaces the hardcoded `maxWebsocketConnectionsPerIP = 128` constant with a configurable limit read from the `<network>_WS_MAX_CONNECTIONS_PER_IP` environment variable.

- New env var: `<network>_WS_MAX_CONNECTIONS_PER_IP` (default `128`, `0` = unlimited)
- The limit is stored directly in `websocketConnectionLimiter` under its mutex, matching the single-storage convention used by sibling configs
- Setting `0` disables the per-IP connection cap entirely
- Documented in `docs/env.md`

## Code review findings applied

- Removed redundant `maxConnectionsPerIP` field from `WebsocketServer` struct (the limiter is the single source of truth)
- The configure method writes to the limiter under its mutex, safe for both startup and any future runtime reconfiguration (e.g. SIGHUP)
- Zero value correctly means unlimited connections

## Testing

- `go build` ✓
- `go test -tags unittest ./server/`: 580 passed ✓


cranycrane:
the rate limit was set specifically using ENV variables for the requested backend. and the new env variable `_WS_MAX_CONNECTIONS_PER_IP ` was already implemented in https://github.com/trezor/blockbook/pull/1679

PR #1693, opened 2026-07-31
feat(eth): automatic self-healing of the internal data error queue

cranycrane:
Stacked on #1682, which fixes Tron internal transaction processing and enables `processInternalTransactions`.

When a block's internal-data fetch fails during sync, the block is indexed without its internal transactions and parked in the internal-data error queue. Until now that queue only drained when an operator noticed it and pressed **Retry fetch** on `/admin/internal-data-errors` — so in practice a transient backend hiccup left a permanent hole in the index.

This PR makes that queue drain itself, and makes a heal restore everything the failed fetch would have produced.

## What changed

**Automatic retry.** `InternalDataErrorRetryLoop` runs a pass shortly after startup — picking up failures from the previous run while the backend cache is still warm — and hourly thereafter. The underlying routine runs at most once at a time and is a no-op on an empty queue, so the periodic trigger is free when there is nothing to do.

**Retry cap, with an operator override.** The automatic loop keeps the existing `maxNumberOfRetires` cap so it can never hammer a permanently unfetchable block. The admin **Retry fetch** button now clears the retry count instead, so a block that exhausted its budget can be revived deliberately rather than staying abandoned forever.

**Reconnecting is now idempotent.** `internalDataAlreadyStored` skips transactions whose stored internal data already match the recomputed data (or are empty). The check happens *inside* the reconnect, under the same block-write lock as the counter-incrementing write, so reconnecting a block more than once cannot double-count its contract transaction counters. Addresses are compared as descriptors, tolerating casing (raw trace vs EIP55) and format (Tron base58 vs hex), and treating an empty address as the zero address the packed form stores.

**A heal now registers contracts too.** Contracts created or destroyed in a block are a product of its internal-data fetch — so when that fetch failed, `storeBlockSpecificDataEthereumType` stored none, and the reconnect stored internal data only. A contract created in such a block stayed unresolvable through the API even after the block healed. `BackfillContractInfo` closes that: it merges the lifecycle fields into the registry rather than overwriting, so it cannot clobber name/symbol/standard enrichment fetched on demand after sync, and it records a destruction even when the creation was never indexed. It takes `connectBlockMux` because its read-modify-write races the live sync's block connect.

**Smaller cleanups.** `getBlockRetryInternalData` and `storeBlockInternalDataError` dedup the fetch-retry and enqueue paths that were previously inlined. On the Tron side, rejected internal transactions are detected in the existing pass instead of a second loop over the same slice.

## Scope: what this does *not* do

This heals blocks that **failed** an internal-data fetch. It does **not** backfill blocks synced before `processInternalTransactions` was enabled at all — those never entered the error queue, so nothing knows they are incomplete.

Earlier revisions of this PR did that too, via a `-healinternaldata` flag that seeded a watermark at the tip and swept downward to genesis. **That mechanism has been removed.** It was built for Tron, which had synced ~85M blocks with the option off, and in production it sustained ~2.3 blocks/s — putting a full sweep at 10–14 months. Cutting the per-block backend work and parallelising the fetches still only brought it to weeks, so it did not solve the problem it was written for. A reindex is the realistic path for that history, and the flag, its persisted watermark (`InternalDataFrom`) and the startup resolution around it are gone rather than left as a permanently-on-by-flag mechanism nobody should use.

Reviewers who saw the earlier version: commits `feat(eth)!: remove the -healinternaldata backfill sweep` and `fix(eth): register contracts when healing from the internal data error queue` are where that reversal happens. The net diff no longer contains the sweep.

## Testing

- `go test -tags unittest ./db/ ./api/ ./common/ ./bchain/coins/tron/` — all pass.
- New coverage: `TestRocksDB_BackfillContractInfo` (merge semantics, destruction-before-creation, enrichment preserved) and `db/rocksdb_internaldata_test.go` (`Test_emptyInternalData`, `Test_internalDataEqual` — the address casing/format/zero-address equivalences the idempotency check relies on).
- Not verified here: a full `go build ./...` (exceeds the local cgo/rocksdb build budget) and the end-to-end behaviour against a live Tron node — the retry loop's first pass, the admin button's cap reset, and a healed block's contract becoming resolvable.


api/internaldata_backfill.go
```diff
@@ -0,0 +1,148 @@
...
+// storeBackfilledContracts persists the contract lifecycle discovered during the
+// backfill (creations and destructions); ReconnectInternalDataToBlockEthereumType
+// does not store them. Because the sweep runs downward, a contract's destruction is
+// visited before its creation, so BackfillContractInfo records a destruction even
+// when no registry row exists yet and merges the creation height into it later,
+// preserving any name/symbol/standard enrichment fetched on demand after sync.
+func (w *Worker) storeBackfilledContracts(blockSpecificData *bchain.EthereumBlockSpecificData) {
+	for i := range blockSpecificData.Contracts {
+		ci := &blockSpecificData.Contracts[i]
+		if err := w.db.BackfillContractInfo(ci); err != nil {
```
pragmaxim: **Contract-lifecycle write errors are silently dropped and never retried.** `storeBackfilledContracts` only `glog`s a `BackfillContractInfo` failure — it returns nothing — and `backfillBlockInternalData` returns `nil` right after calling it, so the sweep advances the watermark (`SetInternalDataFrom`) and treats the block as healed. A failed lifecycle write (e.g. a transient RocksDB read/write error) is therefore lost permanently: the block isn't enqueued into the error queue either, and once the watermark passes it there's no way to re-seed.

This diverges from normal sync, where `storeBlockSpecificDataEthereumType` propagates `storeContractInfo` errors (`db/rocksdb_ethereumtype.go:1261`), failing the block connect so it gets retried. Suggest returning the first `BackfillContractInfo` error from `storeBackfilledContracts` and propagating it out of `backfillBlockInternalData`, so the height is enqueued/retried instead of silently losing the creation/destruction record.

api/internaldata_backfill.go
```diff
@@ -0,0 +1,148 @@
...
+			continue
+		}
+		if hash == "" {
+			// no block indexed at this height: nothing to heal, but count it so a
+			// gappy index (a run of unindexed blocks) is not silently skipped
+			missing++
+			if missing == 1 {
+				glog.Warningf("internalDataHealing: no block indexed at height %d; the index has gaps in the healed range, missing blocks are skipped", height)
+			}
+		} else if err := w.backfillBlockInternalData(height, hash); err != nil {
```
pragmaxim: **No throttle between successful heals.** On the happy path the loop goes straight from `backfillBlockInternalData` to `h = height` and iterates again, so a full tip→genesis sweep issues millions of `chain.GetBlock` (trace) calls back-to-back. The only pauses are the 10s sleeps on `!synced` / `GetBlockHash` error. The `!synced` check does provide crude backpressure (the sweep parks if the node falls behind), but under normal conditions this is continuous heavy trace load on the backend for the entire duration of the sweep. Consider a small inter-block delay or a configurable rate limit so the backfill can't monopolize the backend.

db/rocksdb_ethereumtype.go
```diff
@@ -762,17 +859,23 @@ func (d *RocksDB) ReconnectInternalDataToBlockEthereumType(block *bchain.Block)
...
-			tx.BlockHeight = block.Height
-			if err = d.processInternalData(blockTx, tx, eid.InternalData, addresses, addressContracts, true); err != nil {
-				return err
-			}
+		// skip transactions whose internal data are already stored (or empty): this
+		// block may be reconnected more than once - the healing sweep and the
+		// error-queue refetch can both reach it - and reprocessing would count its
+		// contract transactions again. The check shares connectBlockMux with the
+		// write below, so a concurrent reconnect cannot race between them.
+		if eid.InternalData == nil || d.internalDataAlreadyStored(tx.Txid, eid.InternalData) {
```
pragmaxim: **The double-count-prevention path has no DB-integration test.** This skip guard is load-bearing: on reprocess, `InternalTxs` / `NonContractTxs` and per-contract `Txs` are incremented unconditionally (only `TotalTxs` is deduped via the pre-loaded address map), so a false negative in `internalDataAlreadyStored` / `internalDataEqual` would permanently double-count a block's counters. Yet `ReconnectInternalDataToBlockEthereumType` appears in zero `_test.go` files — only the pure helpers (`internalDataEqual`, `emptyInternalData`) and `BackfillContractInfo` are covered. Worth an integration test that reconnects the same block twice and asserts the address/contract counters are unchanged on the second call (and that a block synced-before-flag gains the expected internal counters exactly once).

db/rocksdb_ethereumtype.go
```diff
@@ -960,6 +1063,54 @@ func (d *RocksDB) GetEthereumInternalData(txid string) (*bchain.EthereumInternal
...
+	bestHeight, _, err := d.GetBestBlock()
+	if err != nil {
+		glog.Error("ResolveInternalDataFrom: GetBestBlock: ", err)
+		return
+	}
+	if bestHeight == 0 {
+		return
+	}
+	if healRequested {
+		d.is.SetInternalDataFrom(bestHeight + 1)
```
pragmaxim: **The seed watermark isn't persisted synchronously.** It's written only to the in-memory `InternalState` here; persistence relies on `storeInternalStateLoop` (`storeInternalStatePeriodMs` ≈ 59.7s). If the process dies within that first window after seeding, the watermark is lost. On the next restart *without* `-healinternaldata`, `ResolveInternalDataFrom` finds no watermark and doesn't reseed — and because blocks synced with `processInternalTransactions` on will have made `HasAnyEthereumInternalData` return true, the advisory branch is skipped too, so history silently stays unhealed with no signal to the operator. Narrow window, but consider persisting the internal state synchronously right after seeding (here, or in `blockbook.go` right after the `ResolveInternalDataFrom` call).

pragmaxim (commented):
Follow-up on the earlier review — verified these findings against the code and posting them inline. (A fifth observation — `internalDataErrorRetryLoop` checks `common.IsInShutdown()` only after its sleep, which can be up to an hour — is real but benign: `waitForSignalAndShutdown` never waits on that goroutine and it correctly gates new work after waking, so I've left it out as a nit.)

cranycrane:
Replaced by #1705

PR #1694, opened 2026-07-31
feat(metrics): make sync state alertable

pragmaxim:
## Motivation

A Tron instance recently ran desynced for around seventeen hours and nothing alerted. Reconstructing it showed the metrics were the problem, not the alert rules:

- the process was crash-looping, but it was up at every scrape, so a target-down rule could not see it;
- `blockbook_synchronized` was 0 the whole time, but it is only written by `blockbookAppInfoMetric`, which needs three backend RPCs and therefore runs at most every ~15 minutes;
- `blockbook_best_height` was frozen while `blockbook_backend_best_height` kept advancing — the backend was healthy, only the indexer had stopped — and nothing exported a signal that distinguishes those two states;
- `blockbook_avg_block_period` was 0, because it is recomputed from block times and resets on restart. Both `blockbookBackendStuck` and `blockbookBackendStuckFast` were gated on `blockbook_avg_block_period > 0`, so **the guard switched those rules off in exactly the condition that needed them.**

That last point generalises: the observed block period collapses to 0 on restart and during a reindex, which are precisely when a stall matters.

This branch fixes the metric surface so a stalled indexer is expressible as an alert. The rules themselves live in **trezor/bbctl#20**, which consumes everything added here.

## What changed

| Commit | Change |
|---|---|
| `feat(metrics): add initial_sync gauge` | new `blockbook_initial_sync`, so a rule on `blockbook_synchronized` can exclude a planned reindex |
| `fix(metrics): stop zeroing backend height when the backend errors` | both callers built `BackendInfo` from an empty `ChainInfo` on a failed `GetChainInfo`, wiping the cached payload |
| `fix(metrics): make tip age survive a restart and a rollback` | persist the tip-advance reference; compare against the persisted height rather than the in-memory one |
| `fix(metrics): refresh sync gauges without waiting for the app-info loop` | new `api.RefreshSyncMetrics`, RPC-free, called from the sync loop and a ~60s timer |
| `feat(metrics): add mempool_last_sync_timestamp_seconds` | mempool reload freshness, restart-safe |
| `fix(metrics): report the observed block period in fractional seconds` | `blockbook_avg_block_period` no longer truncates sub-second chains to 0 |
| `fix(metrics): label index resync errors by where they came from` | `error` label was the constant `"failure"` at all five sites |
| `fix(configs): set averageBlockTimeMs for the Bitcoin test networks` | they exported `blockbook_average_block_time_seconds = 0` |
| `fix(metrics): report not-in-sync while the backend cannot be queried` | superseded by the two commits below; see behaviour change 5 for what the final state actually does |
| `fix: floor the EVM freshness window, then correct three stale block times` | see behaviour changes 1 and 2 |
| `style: cap the comments added by this branch at five lines` | comment-only |
| `fix(configs): correct BSC/Polygon block times and signet indentation` | superseded for both coins by the two commits below |
| `fix(metrics): harden InternalState averages and sync-state snapshot` | wrap-safe `computeAvgBlockPeriod`; one locked snapshot for `RefreshSyncMetrics` |
| `fix(metrics): make mempool_last_sync absent until the first reload` | `gauge_vec`, so `time() - it` cannot page before the first reload |
| `fix(metrics): single-writer sync gauges and generalize freshness check` | `blockbook_synchronized` had two writers computing it from different tips |
| `fix(configs): keep BSC averageBlockTimeMs at 450ms` | measured value, not the post-Maxwell nominal |
| `fix(metrics): keep the stale-index demotion EVM-only, correct Polygon cadence` | review fixes; see behaviour change 1 |

### New metrics

- `blockbook_initial_sync` — 1 while `is.InitialSync`, else 0. Written at the two points the flag flips, on a ~30s ticker for the duration of the initial build, and on every later refresh.
- `blockbook_mempool_last_sync_timestamp_seconds` — unix time of the last **successful** mempool reload, shaped as a timestamp for the same reasons as `blockbook_alternative_fee_provider_last_sync_timestamp_seconds`, and a `gauge_vec` so the series is absent until that first reload.

### Cadence

`blockbook_synchronized`, `blockbook_best_height` and `blockbook_backend_best_height` were written only by the ~15-minute app-info loop. `api.RefreshSyncMetrics` publishes them from internal state and cached backend info with **no backend round trip**, and is called

- on every `storeInternalStateLoop` tick (~60s) — the guaranteed freshness floor,
- after every `syncIndexLoop` iteration, via `defer` so the early returns are covered (the interesting case is the quiet one, where `resyncIndex` keeps returning `syncNotNeeded`),
- on a ~30s ticker for the duration of the initial build, since neither loop above is running yet,
- at the two points `is.InitialSync` flips,
- from the app-info loop, which no longer computes `blockbook_synchronized` itself.

The `inSync` value comes from the same `systemInfoInSync` the `/api/` response uses, and both normalise by one shared `syncBlockPeriod` helper, so the gauge and the API cannot drift apart.

## Behaviour changes reviewers should weigh

**1. `systemInfoInSync` freshness window: a 30s floor, and the demotion stays EVM-scoped.**

The window was derived from the chain cadence as 12 block times with no lower bound — 3s on Arbitrum, 1.2s on Robinhood, under ordinary RPC and GC jitter. `eth.TipStaleThreshold` already clamps its analogous window to `[30s, 5m]`; this mirrors that bound. Every EVM coin's window stays the same or gets **looser**, except BSC below.

The two directions of the check are scoped differently:

- the **tip rescue** (out-of-sync → synced when the index is within one block of a fresh tip) now applies to any chain with a configured cadence. The raw flag is cleared for the whole of every resync iteration, so without this a chain that takes longer than the 5s start grace to connect a batch reads out-of-sync mid-iteration even though its index is at the tip. This direction only ever relaxes the reported state.
- the **demotion** (synced → out-of-sync when the index has not advanced within the window) stays **EVM-only**, as on master. An intermediate commit dropped that gate; the last commit restores it. EVM tips arrive over a `newHeads` subscription that can die silently and their spacing is near-deterministic, so 12 block times floored at 30s has large headroom. UTXO block arrival is Poisson: at 12 mean spacings a healthy chain crosses the window ~6e-6 of the time per block, and Bitcoin's configured 600s against an observed ~682s makes the real window tighter than nominal. `blockbookApiInSync` (severity `high`) probes the `/api/` field this computes across 17 UTXO endpoints — 5 Bitcoin, 2 each of Bcash/Dogecoin/Litecoin/Zcash/testnet4, plus 2 testnet3, which this branch newly gives a configured cadence and which is the most erratic chain in the fleet. Ungated, that is a handful of spurious high-severity pages a year. A stalled UTXO indexer is instead caught by `BlockbookIndexStalled` (`blockbook_best_height` against `blockbook_backend_best_height`), which is chain-agnostic and reads neither flag.

**2. Three `averageBlockTimeMs` values corrected to observed mainnet spacing.**

Measured over 14 days from `blockbook_backend_best_height`, ignoring restarts and rollbacks, and re-confirmed against 3h/24h rates at review time:

| coin | configured | observed |
|---|---|---|
| BNB Smart Chain | 3000ms | 450ms |
| Avalanche | 2000ms | ~1000ms |
| Polygon | 2000ms | 1500ms |

The 6.7× error on BSC matters beyond the metric: a rule that normalises expected block count by this value needs roughly a 33× slowdown before BSC trips it. Resulting deltas, with the floor applied:

- *freshness window*: Arbitrum 3s→30s, Robinhood 1.2s→30s, Base/Optimism/Avalanche/Polygon 24s→30s (looser); Tron 36s and Ethereum 144s unchanged; **BSC 36s→30s — the only one that narrows**, and at 0.45s blocks 30s is still ~67 blocks of slack against Ethereum's 12.
- *tip watchdog window*: BSC 90s→30s, Avalanche 60s→30s, Polygon 60s→45s. The watchdog polls and reconnects sooner on those three, which is the intended behaviour at their real cadence, and stays inside the existing `[30s, 5m]` clamp.

**3. `blockbook_avg_block_period` changes value.** It is fed from a fractional-seconds average instead of the truncated integer one, so sub-second chains (Arbitrum, BNB Smart Chain, Robinhood) go from a permanent 0 to their real period. No deployed rule reads it, so this is a Grafana-only change in practice; the rules that were gated on it are re-based onto `blockbook_average_block_time_seconds` in trezor/bbctl#20. `is.AvgBlockPeriod` keeps feeding the sync ETA in integer seconds, with one deliberate change: `computeAvgBlockPeriod` now returns early instead of computing a `uint32` subtraction that could wrap to ~4.29e9 when header timestamps are non-monotonic or `RemoveLastBlockTimes` truncates the array, so both the ETA and the gauge hold their previous value across such a window rather than spiking.

**4. `blockbook_index_resync_errors{error}` label values change** from the constant `"failure"` to `get_block_hash`, `get_block`, `connect_block` and `resync`. The in-repo dashboard panel already groups `by (error)`, so it becomes readable with no dashboard change, and no deployed rule selects `error="failure"`.

**5. On a backend error, the cached `BackendInfo` is retained instead of being overwritten with zeros.** `SetBackendInfo` records the error on top of the last known good payload; the guard needs both a reported error and no height, so a backend answering at genesis is still stored normally. The visible effect is on **metrics**: `blockbook_backend_best_height` holds its last value instead of dropping to 0, and `blockbook_app_info` keeps its `backend_*` labels instead of blanking the series for the one instance an operator is trying to find. `/api/` and websocket `getInfo` are unchanged — they still report the live `GetChainInfo` result, so a failed query still yields zeros with `error` set. That is deliberate: the API reports what the backend says right now, the metrics report the last thing it said. The zero-drop mattered because `max_over_time − min_over_time` over 90m turns it into an enormous span, which satisfied `BlockbookIndexStalled`'s right-hand side while `delta(blockbook_best_height) <= 0` was trivially true — so a *backend* outage used to page as "Blockbook index stalled".

**6. `InternalState` gains two persisted fields**, `backendTipLastAdvance` and `backendTipHeight`, and loses one, `initialSync`. It is stored as plain JSON (`Pack` is `json.Marshal`), so this is compatible in both directions: an older database simply has the new keys absent and seeds them on first use, and an older binary ignores them. `initialSync` moves to `json:"-"` because a SIGTERM during an initial build would otherwise store `true`, and a subsequent start without `--sync` never clears it — latching `blockbook_initial_sync` at 1 and permanently gating off any alert that excludes reindexes. Nothing reads the persisted value, and the `initialSync` field of `/api/` comes from `api.BlockbookInfo`, so neither the API nor `blockbook-api.ts` changes.

## Alerting this enables

Implemented in **trezor/bbctl#20**; that PR carries the 30-day backtests behind each threshold.

| Rule | Depends on |
|---|---|
| `BlockbookIndexStalled` | `blockbook_best_height` vs `blockbook_backend_best_height`, both at ≤60s cadence, which lets its window tighten 90m→45m. Detects the failure in the motivation; stays silent through a reindex because the index height is climbing. |
| `BlockbookRestarting` | `process_start_time_seconds` — already exported by the Go client, listed for completeness. |
| `BlockbookBackendStuck` / `BlockbookBackendStuckSlowChain` | re-based on `blockbook_average_block_time_seconds` instead of `blockbook_avg_block_period`, so the rule no longer disables itself on restart or during a reindex. Behaviour changes 2 and the testnet commit make that value correct for every deployed coin. |
| `BlockbookNotSynchronized` (new) | `blockbook_synchronized` gated on `blockbook_initial_sync == 0`. The `and`-shaped gate keeps it inert until this PR deploys, rather than firing through every reindex on an older binary. |
| `BlockbookMempoolReloadStale` (new) | `blockbook_mempool_last_sync_timestamp_seconds`. Absent until the first successful reload, so it needs no reindex gate of its own. |
| `blockbookApiInSyncMempool` | annotated for retirement. The flag behind it is cleared at the *start* of every mempool reload, so it reports "reload in progress" rather than health and reads 0 for a large fraction of samples on a healthy node. |
| `blockbookApiSyncMode` | recommended for deletion. `is.SyncMode` is a static echo of the `--sync` flag, reset when the database is loaded; it cannot represent a runtime condition. |

## Testing

- `go build ./...`, `go vet ./...` clean; `gofmt` clean over every file this branch touches.
- Full unit suite green: 2227 tests across 69 packages (`go test -tags unittest ./...`).
- `contrib/scripts/render_grafana.py --check` passes (105 panels, 89 metrics; the renderer validates every `{{name:…}}` reference, panel ids, grid bounds and overlaps).
- New unit tests: `SetBackendInfo` error/genesis handling; tip-advance persistence across a `Pack`/`Unpack` round trip and on rollback; `computeAvgBlockPeriod` sub-second, insufficient-samples and interval-count cases; `RefreshSyncMetrics` for the synced, stalled, reindexing, unobserved-backend and backend-error cases; and `systemInfoInSync` rows for the freshness floor, the non-EVM tip rescue, the non-EVM demotion exemption and the no-configured-cadence fallback.
- New Grafana panel for `mempool_last_sync_timestamp_seconds` (`time() - it`, with the alert cutoff drawn), matching how its `alternative_fee_provider` sibling is presented.


pragmaxim:
CI status: **Lint** and **Unit Tests** pass. **Connectivity Tests** fail, and the failure is unrelated to this branch.

The only failing subtest is `TestIntegration/ethereum=main/connectivity`, which cannot reach its dev instance (`connect: connection refused` — the service is not listening). It reproduces identically on a re-run, so it is environmental rather than flaky. The same job passed on `master` earlier today, and this branch touches no Ethereum coin config and nothing under `tests/` (21 files changed, see the diff), so it cannot affect whether a remote dev instance accepts connections.

Happy to re-run once that instance is back up.

configs/coins/polygon.json
```diff
@@ -58,7 +58,7 @@
             "mempool_sub_workers": 2,
             "block_addresses_to_keep": 300,
             "additional_params": {
-                "averageBlockTimeMs": 2000,
+                "averageBlockTimeMs": 1500,
```
cranycrane, line 61: This should be 1750 - https://en.bloomingbit.io/feed/news/111602

cranycrane (commented):
## Code review — `feat(metrics): make sync state alertable`

15 findings from a review of this branch at `cf3ec86`, ranked most-severe first. `go vet` is clean and the new tests pass; the changed symbols were traced to their callers and to `configs/grafana/panels.yaml`.

The dominant theme is that the new gauges will **page falsely rather than catch the stall they were added for**. Four independent paths each produce a `0` or a stale timestamp while the index is healthy: the mempool gauge sitting at 0 before the first resync (1), two writers of `blockbook_synchronized` disagreeing on the tip height (2), the 60 s tick sampling mid-resync `false` (3), and the cached-`BackendError` stickiness (4). Conversely the metrics are **blind during the initial sync** (7) — both republishing loops start only after the first `ResyncIndex` returns.

One item is a decision rather than a fix: the `averageBlockTimeMs` reductions (5) are not metric-only — they feed `eth.TipStaleThreshold`, so BSC's watchdog stall threshold drops 90 s → the 30 s floor and its poll interval 30 s → 10 s.

Findings 1-12, 14 and 15 are inline. Finding **13** has no line in this diff, so it is here:

---

**13. Two new metrics have no Grafana panel, and the `synchronized` panel is not gated on `initial_sync`** &nbsp;·&nbsp; `process` — `configs/grafana/panels.yaml`

`AGENTS.md` (*Metrics*, lines 71-81) and `configs/grafana/README.md` make `panels.yaml` + `template.json` the source of truth, and document adding the panel plus re-rendering as part of adding a metric. Neither file is touched by this branch.

So `blockbook_initial_sync` and `blockbook_mempool_last_sync_timestamp_seconds` exist in the binary and in `metrics.yaml` but appear on no dashboard, and `general.synchronized` still plots bare `{{name:synchronized}}{job="blockbook"}` — also missing the `coin="$coin"` filter every neighbouring panel has — rather than the `synchronized == 0 unless initial_sync == 1` form the new help text tells operators to use.

---

<sub>Reviewed with Claude Code at `xhigh` effort. Every finding was adversarially verified against the code before reporting; the conditional in (5) depends on BSC's real post-Maxwell cadence, which is outside the repo.</sub>

bchain/coins/blockchain.go
```diff
@@ -469,6 +469,7 @@ func (c *mempoolWithMetrics) Resync() (count int, err error) {
 	}).Observe(throughput)
 	if err == nil {
 		c.m.MempoolSize.Set(float64(count))
+		c.m.MempoolLastSync.Set(float64(time.Now().Unix()))
```
cranycrane: **1. `mempool_last_sync_timestamp_seconds` is 0 until the first resync, so the prescribed alert pages from process start** &nbsp;·&nbsp; `correctness`

`blockbook_mempool_last_sync_timestamp_seconds` is a plain gauge that exists at 0 from process start, so the alert its own help text prescribes (`time() - metric`) fires continuously until the first successful mempool resync.

**Failure scenario:** Fresh start or reindex: `mempool.Resync()` is first called at `blockbook.go:385`, i.e. only after `ResyncIndex(nil, true)` returns (hours on a UTXO chain, days on an EVM archive). Until then the gauge is 0, so `time() - blockbook_mempool_last_sync_timestamp_seconds` is ~1.75e9 s and the documented rule ("a value climbing past a few multiples of `resyncmempoolperiod` means the reload loop is failing") pages for the whole window.

The sibling metric `alternative_fee_provider_last_sync_timestamp_seconds` avoids this by being a `gauge_vec` (series absent until first success), and this same PR deliberately withholds `backend_best_height` until observed (`api/worker.go:2740`) — this gauge is the one place the pattern was not applied.

api/worker.go
```diff
@@ -2687,25 +2696,68 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
...
+	initialSync := is.GetInitialSync()
+	inSync, bestHeight, lastBlockTime, startSync := is.GetSyncState()
+	if bi.BackendError != "" {
+		// GetSystemInfo reports inSync=false whenever GetChainInfo fails, so the gauge must
+		// too or the two disagree for exactly the outage this metric exists for. The cached
+		// height is the last known good one, so systemInfoInSync would otherwise read "at
+		// the tip and fresh" against a dead backend. Cleared by the next SetBackendInfo.
+		inSync = false
+	} else {
+		inSync = systemInfoInSync(inSync, initialSync, chainType, bestHeight, bi.Blocks,
```
cranycrane: **2. Two writers publish `blockbook_synchronized` from different tip heights, so it oscillates on a ~15-minute period** &nbsp;·&nbsp; `correctness`

`RefreshSyncMetrics` and `blockbookAppInfoMetric` both write `blockbook_synchronized`, but from different inputs — cached `bi.Blocks` vs a live `GetChainInfo` — so the gauge oscillates on a ~15-minute period: the exact disagreement the doc comment claims to prevent.

**Failure scenario:** Sub-second EVM chain, resync iteration in flight (`IsSynchronized=false`):

- The 60 s refresh uses `bi.Blocks` cached by the last `updateBackendInfo()`, which equals the indexed height, so `gap == 0` and it publishes **1**.
- The 15-minute app-info tick calls `GetSystemInfo`, which does a live `GetChainInfo` whose tip is several blocks ahead of the last connected block, so `gap > systemInfoEthereumSyncedGap` and it publishes **0**.

Same instant, same state, opposite values — whichever writer ran last wins until the other runs. The comment at `api/worker.go:2699-2702` ("so the /api/ inSync field and the blockbook_synchronized gauge cannot disagree") is not achieved, because only the `blockPeriod` input was unified, not the tip height.

blockbook.go
```diff
@@ -708,6 +734,12 @@ func storeInternalStateLoop() {
 		glog.Info("storeInternalStateLoop starting with db stats compute disabled")
 	}
 	common.TickAndDebounce(storeInternalStatePeriodMs*time.Millisecond, (storeInternalStatePeriodMs-1)*time.Millisecond, chanStoreInternalState, func() {
+		// Republish the sync-state gauges every tick (~60s) - the guaranteed freshness
+		// floor. syncIndexLoop also refreshes them, but a blockbook whose tip feed went
+		// quiet only wakes on the resyncindexperiod backstop (~15.6 min), which is how far
+		// blockbook_synchronized used to lag reality. RefreshSyncMetrics reads internal
+		// state only, no backend RPCs, so it is cheap to run unconditionally.
+		refreshSyncMetrics()
```
cranycrane, line 774: **3. Publishing from the 60 s tick samples the transiently-false resync state, so `synchronized` now dips to 0 on healthy chains** &nbsp;·&nbsp; `correctness`

Publishing the sync gauges from the ~60 s `storeInternalStateLoop` tick samples a state that is transiently false for the whole duration of every resync iteration, so `blockbook_synchronized` now dips to 0 on healthy busy chains.

**Failure scenario:** `ResyncIndex` calls `is.StartedSync()` (`db/sync.go:209`), which sets `IsSynchronized=false` for the entire iteration. `systemInfoInSync` only forgives the first `systemInfoSyncStartGrace` (5 s), and for `ChainBitcoinType` there is no tip-equality rescue at all (per the new test *does not extend tip equality rescue to bitcoin*).

Any iteration lasting >5 s — a UTXO chain connecting a batch after a gap, an EVM archive with heavy blocks — that overlaps the 60 s tick publishes `synchronized=0` while the index is keeping up; the `syncIndexLoop` `defer` then restores 1. Previously this state was sampled once per ~15 min, so exposure is ~15x higher now, and the dips are indistinguishable from the real stall the metric was added to catch.

api/worker.go
```diff
@@ -2687,25 +2696,68 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
...
+// stall stayed invisible for up to 16 minutes. backend_best_height is held back until a
+// height has been observed, since a 0 reads as "no blocks produced" to the stuck rules.
+func RefreshSyncMetrics(is *common.InternalState, chain bchain.BlockChain, chainType bchain.ChainType, metrics *common.Metrics) {
+	if is == nil || metrics == nil {
+		return
+	}
+	bi := is.GetBackendInfo()
+	initialSync := is.GetInitialSync()
+	inSync, bestHeight, lastBlockTime, startSync := is.GetSyncState()
+	if bi.BackendError != "" {
```
cranycrane: **4. Forcing 0 on the *cached* `BackendError` makes the gauge sticky for up to a full resync period after recovery** &nbsp;·&nbsp; `correctness`

Forcing `synchronized=0` on the cached `BackendInfo.BackendError` makes the gauge sticky for up to a full resync period after the backend recovered, and inverts the intended agreement with `/api/` (which re-queries live).

**Failure scenario:** `SetBackendInfo` is only called by `updateBackendInfo()` at the end of a `ResyncIndex` iteration (`db/sync.go:214`). If `resyncIndex` succeeds (`GetBestBlockHash` fine) but `GetChainInfo` fails, `ResyncIndex` returns nil, so `syncIndexLoop` does **not** take its 2.5 s retry path — the error is cached with no re-probe until the next iteration. On a chain whose push feed is quiet (precisely the stall case) that is the `resyncindexperiod` backstop, ~15.6 min.

For that whole window every 60 s tick publishes `synchronized=0` while `/api/` does a live `GetChainInfo` and reports `inSync=true` — the two disagree in the *opposite* direction from the one the comment describes, and a "sustained 0" alert can fire off a single transient RPC blip.

configs/coins/bsc.json
```diff
@@ -58,7 +58,7 @@
             "mempool_sub_workers": 2,
             "block_addresses_to_keep": 300,
             "additional_params": {
-                "averageBlockTimeMs": 3000,
+                "averageBlockTimeMs": 450,
```
cranycrane, line 61: **5. Lowering `averageBlockTimeMs` is not metric-only — it tightens the tip watchdog and forces ~3x more reconnects** &nbsp;·&nbsp; `config`

Lowering `averageBlockTimeMs` is not metric-only: it directly tightens `eth.TipStaleThreshold`, which drives the tip watchdog's stall threshold and its forced `newHeads` reconnect, plus the erc4626 negative-cache TTL and the Grafana threshold overlay.

**Failure scenario:** `TipStaleThreshold()` (`bchain/coins/eth/ethrpc.go:1109`) is `clamp(30 x avg, 30s, 5min)` and `tipWatchdog` samples at `threshold/3` (min 5 s):

| coin | change | stall threshold | poll interval |
|---|---|---|---|
| BSC | 3000→450 ms | 90 s → **30 s floor** | 30 s → 10 s |
| Avalanche | 2000→1000 ms | 60 s → 30 s | 20 s → 10 s |
| Polygon | 2000→1500 ms | 60 s → 45 s | 20 s → 15 s |

On a rate-limited public RPC whose feed goes quiet for 30-60 s (normal), the watchdog now declares a stall and force-reconnects the subscription roughly 3x as often, producing reconnect churn and `watchdog_stall` events where there was none. It also shortens `erc4626NegativeProbeTTLBlocks` (`api/erc4626.go:98`) and retargets the `sync.tip_feed_age` threshold series (`configs/grafana/panels.yaml:270`).

If 450 ms is not the measured BSC cadence (Maxwell put it at 0.75 s), the watchdog is now permanently over-tight.

blockbook.go
```diff
@@ -541,15 +554,22 @@ func blockbookAppInfoMetric(db *db.RocksDB, chain bchain.BlockChain, txCache *db
...
-		"backend_version":          si.Backend.Version,
-		"backend_subversion":       subversion,
-		"backend_protocol_version": si.Backend.ProtocolVersion}).Set(float64(0))
-	metrics.BackendBestHeight.Set(float64(si.Backend.Blocks))
+	// GetSystemInfo substitutes an empty ChainInfo when the backend cannot be queried, so
+	// every Backend field of si is zero then. Publishing it would blank the app-info
+	// version labels and drive blockbook_backend_best_height to 0, which the stuck rules
+	// read as "no blocks produced". Keep the last published values; the locally derived
+	// gauges below are still refreshed.
+	if si.Backend.BackendError == "" {
```
cranycrane: **6. The `BackendError` guard withholds the whole `app_info` series, including the local version labels** &nbsp;·&nbsp; `observability`

This guard withholds the entire `blockbook_app_info` series — including the purely local blockbook version/commit/buildtime labels — whenever the backend is unreachable, so the app_info panel is empty on exactly the instance you need to identify.

**Failure scenario:** Blockbook starts with its backend down: `blockbookAppInfoMetric` (called at `blockbook.go:341` and then every ~15 min) never executes `BlockbookAppInfo.With(...)`, so the series `blockbook_app_info` is never created at all. `configs/grafana/panels.yaml:549` plots `{{name:app_info}}` and renders empty, and any `count by (blockbook_version)` inventory query loses the instance. Before the change the series existed with empty `backend_*` labels.

The four `blockbook_*` labels come from `common.GetVersionInfo()` and need no backend; only `backend_version`/`subversion`/`protocol_version` and the height needed guarding — and those could now be sourced from `is.GetBackendInfo()`, which this PR made retain the last known good payload.

blockbook.go
```diff
@@ -362,7 +362,8 @@ func mainWithExitCode() int {
 
 	if *synchronize {
 		internalState.SyncMode = true
-		internalState.InitialSync = true
+		internalState.SetInitialSync(true)
+		refreshSyncMetrics()
```
cranycrane, line 366: **7. This is the only sample taken during the entire initial sync — both republishing loops start after it** &nbsp;·&nbsp; `correctness`

`refreshSyncMetrics` never runs during the initial sync, contradicting its own doc comment: both republishing loops are started only after `ResyncIndex(nil, true)` returns.

**Failure scenario:** The comment at `blockbook.go:531-534` claims the gauges are refreshed "on every storeInternalStateLoop tick (~60s), after every sync iteration". But `go syncIndexLoop()` (line 390) and `go storeInternalStateLoop()` (line 395) both start *after* the initial `ResyncIndex` at line 367 has completed.

During the initial build/reindex — the window `blockbook_initial_sync` exists to describe, hours to days — the only sample is this single call, taken before any block was connected and before the first `SetBackendInfo`, so `backend_best_height` is never published (`bi.Blocks == 0` hits the guard) and `synchronized`/`initial_sync` are one frozen sample. An initial sync that stalls partway is invisible on every gauge this PR adds except `blockbook_best_height` (which `db/sync.go` still writes per block).

api/worker.go
```diff
@@ -2687,25 +2696,68 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
...
+	synchronized := 0.0
+	if inSync {
+		synchronized = 1
+	}
+	metrics.Synchronized.Set(synchronized)
+	initial := 0.0
+	if initialSync {
+		initial = 1
+	}
+	metrics.InitialSync.Set(initial)
```
cranycrane, line 2781: **8. `initial_sync` is backed by a persisted flag that can latch at 1 and permanently gate off the alert** &nbsp;·&nbsp; `correctness`

`blockbook_initial_sync` is backed by the persisted `InitialSync` field, which can latch at 1 and permanently suppress the alert the metric was added to gate.

**Failure scenario:** `RocksDB.Close()` (`db/rocksdb.go:474`) persists internal state on graceful shutdown, so a SIGTERM during the initial sync stores `initialSync: true` (`json:"initialSync"`, `common/internalstate.go:71`). Restarting the same DB **without** `-sync` — a query-only replica, `-blockfrom` bulk mode — never reaches `SetInitialSync(false)` at `blockbook.go:392`, so `GetInitialSync()` stays true forever.

The documented rule `synchronized == 0 unless initial_sync == 1` is then permanently gated off, and `server/public.go:450` keeps the internal explorer disabled. Nothing resets the flag except a `-sync` run that completes.

common/internalstate.go
```diff
@@ -396,14 +421,29 @@ func (is *InternalState) GetAvgBlockPeriod() uint32 {
...
+// intervals by avgBlockPeriodSample (~1% high) and is left alone so the ETA is unchanged.
 func (is *InternalState) computeAvgBlockPeriod() {
 	last := len(is.BlockTimes) - 1
 	first := last - avgBlockPeriodSample - 1
 	if first < 0 {
 		return
 	}
-	is.AvgBlockPeriod = (is.BlockTimes[last] - is.BlockTimes[first]) / avgBlockPeriodSample
+	span := is.BlockTimes[last] - is.BlockTimes[first]
+	is.AvgBlockPeriod = span / avgBlockPeriodSample
```
cranycrane, line 475: **9. Unguarded uint32 subtraction wraps on non-monotonic block times, spiking the newly documented gauge to ~4.25e7 s** &nbsp;·&nbsp; `correctness`

The uint32 subtraction feeding both averages is unguarded, so a non-monotonic 101-block window wraps and the newly documented `blockbook_avg_block_period` gauge spikes to ~4.25e7 seconds.

**Failure scenario:** `span := is.BlockTimes[last] - is.BlockTimes[first]` on uint32: if `BlockTimes[last] < BlockTimes[first]` — header timestamps are not guaranteed monotonic (`GetBlockHeightOfTime` at line 402 documents "the block times can sometimes be out of order"), and `RemoveLastBlockTimes` re-runs the computation on a truncated array after a disconnect — span wraps to ~4.29e9.

`AvgBlockPeriodSeconds` becomes ~4.25e7, published straight onto the gauge the yaml now tells operators to use for normalization, and `AvgBlockPeriod` feeds `etaSeconds = int64(etaBlocks * w.is.AvgBlockPeriod)` (`api/worker.go:333`), which overflows into a nonsense ETA on the explorer.

`if is.BlockTimes[last] <= is.BlockTimes[first] { return }` or int64 arithmetic fixes both.

api/worker.go
```diff
@@ -2687,25 +2696,68 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
...
-	// covers sub-second chains (e.g. Arbitrum at 250ms) that round to 0 seconds.
-	// Fall back to the runtime-observed average when the coin does not configure one.
-	if p, ok := w.chain.(interface {
+// syncBlockPeriod returns the block cadence the sync freshness checks normalize by.
+// It prefers the configured averageBlockTimeMs - stable, available before enough blocks
+// are observed, sub-second capable, and the same value the tip watchdog uses - and falls
+// back to the observed average. Shared by GetSystemInfo and RefreshSyncMetrics so the
+// /api/ inSync field and the blockbook_synchronized gauge cannot disagree.
+func syncBlockPeriod(chain bchain.BlockChain, is *common.InternalState) time.Duration {
+	blockPeriod := time.Duration(is.GetAvgBlockPeriod()) * time.Second
```
cranycrane: **10. Fallback still uses the truncating `GetAvgBlockPeriod()`; a 0 here silently disables every EVM sync check** &nbsp;·&nbsp; `correctness`

`syncBlockPeriod` still falls back to the truncating integer `GetAvgBlockPeriod()` — the very truncation this PR added `GetAvgBlockPeriodSeconds()` to fix. A 0 here disables every EVM sync check silently.

**Failure scenario:** For a `ChainEthereumType` coin whose RPC does not expose `AverageBlockTimeDuration` (or returns an error — e.g. a chain reached through `blockChainWithMetrics` wrapping a non-EVM backend, `bchain/coins/blockchain.go:527`), `blockPeriod` becomes `time.Duration(is.GetAvgBlockPeriod()) * time.Second`, which truncates a 250 ms cadence to **0**.

`systemInfoInSync` then returns at `blockPeriod <= 0`, skipping both the freshness check and the `gap <= 1` tip rescue, so `blockbook_synchronized` silently degrades to the raw `IsSynchronized` flag on exactly the sub-second chains this PR targets. `common/internalstate.go:446` already computes the correct fractional value under the same lock; this is the only remaining consumer of the truncating form besides the sync ETA.

api/worker.go
```diff
@@ -2687,25 +2696,68 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
...
+// RefreshSyncMetrics publishes the sync-state gauges from internal state and the cached
+// backend info with no backend round trip, so it can run on a short timer. These gauges
+// used to be written only by the ~15-minute app-info loop, which needs three RPCs, so a
+// stall stayed invisible for up to 16 minutes. backend_best_height is held back until a
+// height has been observed, since a 0 reads as "no blocks produced" to the stuck rules.
+func RefreshSyncMetrics(is *common.InternalState, chain bchain.BlockChain, chainType bchain.ChainType, metrics *common.Metrics) {
+	if is == nil || metrics == nil {
+		return
+	}
+	bi := is.GetBackendInfo()
```
cranycrane: **11. Three independent locked snapshots can tear, mixing a pre-resync backend height with a post-resync indexed height** &nbsp;·&nbsp; `correctness`

`RefreshSyncMetrics` builds its decision from three independent locked snapshots, so it can mix a pre-resync backend height with a post-resync indexed height and flip `synchronized` for a sample.

**Failure scenario:** `GetBackendInfo()`, `GetInitialSync()` and `GetSyncState()` each take and release `is.mux` separately. A concurrent `updateBackendInfo()`/`FinishedSync()` can land between them, so `gap = int64(bi.Blocks) - int64(bestHeight)` is computed from a torn pair:

- `bi.Blocks` from before a catch-up + `bestHeight` from after → negative gap, rescue rejected by `gap >= 0`, publishes 0 for a synced index.
- The reverse → inflated gap.

A single locked snapshot accessor, or reusing `GetSyncState`'s pattern to return all four values at once, removes it.

db/rocksdb.go
```diff
@@ -746,9 +746,11 @@ func (d *RocksDB) ConnectBlock(block *bchain.Block) error {
 	if err := d.WriteBatch(wb); err != nil {
 		return err
 	}
-	avg := d.is.SetBlockTime(block.Height, uint32(block.Time))
+	d.is.SetBlockTime(block.Height, uint32(block.Time))
```
cranycrane: **12. Dropped the last consumers of the `SetBlockTime*` return values, and now takes `is.mux` twice per connected block** &nbsp;·&nbsp; `efficiency`

The diff dropped the last consumers of `SetBlockTime`/`SetBlockTimes`' return value but left the returns and their doc comments in place, and replaced one lock acquisition per connected block with two on the hottest write path.

**Failure scenario:** `SetBlockTime` (`common/internalstate.go:367`, *"returns AvgBlockPeriod"*) and `SetBlockTimes` (line 352) now have no caller that uses the result — the only call sites are `db/rocksdb.go:749` and `:2302`, both of which discard it and immediately re-enter `GetAvgBlockPeriodSeconds()`, taking `is.mux` a second time for every single connected block.

On a 250 ms chain that is a second uncontended-but-real lock round trip per block on the same mutex `UpdateBestHeight` also takes, plus a value that can have changed between the two calls. Returning the float (or both) from `SetBlockTime` keeps one acquisition and removes the now-misleading doc comments.

api/worker_test.go
```diff
@@ -530,3 +555,117 @@ func fanInVout(n uint32, scriptHex string) bchain.Vout {
...
 
 func fanInHash(n uint64) string { return fmt.Sprintf("%064x", n) }
+
+var (
+	refreshMetricsOnce sync.Once
+	refreshMetrics     *common.Metrics
+	refreshMetricsErr  error
+)
+
+func getRefreshTestMetrics(t *testing.T) *common.Metrics {
```
cranycrane: **14. Shared memoized `Metrics` makes the four new tests order-dependent, and registers on the global registry** &nbsp;·&nbsp; `test-quality`

The new tests share one package-level `Metrics` across four functions with order-dependent gauge state, and register ~60 collectors on the global `prometheus.DefaultRegisterer` instead of a private test registry.

**Failure scenario:** `getRefreshTestMetrics` memoizes one `*common.Metrics` for all four tests. `TestRefreshSyncMetrics` leaves `Synchronized=0` and `InitialSync=1` behind; `TestRefreshSyncMetricsBackendError` then asserts `Synchronized == 1` after only its own call, which holds solely because nothing interleaves. Under `go test -shuffle=on` or a future `t.Parallel()` these give false failures.

`common/metrics_test.go:17` has `useTestPrometheusRegistry` for exactly this reason (swap `DefaultRegisterer` for a private registry, restore in `Cleanup`), but it is unexported, so the api tests permanently register every Blockbook collector on the process-wide default registry.

configs/coins/bitcoin_signet.json
```diff
@@ -64,7 +64,9 @@
             "xpub_magic_segwit_p2sh": 71979618,
             "xpub_magic_segwit_native": 73342198,
             "slip44": 1,
-            "additional_params": {}
+            "additional_params": {
+            	"averageBlockTimeMs": 600000
```
cranycrane: **15. Literal TAB indentation — the only tab in any file under `configs/coins/`** &nbsp;·&nbsp; `style`

This added `averageBlockTimeMs` line is indented with a literal TAB — the only tab in any file under `configs/coins/`.

**Failure scenario:** `grep -rlP "\t" configs/coins/*.json` matches exactly this one file. Valid JSON, but the config tree is machine-processed by `build/templates` and every other coin file uses 4 spaces, so the file diverges from the generator's canonical formatting and any reformat produces spurious churn.

pragmaxim:
## Review addressed

Thanks for the thorough pass — all 15 findings are addressed. `go build ./...`, `go vet ./...`, the full `-tags unittest` suite, and `contrib/scripts/render_grafana.py --check` are all green.

| # | Finding | Fix | Commit |
|---|---|---|---|
| 1 | `mempool_last_sync` reads 0 from process start | now a `gauge_vec[chain]` — series **absent until the first successful reload** | `174aef15` |
| 2 | two writers publish `synchronized` from different tips | `RefreshSyncMetrics` is the **sole writer**; app-info just calls it | `7674d042` |
| 3 | 60s tick samples the transient resync-false state | tip/freshness **rescue generalized to any chain with a configured cadence** (not just EVM); the "does not extend to bitcoin" case now asserts synced | `7674d042` |
| 4 | sticky `0` on the cached `BackendError` | force removed; gauge derives from the snapshot, outages are caught by `tip_age`/backend-stuck; test rewritten | `7674d042` |
| 5 | lowering `averageBlockTimeMs` tightens the watchdog | **BSC kept at 450ms** (observed spacing — see thread); **Polygon 1500→1750ms**, archive+non | `3d1b180a`, `8eca3c6b` |
| 6 | `app_info` series withheld when backend down | always published; backend labels fall back to last-good `BackendInfo` | `7674d042` |
| 7 | no refresh during the initial sync | ~30s ticker during the initial `ResyncIndex`; comment corrected | `7674d042` |
| 8 | persisted `InitialSync` can latch at 1 | field is now `json:"-"` (runtime-only) | `a16ce364` |
| 9 | uint32 wrap on non-monotonic block times | guarded (`BlockTimes[last] <= BlockTimes[first]` → keep last good) | `a16ce364` |
| 10 | fallback used truncating `GetAvgBlockPeriod()` | uses fractional `GetAvgBlockPeriodSeconds()` | `7674d042` |
| 11 | three locked snapshots can tear | single `GetSyncMetricsSnapshot()` | `a16ce364` + `7674d042` |
| 12 | dead `SetBlockTime*` returns + double lock/block | return the fractional average, one `is.mux` per block | `a16ce364` |
| 14 | shared memoized `Metrics` on global registry | fresh `Metrics` on a private registry per test | `7674d042` |
| 15 | literal TAB in `bitcoin_signet.json` | replaced with 4 spaces | `3d1b180a` |

Notes:
- **#3** was implemented as extend-the-rescue (index at tip + recently synced ⇒ synced) rather than left as a Bitcoin-specific exclusion; `systemInfoInSync` no longer branches on chain type and the `Ethereum*` constants were renamed.
- **#5** — BSC stays at **450ms**, the observed mainnet spacing (details in the thread); at 0.45s blocks `TipStaleThreshold` = `clamp(30×0.45, 30s, 5min)` hits the **30s floor**, so the watchdog window is unchanged by the exact value. Only **Polygon** is corrected (1500→1750ms → tip window 60s → ~52.5s, inside the `[30s, 5min]` clamp). The "behaviour changes" table in the description should reflect the Polygon change.

configs/coins/bsc.json
pragmaxim: Keeping BSC at **450ms**. Good catch on the side effects (tip watchdog / erc4626 TTL / Grafana overlay), but 450ms is the observed BSC mainnet spacing the maintainers measure, not a transcription of the nominal Maxwell target — so the watchdog tightening it implies is intended. At 0.45s blocks `TipStaleThreshold` = clamp(30×0.45, 30s, 5min) hits the **30s floor** (same as it would at 0.75s), so the watchdog window itself is unchanged by the exact value; the value mainly matters for the stuck-rule normalisation. Reverted the 750ms change in `8eca3c6b`. The Polygon correction (1500→**1750ms**) stays.

cranycrane (commented):
The diagnosis here is excellent and most of the implementation is sound — several changes fix genuine latent bugs beyond the stated scope (the `InitialSync` → `json:"-"` change in particular: a SIGTERM mid-initial-build used to persist `initialSync=true`, which never cleared and permanently disabled the internal explorer via `server/public.go:450`).

Verified locally at `8eca3c6b`: `go build ./...`, `go vet ./...`, `go test -tags unittest ./common/... ./api/... ./db/...`, `-race`, `-shuffle=on`, and `render_grafana.py --check` all pass. The testing claims in the description hold.

Three findings below. **H1 I'd consider blocking**; H2 and H3 leave the repo self-inconsistent.

---

## H1 (blocking) — dropping the chain-type gate extends the *demotion* branch to UTXO chains, causing false out-of-sync on healthy nodes

`api/worker.go:2678` — master gated on `chainType != bchain.ChainEthereumType || blockPeriod <= 0`; this PR keeps only `blockPeriod <= 0`.

Three things compound:

1. **BTC chains satisfy the interface.** `BitcoinRPC.AverageBlockTimeDuration` exists at `bchain/coins/btc/bitcoinrpc.go:53` and is forwarded by `blockChainWithMetrics` (`bchain/coins/blockchain.go:523`). Every BTC-derived altcoin inherits it.

2. **`lastBlockTime` is not a block time.** `GetSyncState` returns `is.LastSync` — the wall-clock time Blockbook last *connected a block*, not a header timestamp. On a quiet chain the path is `syncNotNeeded → FinishedSyncNoChange()`, which by design **does not touch `LastSync`** (`common/internalstate.go:235-239`). So `isFresh` really means "we connected a block recently", which on a UTXO chain is a property of the chain, not of the indexer.

3. **The demotion branch never consults heights.** `if inSync && !isFresh { return false }` (`api/worker.go:2698`) fires with the index byte-for-byte at the backend tip.

**Failure scenario:** healthy Bitcoin node, index at tip, `IsSynchronized=true`. The chain produces no block for >2h (`12 × 600s`). `/api/` returns `inSync=false` and `blockbook_synchronized` drops to 0 — on *every* Bitcoin instance simultaneously, since it's a chain property, so there is no healthy peer to fail over to. Rough Poisson rates for coins that already carried `averageBlockTimeMs` on master and were previously excluded by the chain-type gate: BTC ~0.3 events/yr (2h window), Litecoin ~1.3/yr (30m), Zcash ~2.6/yr (15m), Dogecoin ~3.2/yr (12m).

It's sharper for the three configs **this PR adds** — `bitcoin_testnet`, `bitcoin_testnet4`, `bitcoin_signet` at `averageBlockTimeMs: 600000`. Testnet3/4 block production is bursty (min-difficulty rule, intermittent miners) and signet depends on a single signer, so multi-hour gaps are routine rather than exceptional. Those instances will report out-of-sync a large fraction of the time while completely healthy. Note `tests/openapi/src/context.ts:245` hard-aborts the OpenAPI suite on `inSync=false` unless `OPENAPI_ALLOW_OUT_OF_SYNC=1`.

**The doc comment is also inaccurate.** `api/worker.go:2663-2668` says the path applies to "any chain with a known block period (a configured `averageBlockTimeMs`)" and that "a chain with no configured cadence (`blockPeriod <= 0`) keeps the raw flag unchanged." Neither holds: `syncBlockPeriod` seeds `blockPeriod` from `is.GetAvgBlockPeriodSeconds()` first (`api/worker.go:2714`), which is non-zero for any coin with ≥102 indexed block times — `setBlockTimes` runs at DB open. So the ~69 coin configs with no `averageBlockTimeMs` at all are equally subject, at whatever their observed 101-block average happens to be.

The rationale given in the comment ("the raw flag is cleared for the whole of every resync iteration…") motivates only the **rescue** branch at `:2694`, which could have been generalized to non-EVM on its own. It does not motivate the **demotion** branch, which is the newly harmful one.

Suggested fix — gate the demotion on the index actually being behind:

```go
if inSync && !isFresh && int64(backendBlocks)-int64(bestHeight) > systemInfoSyncedGap {
    return false
}
```

"The chain is quiet" is already covered by `blockbook_tip_age_seconds`.

Related: the parity claim at `api/worker.go:2650-2654` is half-true — `eth.TipStaleThreshold` (`bchain/coins/eth/ethrpc.go:1109-1121`) clamps at **both** ends (30s floor, 5min cap); `systemInfoInSync` only floors. And the same comment says the floor "lets a fast chain's configured cadence be corrected downwards without tightening this check" — for BSC the check *is* tightened, 36s → 30s, because the floor sits below the old window.

---

## H2 — `docs/sync.md` is now factually wrong, and the tip watchdog gets materially more aggressive on BSC/Avalanche

`docs/sync.md:140` states: *"Per-chain values: Polygon/Optimism/Base/Avalanche 60 s, BSC/Tron 90 s"* and *"on Polygon's 2 s blocks"*. With `TipStaleThreshold = clamp(30 × avg, 30s, 5min)` (`bchain/coins/eth/ethrpc.go:1109-1127`) and the new config values:

| coin | avg ms | `TipStaleThreshold` | sample interval `clamp(thr/3, 5s, 60s)` |
|---|---|---|---|
| `bsc`, `bsc_archive` | 3000 → **450** | 90s → **30s** (floor; raw 13.5s) | 30s → **10s** |
| `avalanche(_archive)` | 2000 → **1000** | 60s → **30s** | 20s → **10s** |
| `polygon(_archive)` | 2000 → **1750** | 60s → **52.5s** | 20s → **17.5s** |

Beyond the stale doc, this is a real behaviour change worth calling out in the description alongside the freshness-window table. `tipWatchdogTick` (`ethrpc.go:1150-1200`) does a `refreshBestHeaderFromChain(true)` **plus a full `reconnectRPC()`** on a stall, and all three coins use a live `newHeads` subscription (`configs/coins/{bsc,polygon,avalanche}.json`). Net: 3× more watchdog evaluations per minute on BSC and Avalanche, and any 30-second feed hiccup behind a load balancer now forces a tip poll + RPC reconnect where 90s (BSC) / 60s (Avalanche) was needed before.

`docs/sync.md:143` ("BTC-family coins do not use this cached-tip feed and are unaffected") is also now misleading given H1 — they're unaffected by the *watchdog*, but no longer by `systemInfoInSync`.

---

## H3 — `averageBlockTimeMs: 450` for BSC looks wrong, and it feeds the exact gauge the new alerts normalize by

`configs/coins/bsc.json:61`, `configs/coins/bsc_archive.json:62`.

BSC's documented mainnet cadence after the **Maxwell** hardfork (June 2025) is **0.75s** — 3s → 1.5s at Lorentz (Apr 2025) → 0.75s at Maxwell. I'm not aware of any BSC mainnet release putting block time at 450ms; the sub-second roadmap step after Maxwell targets ~150ms, not 450ms. Since the branch carries a commit titled *"keep BSC averageBlockTimeMs at 450ms"*, this was clearly considered — could you cite the measurement?

The impact cuts both ways:

- **Inert for both watchdogs.** `30 × 0.45 = 13.5s` and `30 × 0.75 = 22.5s` both clamp to the 30s floor; `12 × 0.45` and `12 × 0.75` both clamp to `systemInfoMinStale`. So 450 vs 750 changes nothing for the code paths this PR is about.
- **Not inert for `blockbook_average_block_time_seconds`.** That gauge is precisely what the description proposes re-basing `blockbookBackendStuck` / `blockbookBackendStuckFast` onto. The documented rule `blockbook_tip_age_seconds > 30 * blockbook_average_block_time_seconds` (`docs/sync.md:186`) then fires at **13.5s ≈ 18 real blocks** on BSC instead of the intended 30 — ~40% tighter than designed, on the one chain whose value moved 6.7×.
- Minor secondary: `api/erc4626.go:88-99` converts the 15-minute negative-cache TTL (`erc4626.go:28`) into a block count, so an understated block time stretches the effective TTL ~1.67× in wall-clock.

I couldn't verify Avalanche 1000ms / Polygon 1750ms to a citable standard — flagging only that Polygon is commonly documented at ~2.1s, so 1750 looks on the aggressive side.

---

I have a longer list of non-blocking items (a reproduced ~1/2000 timing flake in `TestBackendTipAdvanceOnRollback`, the `RefreshSyncMetrics` tests passing `chain = nil` so they never reach the freshness path they're meant to cover, `blockbook_backend_best_height` still exporting 0 at cold start despite the `> 0` guard, and the `metrics.yaml` help strings growing the HELP payload from 15.2 KB to 18.7 KB per scrape) — happy to post those separately if useful.

cranycrane (approved):

PR #1695, opened 2026-07-31
feat(eth): make ENS reverse aliases opt-in per chain

cranycrane:
## Summary

Adds `enable_ens_reverse_aliases`, a per-chain opt-in for **ENS reverse aliasing** (address → ENS-name labels). It is **off by default on every chain**, including Ethereum mainnet — no coin config opts in, so `configs/` is unchanged by this PR. Contract/token-name aliases and the forward `ResolveENS` path are untouched. Also documents the reverse path as not production-ready and consolidates all ENS code into one file.

## Why

The reverse labels are an event projection of `NameRegistered` logs with no notion of expiry or ownership transfer, so a name lingers on its original owner address indefinitely and can appear on several addresses at once (stale/duplicate labels) — a display-trust hazard. Nothing restricts the emitter either: any contract can emit a `NameRegistered`-shaped event, so the labels are spoofable. The live, expiry-correct **forward** path (`ResolveENS`/`CheckENSExpiration`) is the supported name→address route and is unaffected. ENS is also mainnet-only, so labels recorded on other EVM chains came from arbitrary `NameRegistered`-shaped events anyway.

It is an opt-**in** rather than an opt-out because a chain must not inherit this behaviour by *forgetting* a config line — which had already happened: `hyperevm_archive.json` set `address_aliases` with no opt-out and kept recording labels. With the polarity inverted, a new EVM chain is safe by construction and the 14 per-chain config lines are unnecessary.

## What changed

**Per-chain opt-in (`enable_ens_reverse_aliases`)**
- **Write:** `processEventsForBlock` skips `getEnsRecord` unless the chain opted in → nothing new is written to `cfAddressAliases`.
- **Read:** the API worker keeps the `Contract`/token-name alias (separate `cfContracts` path) and skips only the `GetAddressAlias` ENS lookup → existing labels stop being served **without a reindex**.
- `UseEnsReverseAliases()` is `address_aliases && enable_ens_reverse_aliases`, so `address_aliases` can stay `true` and contract/token-name annotations are preserved.
- `Configuration.ApplyToParser` is the single place config reaches the parser; `NewTronRPC` calls it too, since `NewTronParser` discards the parser `NewEthereumRPC` configured. `UseEnsReverseAliases()` is declared on `bchain.BlockChainParser` (false on `BaseParser`), like `UseAddressAliases`, so no coin-specific type assertion is needed in the api layer.
- `cfAddressAliases` is left dormant in the schema — **no RocksDB migration**; enabling is a one-line config flip per chain.
- **Enabling later does not backfill.** Labels are only recorded as blocks are synced, so a chain that opts in afterwards shows no label for names registered while the flag was off, until it is reindexed. Noted on the config field and on `UseEnsReverseAliases`.

**Docs:** a brief `WARNING` on `getEnsRecord` — the reverse approach is not production-ready (no trusted-emitter allowlist, no name validation, no expiry/ownership tracking) and is disabled by default.

**Refactor:** all ENS code (constants, `getEnsRecord`, `SetEnsSuffix`, `UseEnsReverseAliases`, and the forward path `ensNameHash`/`keccak256`/`parseENSAddressFromResult`/`ensContracts`/`ResolveENS`/`CheckENSExpiration`) moved into a single `bchain/coins/eth/ens.go`. Pure move, no behavior change.

## Preserved / unaffected

- Contract & token names in `addressAliases` responses ✅
- Forward `foo.eth → address` resolution and live expiry check ✅
- No reindex or schema migration needed to turn the labels off ✅

## Testing

`go build`, `go vet` and `gofmt` clean. Both gates and every config→parser→worker hop are now covered, and each test was verified by mutation (breaking the code it guards makes it fail):

- `TestGetAddressAliases_EnsReverseGate` — read path over a real RocksDB holding both alias kinds: closing the gate drops only the ENS label, never the contract name. Asserting the whole map also pins *where* the `continue` sits, so hoisting it above the contract lookup fails.
- `TestProcessEventsForBlock_EnsReverseGate` — write path: a `NameRegistered` log yields no `AddressAliasRecord` when gated, while the tx logs themselves are returned either way.
- `TestNewWorker_ReadsEnsReverseOptInFromParser` — the parser→worker hop, pinning the `address7b.eth` wire format.
- `TestNewEthereumRPC_PropagatesEnsReverseOptIn` / `TestNewTronRPC_PropagatesEnsReverseOptIn` — config→parser wiring, including opt-in without `address_aliases`.
- `TestTronParser_UseEnsReverseAliases` — the predicate truth table.
- `Test_PublicServer_EthereumType` opts in, keeping the alias-rendering path covered rather than dropping the expectations.

Plus the pre-existing eth ENS tests (`Test_getEnsRecord`, `TestENSFunctionSelectors`, `TestCheckENSExpiration_ParsesPaddedResult`).

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -63,9 +63,9 @@ func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool, 
 				xpubCfg.MaxCacheEntries, xpubCfg.MaxCacheExpirationSeconds, xpubCfg.DefaultAddressesGap, xpubCfg.MaxAddressesGap)
 		}
 	}
-	// default enabled for parsers that don't expose the per-chain toggle
+	// default enabled for non-EthereumType parsers, which have no ENS aliases
 	useEnsReverseAliases := true
-	if p, ok := chain.GetChainParser().(interface{ UseEnsReverseAliases() bool }); ok {
+	if p, ok := chain.GetChainParser().(eth.EthereumLikeParser); ok {
 		useEnsReverseAliases = p.UseEnsReverseAliases()
 	}
 	w := &Worker{
--- a/api/worker.go
+++ b/api/worker.go
@@ -63,12 +63,6 @@ func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool, 
 				xpubCfg.MaxCacheEntries, xpubCfg.MaxCacheExpirationSeconds, xpubCfg.DefaultAddressesGap, xpubCfg.MaxAddressesGap)
 		}
 	}
-	// opt-in per chain; the fallback stays off so a parser that predates the flag
-	// never serves ENS labels (only read for EthereumType, where all parsers implement it)
-	useEnsReverseAliases := false
-	if p, ok := chain.GetChainParser().(eth.EthereumLikeParser); ok {
-		useEnsReverseAliases = p.UseEnsReverseAliases()
-	}
 	w := &Worker{
 		db:                   db,
 		txCache:              txCache,
@@ -76,7 +70,7 @@ func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool, 
 		chainParser:          chain.GetChainParser(),
 		chainType:            chain.GetChainParser().GetChainType(),
 		useAddressAliases:    chain.GetChainParser().UseAddressAliases(),
-		useEnsReverseAliases: useEnsReverseAliases,
+		useEnsReverseAliases: chain.GetChainParser().UseEnsReverseAliases(),
 		mempool:              mempool,
 		is:                   is,
 		fiatRates:            fiatRates,
--- a/api/worker.go
+++ b/api/worker.go
@@ -30,11 +30,13 @@ type Worker struct {
 	chainParser       bchain.BlockChainParser
 	chainType         bchain.ChainType
 	useAddressAliases bool
-	mempool           bchain.Mempool
-	is                *common.InternalState
-	fiatRates         *fiat.FiatRates
-	metrics           *common.Metrics
-	xpubConfig        XpubConfig
+	// useEnsReverseAliases gates serving ENS reverse labels (EthereumType only).
+	useEnsReverseAliases bool
+	mempool              bchain.Mempool
+	is                   *common.InternalState
+	fiatRates            *fiat.FiatRates
+	metrics              *common.Metrics
+	xpubConfig           XpubConfig
 }
 
 var getTickersForTimestamps = func(fr *fiat.FiatRates, timestamps []int64, vsCurrency string, token string) (*[]*common.CurrencyRatesTicker, error) {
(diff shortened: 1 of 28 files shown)


pragmaxim (approved):
LGTM

bchain/coins/eth/ethrpc.go
```diff
@@ -1421,12 +1416,15 @@ func (b *EthereumRPC) processEventsForBlock(blockNumber string) (map[string][]*b
 		return nil, nil, errors.Annotatef(err, "%s blockNumber %v", method, blockNumber)
 	}
 	r := make(map[string][]*bchain.RpcLog)
+	recordEns := b.Parser.UseEnsReverseAliases()
```
pragmaxim, line 1433: **[medium] Enabling the flag later doesn't restore the off-window — and nothing says so.**

Gating the write path here means blocks indexed while `enable_ens_reverse_aliases` is off never get their `NameRegistered` events projected into `cfAddressAliases`, and there is no backfill path. So the "enabling is a one-line config flip" in the PR description is only forward-looking: a chain that opts in later (or toggles off and back on) permanently serves a gap — names registered during the off-window show no label, mixed with correct labels on either side, until a full reindex.

Since the flag is presented as a per-chain runtime opt-in, it'd be worth a sentence on the `EnableEnsReverseAliases` config comment (or the `getEnsRecord` WARNING, or a startup log when the flag flips on over an existing DB) so operators know enabling does not recover labels for blocks synced while it was off.

api/worker.go
```diff
@@ -61,18 +63,25 @@ func NewWorker(db *db.RocksDB, chain bchain.BlockChain, mempool bchain.Mempool,
 				xpubCfg.MaxCacheEntries, xpubCfg.MaxCacheExpirationSeconds, xpubCfg.DefaultAddressesGap, xpubCfg.MaxAddressesGap)
 		}
 	}
+	// opt-in per chain; the fallback stays off so a parser that predates the flag
+	// never serves ENS labels (only read for EthereumType, where all parsers implement it)
+	useEnsReverseAliases := false
+	if p, ok := chain.GetChainParser().(eth.EthereumLikeParser); ok {
```
pragmaxim: **[medium] Coin-specific type assertion diverges from the generic parser-flag mechanism used one line below.**

This introduces a second, parallel plumbing mechanism next to the sibling flag: `UseAddressAliases()` is declared on `bchain.BlockChainParser` with a `BaseParser` default, while this one requires the chain-agnostic api layer to know about `eth.EthereumLikeParser`. The `ok == false` fallback fails safe (as the description notes) but also fails *silent*: any EthereumType parser that doesn't implement `eth.EthereumLikeParser` (a wrapping/decorating parser, a future EVM-like coin with its own parser type) has its opt-in swallowed — the write path records reverse labels the worker never serves, with no error or log.

Declaring `UseEnsReverseAliases() bool` on `bchain.BlockChainParser` with a `false` default on `BaseParser` would remove the assertion, the fallback comment, and the cached worker field (`w.chainParser.UseEnsReverseAliases()` at the call site suffices), matching how `useAddressAliases` already works.

bchain/coins/tron/tronrpc.go
```diff
@@ -143,6 +143,9 @@ func NewTronRPC(config json.RawMessage, pushHandler func(bchain.NotificationType
 	tronRpc.Parser.AddrContractsCacheMinSize = ethChainConfig.AddressContractsCacheMinSize
 	tronRpc.Parser.AddrContractsCacheMaxBytes = ethChainConfig.AddressContractsCacheMaxBytes
 	tronRpc.Parser.AddrContractsCacheBulkMaxBytes = ethChainConfig.AddressContractsCacheBulkMaxBytes
+	// NewTronParser recreates the parser, so re-apply the ENS reverse opt-in that
+	// NewEthereumRPC set on the parser it built (otherwise Tron ignores the flag).
+	tronRpc.Parser.EnableEnsReverseAliases = ethChainConfig.EnableEnsReverseAliases
```
pragmaxim: **[medium] This extends the pattern that caused the original regression instead of fixing it.**

`NewTronParser` discards the parser `NewEthereumRPC` already configured, so every config-derived `EthereumParser` field must be manually re-applied here — this line makes it 7. Nothing (compiler or test) forces this list to stay in sync with `NewEthereumRPC`, so the next config-derived parser field added on the eth side will again be silently dropped for Tron — the exact class of bug this PR fixes.

Worth fixing the pattern rather than the instance: either have `NewTronParser` start from the already-configured `EthereumParser` (copy the struct, then override the Tron-specific `AmountDecimalPoint`/`FormatAddressFunc`/`FromDescToAddressFunc`/`EnsSuffix`), or extract the config→parser application in `ethrpc.go` (~L269–280) into a shared helper both constructors call, so there is a single point of addition.

pragmaxim (commented):
Review findings (critical/high/medium only):

pragmaxim (approved):
LGTM

PR #1700, opened 2026-08-03
feat: fetch ERC-20 balances via Multicall3 (per-chain address for Tron)

### LINKED ISSUE #1683: eth: fetch ERC-20 balances via Multicall3 aggregate3 (with per-chain Multicall3 address for Tron)
## Summary

ERC-20 balance lookups for an address currently issue **one `eth_call` (`balanceOf`) per token contract**, delivered as a JSON-RPC batch. On metered RPC providers this costs `per_call_credits × N` (e.g. 20 × N on QuickNode's Ethereum tier) and executes N independent calls that are not block-consistent.

We already have a Multicall3 `aggregate3` primitive — `EthereumTypeMulticallAggregate3` (`bchain/coins/eth/multicall.go`), used today only by ERC-4626 enrichment (`api/erc4626.go`). Routing balances through it collapses N sub-calls into a **single `eth_call`** (~20 credits flat, all at one block).

## Current behavior

- `EthereumTypeGetErc20ContractBalancesAtBlock` (`bchain/coins/eth/contract.go`) → `erc20BalancesBatchAtBlock` builds `[]rpc.BatchElem` of `eth_call`/`balanceOf` and sends via `BatchCallContext`.
- A JSON-RPC batch is **1 HTTP request but N billable method calls**; batching saves round-trips, not credits or rate-limit budget. Each sub-call also runs at its own `latest`, so results are not guaranteed block-consistent.

## Proposed change

**1. Route balances through `aggregate3` with batch fallback.**
In `EthereumTypeGetErc20ContractBalancesAtBlock`, try Multicall3 first (all `balanceOf(holder)` calls with `AllowFailure=true`), decode `(success, bytes)[]` → balances. Fall back to the existing JSON-RPC batch path when Multicall3 isn't available (`errMulticall3NotDeployed`) or on aggregate error. Preserve the current nil-on-failure semantics for dead / non-conforming tokens.

**2. Make the probed Multicall3 address per-chain.**
Today `multicall3Address` is a single hardcoded canonical constant (`0xcA11bde05977b3631167028862bE2a173976CA11`, `multicall.go:18`). Tron's Multicall3 is deployed at a **non-canonical** address, so the canonical-only probe misses it.

## Multicall3 availability across supported EVM chains (verified 2026-07-29)

| Chain | Chain ID | Multicall3 | At canonical `0xcA11…CA11`? | Path after change |
|---|---|---|---|---|
| Ethereum, ETC, Optimism, BSC, Polygon, Base, Arbitrum One/Nova, Avalanche, Sepolia, Hoodi | — | ✅ | ✅ | `aggregate3` |
| Robinhood mainnet | 4663 | ✅ | ✅ (verified via robinhoodchain.blockscout.com) | `aggregate3` |
| Tron mainnet | 728126428 | ✅ | ❌ `0x32a4f47a74a6810bd0bf861cabab99656a75de9e` (`TEazPvZ…`) | `aggregate3` **after** per-chain address override |
| Robinhood testnet | 46630 | ❌ | — | `eth_call` batch (unchanged) |
| Tron Nile testnet | 3448148188 | ❌ | — | `eth_call` batch (unchanged) |

## Tron override

Add a per-chain override so **Tron mainnet only** (728126428) probes/uses `0x32a4f47a74a6810bd0bf861cabab99656a75de9e`. Tron Nile and every other chain keep the canonical default. Chains without a Multicall3 deployment (incl. Robinhood testnet, Tron Nile) continue to fall back to the `eth_call` batch — this is intended and requires no work.

## Impact

- **Cost:** on metered providers, ERC-20 balance credit cost drops from ~`20 × N` to ~`20` (flat) on chains with Multicall3.
- **Consistency:** all balances observed at a single block.
- **Safety:** no behavior change on chains without Multicall3 (batch fallback preserved); `AllowFailure=true` keeps dead/non-conforming tokens producing `nil` rather than failing the whole request.

## References

- Multicall3 canonical deployments: https://www.multicall3.com/deployments · https://github.com/mds1/multicall3
- Tron Multicall3: https://tronscan.org/#/contract/TEazPvZwDjDtFeJupyo7QunvnrnUjPH8ED
- Robinhood mainnet Multicall3: https://robinhoodchain.blockscout.com/address/0xcA11bde05977b3631167028862bE2a173976CA11
- Existing primitive: `EthereumTypeMulticallAggregate3` (`bchain/coins/eth/multicall.go`), consumer `api/erc4626.go`

cranycrane:
Closes #1683.

Routes ERC-20 balance lookups through Multicall3 `aggregate3`: one billable `eth_call` per chunk instead of N (a JSON-RPC batch is 1 HTTP request but N metered method calls). Every balance within a chunk is pinned to the same block; chunks are separate `eth_call`s, so a >100-contract request can still span blocks when no explicit block number is passed (unchanged from the batch path, which chunks at the same size). Falls back per-chunk to the existing JSON-RPC batch path on any failure (chain without Multicall3, transient error, decode/count mismatch), preserving the same-length/order and nil-on-failure contract callers rely on.

Adds a code-set per-chain override (`EthereumRPC.Multicall3AddressOverride`, empty = canonical) for chains that deploy Multicall3 at a non-canonical address, and wires Tron mainnet's (`0x32a4f47a…`) in `NewTronRPC`. A deployment address is not derivable from the chain, so it is a property of the coin and set in code.

**Observability**
- `eth_call_requests{mode="multicall"}` — new: aggregate3 sub-call volume (the fast path), so the credit savings are measurable. aggregate3 calls were previously counted as `mode="single"`.
- `eth_call_multicall_requests` — new counter (added to `configs/metrics.yaml`) for *physical* aggregate3 requests, i.e. the billable count. The `mode` series needs no `metrics.yaml` change since `mode` is a dynamic label.
- `rpc_fallback_calls{component="erc20_multicall", reason="error"}` — unexpected fallbacks to the JSON-RPC batch (the expected "Multicall3 not deployed" case is silent).
- The deployment probe logs once per process whether Multicall3 was found on the chain.

**Dashboards will shift on deploy**: the `mode="multicall"` reclassification moves the existing erc4626 enrichment's aggregate3 calls from `single` → `multicall`, and `eth_call_batch_size` now records aggregate3 chunk sizes alongside JSON-RPC batch sizes. Update any alerts on those series before rollout.

## Verification

Unit tests cover the address resolver/override, aggregate3 encode/decode (including malformed-response rejection), chunking, every fallback tier, the starved-tail rule, and the fallback metrics. `go build ./...`, `go vet ./...` and `go test ./...` pass, except two packages whose tests bind a local TCP port — blocked by the sandbox used here, unrelated to this change.

Beyond unit tests, the change was exercised against live nodes on both affected chain families.

**Tron mainnet** (java-tron JSON-RPC, chain id 728126428, block ~85.09M) — the override path:
- The canonical address `0xcA11…CA11` has no code on Tron, confirming an override is required.
- The deployed bytecode at `0x32a4f47a…` (base58 `TEazPvZwDjDtFeJupyo7QunvnrnUjPH8ED`) contains all 16 Multicall3 selectors, including `aggregate3` (`0x82ad56cb`). The base58 ↔ hex mapping was checked independently.
- `aggregate3` round-trips through this PR's encoder/decoder: correct `balanceOf`, `decimals()`=6 and `symbol()`="USDT" for TRC-20 USDT.
- 130 TRC-20 contracts for one holder resolved in 2 aggregate3 chunks (100 + 30) in ~310 ms. All 130 balances cross-checked against independent single `eth_call`s: **0 mismatches**.
- `eth_call` at an explicit block number is accepted by java-tron.
- Sub-call ceiling: 100/128/200/300/500 all succeed; ~1000 hits Tron's constant-call CPU limit (`CPU timeout for 'SWAP2' operation executing`), which surfaces as a chunk error and falls back cleanly. The chosen chunk size of 100 therefore has ~10× headroom on the most constrained chain.

**Ethereum mainnet** (`ethereum-rpc.publicnode.com`, chain id 1) — the canonical path used by every other chain:
- The probe finds canonical Multicall3 and `aggregate3` round-trips (`decimals()`=6, `symbol()`="USDC").
- 130 ERC-20 contracts harvested from recent Transfer logs, for one holder, resolved in 2 aggregate3 chunks in ~350 ms; a 40-contract sample cross-checked against independent single `eth_call`s: **0 mismatches**.
- Explicit block number accepted.
- Sub-call ceiling: 100 through 2000 all succeed; 3000 hits the provider's request-body size limit, not a gas limit. No gas cliff was reachable.

**ERC-4626 enrichment** (mainnet) — the other consumer of `EthereumTypeMulticallAggregate3`, re-checked because this PR makes the decoder reject any response whose result count differs from the request:
- Detection over 85 candidates — 81 distinct ERC-20s actually held by three mainnet addresses (sourced from eth.trezor.io: aTokens, variable-debt tokens, rebasing tokens, spam tokens, blue chips) plus 4 known vaults — issued 2 aggregate3 calls totalling 170 sub-calls with no errors. The first chunk carries 128 sub-calls, the largest shape any caller produces.
- Exactly 4 of 85 flagged, with correct underlying assets and **no false positives**: sDAI→DAI, sUSDe→USDe, sfrxETH→frxETH, steakUSDC→USDC.
- Cold path (2 aggregate3) and warm path (1 aggregate3) both returned complete data with no partial-failure errors for all four vaults. `totalAssets` matched an independent single `eth_call` in every case, and the `convertToAssets`/`convertToShares` pairs are reciprocal to 6 decimal places.
- A non-vault (USDC) is reported as a deterministic "not a vault" rather than an error, so the negative cache behaves as intended.

**Deployed instance** (`blockbook-dev3`, Optimism Archive, this branch at `00573b0`, in sync):
- Every balance request takes the aggregate3 path: 30.5k sub-calls coalesced into 332 physical `eth_call`s (~92 each), chunking exact (105 contracts → 2 requests, 838 → 9, 1353 → 14). `rpc_fallback_calls` has no series at all — no probe error, chunk failure or starved tail occurred across the whole session, so 100 `balanceOf` per aggregate3 does not starve on OP.
- Correctness: 214 balances cross-checked against forced single `eth_call`s on the same instance, and 1571 against the batch-path production instance `op-b3` — **0 mismatches** (two apparent diffs on an actively-traded holder re-read as liveness). Contracts dropped as `elem` errors return no balance via single calls either, i.e. genuinely dead `balanceOf`.
- 40 concurrent requests over 8 heavy holders: 21,615 balances in 1.4 s, no errors, byte-identical results per address on every repeat.

**Performance vs. the batch path**, measured against `op-b3` (master, `f1d243b`) — same reth v2.3.0 build at the same height, same network path, server-side via `rpc_latency{method="EthereumTypeGetErc20ContractBalances"}`, 8 repeats warm:

| ERC-20 contracts | multicall | batch | per balance |
|---|---|---|---|
| 27 | 10.9 ms | 20.7 ms | 0.40 vs 0.77 ms |
| 187 | 31.7 ms | 102.8 ms | 0.17 vs 0.55 ms |
| 838 | 138 ms | 470 ms | 0.16 vs 0.56 ms |
| 1353 | 191 ms | 748 ms | 0.14 vs 0.55 ms |

Per-balance cost stays flat at ~0.55 ms on the batch path — each balance is its own EVM execution — while multicall amortizes it down to 0.14 ms. Physical request counts are identical (both chunk at 100), so the win is the backend executing 14 aggregate3 calls instead of 1353 `eth_call`s. On the identical single-call path `op-b3` is 1.2× *faster*, so these ratios understate the gain. Under concurrency 8 the tail widens: p95 406 ms vs 945 ms.

**Gas-starvation handling**: `aggregate3` reports a sub-call that reverted with no returndata identically to one starved of gas. Starvation *usually* shows up as a trailing run — a sub-call that exhausts the gas leaves `aggregate3` only the 1/64 the EVM withholds, rarely enough for the calls after it — so a chunk whose last element did not fail empty was very likely never starved, and its empty failures are treated as bare reverts and stay nil (matching the batch path); only when the tail is starved are the chunk's empty failures re-read through the JSON-RPC batch. This is a heuristic, not an invariant: it assumes comparably priced sub-calls, and a mid-chunk gas hog can spare enough for the rest to succeed, leaving its own empty failure read as a bare revert and its balance nil. Accepted because treating every empty failure as starved costs a metered call per dead token per request — contracts without `balanceOf` (dead or non-conforming tokens, common in holders' lists) are otherwise indistinguishable from starvation.

Measured on live Tron and Ethereum mainnet with a real contract lacking `balanceOf`, a 26-contract request with 25% such tokens bills **1** metered call, down from 6 before this rule (the JSON-RPC batch path bills 26). Two residual cases the signal cannot distinguish: a dead token landing last in a chunk, which is still re-read (one wasted call, no correctness impact); and mid-chunk starvation, whose balance stays nil. On Tron — the tightest budget of the chains here — exceeding the constant-call limit surfaces as a whole-call failure (`CPU timeout for 'SWAP2' operation executing`) that lands on the chunk-error path and falls back cleanly, rather than as per-element empty reverts, so the heuristic is not the mechanism that fires there. `rpc_fallback_calls{component="erc20_multicall", reason="elem_fallback"}` counts starved elements if it ever does.


bchain/coins/eth/contract.go
```diff
@@ -539,43 +549,133 @@ func (b *EthereumRPC) erc20BatchSize() int {
...
-	batcher, ok := b.RPC.(batchCaller)
-	if !ok {
+	// Same calldata for all balanceOf calls; only the contract address varies per element.
+	callData := erc20BalanceOfCallData(addrDesc)
+	// Multicall3 needs no batcher (single eth_call); the batcher is only for the batch path.
+	batcher, hasBatcher := b.RPC.(batchCaller)
+
+	// Prefer Multicall3, but only when the cached probe reports it deployed — otherwise a
+	// non-multicall chain would build and hex-encode calls on every request just to discard them.
+	if deployed, _ := b.probeMulticall3(); deployed {
```
pragmaxim: **1. Probe failure is not cached, and this path is now hot.**

`probeMulticall3` deliberately does not cache a transient `eth_getCode` error ("will retry on next call"). That was fine when only the erc4626 enrichment probed. Now *every* account-info request with more than one token pays an extra `eth_getCode` before falling back to the batch — forever — on any chain whose provider errors on or restricts `eth_getCode`. It is also silent, since the error is discarded here (`deployed, _ :=`).

Net effect on such a chain: +1 metered call per balances request, permanently, with nothing in the metrics to show it.

Suggestion: a consecutive-failure counter that latches `multicall3NotDeployed` after K errors, or at least backoff plus a `ObserveChainDataFallback("erc20_multicall", "probe_error")` so it is visible.

bchain/coins/eth/contract.go
```diff
@@ -539,43 +549,133 @@ func (b *EthereumRPC) erc20BatchSize() int {
...
+				}
+				b.ObserveChainDataFallback("erc20_multicall", "elem_fallback")
+				if subBalances, berr := b.erc20BalancesBatchChunked(batcher, callData, sub, blockNumber); berr == nil {
+					for j, idx := range failed {
+						mcBalances[idx] = subBalances[j]
+					}
+				} else {
+					glog.Warningf("erc20 multicall3 elem fallback failed for %d contract(s): %v", len(sub), berr)
+				}
+			}
```
pragmaxim: **2. Re-resolve cost on revert-heavy addresses may erase the savings.**

Every `Success=false` element is re-resolved on every request, and the code cannot distinguish gas starvation from a deterministic revert. But `ErrInvalidErc20Balance`'s own comment (contract.go:508-512) says reverting/dead/non-conforming tokens are "benign and common for dead/non-conforming/rebasing tokens" — and those are precisely the tokens that linger in holders' contract lists.

So for an address with, say, 30 such tokens, this path now costs 1 aggregate3 **plus** a 30-element batch — more metered calls than the old path, on a request that repeats for every account view, with no caching between requests.

Cheap partial mitigation: skip the re-resolve when `len(results[i].Data) > 2`. A revert carrying reason data or a Solidity `Panic(uint256)` cannot be an out-of-gas sub-call, so those are known-genuine reverts. Bare `revert()` / `require(false)` still returns empty returndata, so this is a filter rather than a proof — but it should remove most of the repeated work while keeping the gas-starvation guard intact.

bchain/coins/eth/contract.go
```diff
@@ -539,43 +549,133 @@ func (b *EthereumRPC) erc20BatchSize() int {
...
+		if err == nil {
+			// aggregate3 shares one gas budget, so a Success=false element can be a gas-starved
+			// (not truly empty) balance. Re-resolve those via independent eth_calls (full gas cap);
+			// a real revert stays nil. Without a batcher we can't, so leave them nil (best effort).
+			if len(failed) > 0 && hasBatcher {
+				sub := make([]bchain.AddressDescriptor, len(failed))
+				for j, idx := range failed {
+					sub[j] = contractDescs[idx]
+				}
+				b.ObserveChainDataFallback("erc20_multicall", "elem_fallback")
```
pragmaxim: **4. `elem_fallback` counts requests, not elements.**

This increments once per request regardless of `len(failed)`, so the metric cannot distinguish "one dead token re-resolved" from "80 of 100 re-resolved". That element count is precisely the number that tells you whether the credit savings this PR is measuring are actually being realized, or being eaten by the re-resolve path (see the comment on the surrounding block).

`ChainDataFallbacks.With(...).Add(float64(len(failed)))` — or a separate histogram of the failed-subset size — would make it actionable.

bchain/coins/eth/contract.go
```diff
@@ -539,43 +549,133 @@ func (b *EthereumRPC) erc20BatchSize() int {
...
+			return nil, nil, errors.Errorf("multicall3 erc20 balances: got %d results, want %d", len(results), len(calls))
+		}
+		for i := range results {
+			if !results[i].Success {
+				failed = append(failed, start+i)
+				continue
+			}
+			// nil on unparseable output (empty/short), matching the batch path
+			balances[start+i] = parseSimpleNumericProperty(results[i].Data)
+		}
```
pragmaxim: **3. Metric continuity gap in the multicall path.**

The batch path emits `observeEthCallError("batch", "invalid")` when `parseSimpleNumericProperty` returns nil (contract.go:752) and `observeEthCallError("batch", "elem")` on per-element failure (contract.go:730). The multicall path emits neither: a nil parse is silent, and `Success=false` is silent.

Consequence: `eth_call_errors{type="invalid"}` volume drops off the dashboard the moment a chain flips onto multicall, and there is no signal at all for per-element aggregate3 failures. For a PR whose explicit goal is that "the credit savings are measurable" and that per-token read volume "stays continuous", this is the one place where continuity breaks.

Suggestion: `observeEthCallError("multicall", "elem")` on `!results[i].Success` and `observeEthCallError("multicall", "invalid")` when the parse yields nil, mirroring the batch path.

bchain/coins/eth/multicall.go
```diff
@@ -96,14 +111,16 @@ func (b *EthereumRPC) probeMulticall3() (bool, error) {
 
 		ctx, cancel := context.WithTimeout(context.Background(), b.Timeout)
 		defer cancel()
+		// probe the same address aggregate3 will call (override-aware)
+		addr := b.multicall3ContractAddress()
 		var code string
-		if err := b.RPC.CallContext(ctx, &code, "eth_getCode", multicall3Address, "latest"); err != nil {
-			glog.Warningf("multicall3 probe at %s failed: %v (will retry on next call)", multicall3Address, err)
+		if err := b.RPC.CallContext(ctx, &code, "eth_getCode", addr, "latest"); err != nil {
```
pragmaxim, line 154: **5. Probe is at `latest`, but aggregate3 runs at `blockNumber`.**

For a `blockNumber` below the chain's Multicall3 deployment, `eth_call` to a then-codeless address returns `"0x"`, which `decodeAggregate3Result` rejects as "response too short". That surfaces as an error, so it is *safe* — the caller logs a warning, bumps `rpc_fallback_calls{reason="error"}`, and falls back to the batch — but it happens on every such request, and it pollutes the "unexpected fallback" metric that is supposed to be quiet.

Impact today is essentially nil: `bchain/erc20_batch_integration.go:103` is the only caller passing a non-nil block, and it uses the best block height. But it is a trap for the next caller of `...AtBlock`. Worth either a comment here, or skipping the multicall path when `blockNumber` is non-nil and older than the probe.

pragmaxim (commented):
Reviewed `138b9c32`. Verified locally: `go build ./...`, `go vet`, and `go test ./bchain/coins/eth/... ./bchain/coins/tron/...` (150 tests) all pass. Also confirmed the Tron constant is correct — base58 `TEazPvZwDjDtFeJupyo7QunvnrnUjPH8ED` decodes with a valid checksum to `0x41` + `32a4f47a74a6810bd0bf861cabab99656a75de9e`.

The design is sound and the risky part is handled well: because `api/worker.go:1298-1305` treats map presence as "already covered, do not single-call", a gas-starved `Success=false` would otherwise surface as a *wrong* balance rather than a missing one — the per-element re-resolve is exactly the right guard. Test coverage is targeted at the actual claims (call-count assertions, whole-chunk fallback, length mismatch, chunk bounding, genuine-revert-stays-nil).

Five things worth addressing, left as inline comments. I'd consider (1) and (3) blocking: (1) is a silent per-request cost regression on a hot path, and (3) undercuts the PR's own observability story. (2) is the one to watch in production — most relevant for Tron, where the constant-call energy cap for a 100-element aggregate3 is still unverified against a live node.

Nits not worth inline comments: `min(erc20BatchSize, multicall3MaxCallsPerAggregate)` couples two limits that bound different things (request count vs gas), so lowering `erc20_batch_size` for a provider's batch limit needlessly shrinks aggregate3 chunks too (latent — both default to 100); `probeMulticall3`'s doc comment still says "deployed at the canonical address" though it is now override-aware; `mode == ""` as a "don't instrument" sentinel in `ethCallAtBlock` would read better as a bool; `Multicall3AddressOverride` is unvalidated, so a typo degrades to "not deployed" instead of failing loudly; and the balances path probes, then `EthereumTypeMulticallAggregate3` probes again (harmless atomic load).

Test gap: no coverage for `Success=false` *without* a batcher — the one branch where a user-visible wrong balance is possible by design.

bchain/coins/eth/contract.go
cranycrane: Fixed in `05436d95`. The probe now latches to not-deployed after `multicall3MaxProbeFailures` (5) consecutive transient errors, so an `eth_getCode`-restricting provider stops paying a probe per request; and the transient case now emits `ObserveChainDataFallback("erc20_multicall", "probe_error")` so it is visible.

bchain/coins/eth/contract.go
cranycrane: Fixed in `05436d95` via your returndata heuristic: only `Success=false` elements with **empty** returndata are re-resolved (possible gas starvation). A failure carrying revert returndata (Error/Panic) is a genuine revert and stays nil, so revert-heavy addresses no longer pay a re-resolve batch every request. Covered by `TestEthereumTypeGetErc20ContractBalancesGenuineRevertNotReresolved`.

bchain/coins/eth/contract.go
cranycrane: Fixed in `05436d95`. Added `observeChainDataFallback(component, reason, count)`; the call site now records `Add(len(failed))`, so the metric counts elements re-resolved rather than requests. (After the comment-2 change `failed` is only the empty-returndata subset, so this now tracks the potential-gas-starvation volume specifically.)

bchain/coins/eth/contract.go
cranycrane: Fixed in `05436d95`. The multicall path now emits `observeEthCallError("multicall", "elem")` on `!Success` and `observeEthCallError("multicall", "invalid")` on a nil parse, mirroring the batch path so `eth_call_errors` continuity holds when a chain moves onto aggregate3.

bchain/coins/eth/multicall.go
cranycrane: Documented in `05436d95` rather than adding deployment-block tracking: the only non-nil-block caller uses the best block height, and a historical block below deployment already falls back safely (aggregate3 returns `0x` → decode error → batch). Added a comment at the probe explaining the trap for future `...AtBlock` callers, and fixed the stale "canonical address" wording in the probe doc.

bchain/coins/eth/multicall.go
```diff
@@ -96,14 +116,26 @@ func (b *EthereumRPC) probeMulticall3() (bool, error) {
...
+			// Latch to not-deployed after repeated transient failures so a provider that
+			// restricts eth_getCode does not incur an extra probe on every request forever.
+			if b.multicall3ProbeFailures.Add(1) >= multicall3MaxProbeFailures {
+				glog.Warningf("multicall3 probe at %s failed %d times; disabling multicall, using JSON-RPC batch: %v", addr, multicall3MaxProbeFailures, err)
+				b.multicall3Probe.Store(multicall3NotDeployed)
+				return probeResult{}, nil
+			}
```
pragmaxim: **The latch is permanent, and its blast radius is wider than balances.**

`multicall3ProbeFailures` never resets and the K-th failure stores `multicall3NotDeployed` for the process lifetime. Three things make that a harsher trade than the problem it fixes:

1. **Five consecutive probe rounds is a low bar.** A backend restart, or a brief provider 5xx window while the API keeps serving address requests, reaches it easily. Blockbook processes then run for weeks with multicall off — the entire saving this PR exists for, gone until someone restarts the process.
2. **It disables more than balances.** `probeMulticall3` is shared, so the latch also permanently kills ERC-4626 enrichment: `api/erc4626.go:184-187` silently `continue`s on the resulting `errMulticall3NotDeployed`.
3. **The decisive event is invisible in metrics.** See the separate comment on `contract.go:573`.

Suggestion: re-arm instead of latching. An `atomic.Int64` holding a `nextProbeAt` deadline keeps the "no probe on every request" property while recovering after the outage — on the K-th failure set the deadline ~10 minutes out and return `(false, nil)` *without* storing `multicall3NotDeployed`, then allow one re-probe once it passes. Alternatively, reset `multicall3ProbeFailures` whenever any other RPC call succeeds.

`TestProbeMulticall3_LatchesNotDeployedAfterRepeatedFailures` would need adjusting; nothing else in the suite depends on this.

bchain/coins/eth/contract.go
```diff
@@ -539,43 +549,148 @@ func (b *EthereumRPC) erc20BatchSize() int {
...
+	callData := erc20BalanceOfCallData(addrDesc)
+	// Multicall3 needs no batcher (single eth_call); the batcher is only for the batch path.
+	batcher, hasBatcher := b.RPC.(batchCaller)
+
+	// Prefer Multicall3, but only when the cached probe reports it deployed — otherwise a
+	// non-multicall chain would build and hex-encode calls on every request just to discard them.
+	deployed, probeErr := b.probeMulticall3()
+	if probeErr != nil {
+		// Transient probe failure — visible so an eth_getCode-restricting provider is not silent.
+		b.ObserveChainDataFallback("erc20_multicall", "probe_error")
```
pragmaxim, line 571: Follow-on to the latch comment on `multicall.go:126`: `probe_error` fires only while `probeErr != nil`, but the call that actually trips the latch returns `(false, nil)`, so the transition to permanently-disabled emits no metric at all — just the one `glog.Warningf`. That is the one event you would want to alert on, and it is the only one not counted, which sits a little awkwardly next to the `mode="multicall"` error metrics added in this same commit.

A distinct `reason="probe_disabled"` (or `"probe_latched"`) on the latching branch would close it.

pragmaxim (commented):
Re-reviewed `05436d95`. Verified: `go build ./...`, `go vet ./bchain/... ./api/...`, and `go test ./bchain/coins/eth/... ./bchain/coins/tron/... ./api/...` — 191 tests pass.

**All five are resolved, and correctly:**

- **(2) revert filter** — `len(results[i].Data) <= 2` is exactly the right predicate. `decodeAggregate3Result` always emits `"0x"+hex`, so length 2 is precisely empty returndata; out-of-gas and invalid-opcode produce empty returndata, while an explicit `REVERT` carrying `Error`/`Panic` data cannot. The `Panic(0x11)` test asserting a size-1 re-resolve batch pins the distinction well.
- **(3) metrics** — `observeEthCallError("multicall", "elem"/"invalid")` now mirrors the batch path exactly, including counting benign reverts under `"elem"` the way `erc20BalancesBatchAtBlock` does.
- **(4) element count** — clean; keeping `ObserveChainDataFallback` as a `count=1` wrapper avoids touching other callers.
- **(5) historical blocks** — the comment is accurate and sufficient given only the integration helper passes a non-nil block today.
- Plus the `probeMulticall3` doc fix and the no-batcher `Success=false` test I mentioned as gaps.

One new issue introduced by the fix for (1), inline below. It is not a correctness problem — the code always degrades safely to the batch path — but the failure mode is a silent, permanent loss of the savings this PR exists for, which I think is a worse trade than the extra probe it removes.

Nit: `rpc_fallback_calls_total{component="erc20_multicall", reason="elem_fallback"}` now counts *elements* while every other series in that family counts *requests*, so `sum by (component)` mixes units. A reason name that signals it, or a separate metric, would avoid a confusing dashboard later.

Otherwise this looks good to merge.

pragmaxim (approved):
LGTM

bchain/coins/eth/multicall.go
```diff
@@ -19,18 +23,48 @@ const multicall3Address = "0xcA11bde05977b3631167028862bE2a173976CA11"
...
+// provider, yet multicall is never disabled for good.
+const multicall3ProbeSuspendInterval = 10 * time.Minute
+
+// multicall3MaxCallsPerAggregate bounds one aggregate3 eth_call by its shared gas budget;
+// deliberately independent of erc20BatchSize, which exists for JSON-RPC request-count limits.
+const multicall3MaxCallsPerAggregate = 100
+
+// errMulticall3NotDeployed is returned on chains where Multicall3 is not deployed
+// at the probed address (canonical, or a per-chain override); cached for the process lifetime.
+var errMulticall3NotDeployed = errors.New("multicall3 not deployed on this chain")
```
pragmaxim, line 48: **[medium]** During the probe-suspension window, `probeMulticall3` returns `(false, nil)`, so `EthereumTypeMulticallAggregate3` maps a *transient* condition to this sentinel — whose doc says the verdict is "cached for the process lifetime." No functional impact today: the only external consumer (the erc4626 enrichment) treats every error the same (skip chunk, retry next request). But the sentinel invites a future caller to match on it and permanently disable multicall, latching a 10-minute RPC outage into process-lifetime disablement. Consider a distinct sentinel (or a wrapped transient error) for the suspended case, or at minimum adjusting this doc comment to note it may also be returned transiently while probing is suspended.

bchain/coins/eth/contract.go
```diff
@@ -539,43 +545,155 @@ func (b *EthereumRPC) erc20BatchSize() int {
...
+				// a reverting balanceOf yields Success=false, not a failed batch
+				AllowFailure: true,
+			}
+		}
+		results, err := b.EthereumTypeMulticallAggregate3(calls, blockNumber)
+		if err != nil {
+			// Chunk failures are usually systemic: stop at the first, keep decoded chunks and
+			// leave the rest to the caller — one fallback event and one log line per request.
+			holes = append(holes, valid[start:]...)
+			b.ObserveChainDataFallback("erc20_multicall", "error")
```
pragmaxim, line 658: **[medium]** The probe has a suspension mechanism, but the chunk path has no breaker: on a provider whose `eth_call` gas or response-size caps make a 100-sub-call aggregate3 fail systematically (the probe still succeeds — code exists at the address), every balances request pays one wasted metered call per chunk plus a warning log, indefinitely. It degrades gracefully to the batch path and is visible via `rpc_fallback_calls{component="erc20_multicall", reason="error"}`, so this is operational rather than a bug — but a "suspend aggregate3 after K consecutive chunk failures" latch mirroring the probe suspension would make such a provider cost-neutral instead of cost-plus. Fine as a follow-up.

pragmaxim (commented):
Reviewed the full diff against the branch — ABI layout math, decoder hardening (count pinned to request, aliased-head bound, negative-wrap rejection), fallback tiers, hole index mapping, probe state machine, and metric semantics all check out; no correctness bugs found. `go build` / `go vet` / package tests pass locally. Two medium observations posted inline; the deploy-time notes (dashboard series shift, accepted mid-chunk starvation trade-off) are already covered in the PR description.

PR #1703, opened 2026-08-05
fix(db): keep contract registry consistent for same-block lifecycles and reorgs

cranycrane:
Two latent bugs in the shared Ethereum-type contract registry (`cfContracts`), plus one concurrency fix. All three are live on every eth-type chain today: `processCallTrace` ([`bchain/coins/eth/ethrpc.go#L1465-L1485`](https://github.com/trezor/blockbook/blob/master/bchain/coins/eth/ethrpc.go#L1465-L1485)) has always emitted contract creation and destruction records into `blockSpecificData.Contracts`. They surfaced while adding Tron as a second producer (#1682).

**1. A contract created and destroyed in the same block was recorded as alive.**

`storeContractInfo` resolves a destruction record (`CreatedInBlock == 0 && DestructedInBlock != 0`) against `GetContractInfo`, which cannot see the uncommitted write batch — and the creation's LRU entry was just invalidated. So for the ephemeral `create → use → selfdestruct` pattern (common, MEV-style) the lookup returned `nil` and the destruction was **silently dropped**, leaving a row that claims the contract is still alive and no `destructedInBlock` in the API.

Registry writes of one block now share an in-batch map that destructions merge into.

**2. Nothing removed `cfContracts` rows on disconnect.**

`DisconnectBlockRangeEthereumType` reverted address/contract indexes and protocol rows, but never the registry itself. A creation in a reorged-away block therefore left a stale row behind — and any row at all makes `GetBalanceHistory` refuse the address outright:

```go
// api/worker.go:2083-2092 — do not get balance history for contracts
if ci != nil {
    return nil, NewAPIError("GetBalanceHistory for a contract not allowed", true)
}
```

so the address loses balance history permanently, until a reindex. The disconnect now deletes rows whose creation fell in the disconnected range and resets disconnected destructions back to alive, composing across the whole range (a creation in block N and its destruction in N+3 both revert correctly) and keying off the internal-data rows that already drive the rest of the rollback.

**3. The resolved record was mutated in place on the shared LRU value.**

`GetContractInfo` returns the pointer held by `cachedContracts`; `storeContractInfo` then assigned `DestructedInBlock` on it, publishing an uncommitted destruction height to any concurrent reader that still held the entry. The record is now copied before mutation.

**Scope:** `storeContractInfo` is private and gains an `inBatch` parameter; the public `StoreContractInfo` signature is unchanged and passes `nil`, preserving the previous single-row behaviour for the admin path in `server/internal.go`. No config, index format, or migration change — packed row layout is untouched.

**Not healed by this PR:** rows already stale in a deployed index stay stale; a reindex restores them (`CreatedInBlock`/`DestructedInBlock` are sync-owned and cannot be re-fetched from a node). This stops new ones from appearing.

**Testing:** `TestStoreContractInfo_MergesSameBatchCreateAndDestroy` and `TestDisconnectBlockRange_RollsBackContractRegistry` cover both bugs and fail on master. `go test -tags unittest ./db/... ./api/...` passes.

**Relationship to #1682:** prerequisite. That PR enables internal transaction processing for Tron, which starts producing registry events on a chain that previously wrote none — Tron must not be reindexed before this lands, or the reindex bakes the wrong rows in.

### DIFF
--- a/db/rocksdb_ethereumtype.go
+++ b/db/rocksdb_ethereumtype.go
@@ -1312,14 +1312,28 @@ func (d *RocksDB) disconnectAddress(btxID []byte, internal bool, addrDesc bchain
 	return nil
 }
 
-// rollbackContractRegistryRow reverts the cfContracts row of a contract whose
-// creation or destruction happened in the disconnected block. rollback carries
-// the pending row states of the whole disconnected range (nil = delete row), so
-// a destruction reset in a later block composes with the creation removal in an
-// earlier one.
-func (d *RocksDB) rollbackContractRegistryRow(addrDesc bchain.AddressDescriptor, height uint32, destroyed bool, rollback map[string]*bchain.ContractInfo) error {
+// contractRegistryRollback holds the pending cfContracts rows of a whole
+// disconnected range, so a later block's destruction reset composes with an
+// earlier block's creation removal.
+type contractRegistryRollback struct {
+	// nil = delete the row
+	rows map[string]*bchain.ContractInfo
+	// height whose destruction frame was already rolled back, per key
+	destroyedAt map[string]uint32
+}
+
+func newContractRegistryRollback() *contractRegistryRollback {
+	return &contractRegistryRollback{
+		rows:        make(map[string]*bchain.ContractInfo),
+		destroyedAt: make(map[string]uint32),
+	}
+}
+
+// rollbackContractRegistryRow reverts the cfContracts row of a contract created
+// or destroyed in the disconnected block.
+func (d *RocksDB) rollbackContractRegistryRow(addrDesc bchain.AddressDescriptor, height uint32, destroyed bool, rollback *contractRegistryRollback) error {
 	key := string(addrDesc)
-	ci, pending := rollback[key]
+	ci, pending := rollback.rows[key]
 	if !pending {
 		stored, err := d.GetContractInfo(addrDesc, "")
 		if err != nil {
@@ -1337,25 +1351,30 @@ func (d *RocksDB) rollbackContractRegistryRow(addrDesc bchain.AddressDescriptor,
 		return nil
 	}
 	if destroyed {
+		rollback.destroyedAt[key] = height
 		if ci.DestructedInBlock == height {
 			ci.DestructedInBlock = 0
-			rollback[key] = ci
+			rollback.rows[key] = ci
 		}
 	} else if ci.CreatedInBlock == height {
-		rollback[key] = nil
+		// a same-block destroy-then-recreate overwrote the pre-range CreatedInBlock,
+		// so only an unseen destruction proves the contract originated here - keeping
+		// a stale row beats dropping a contract that is live on the canonical chain
+		if destroyedAt, ok := rollback.destroyedAt[key]; !ok || destroyedAt != height {
+			rollback.rows[key] = nil
+		}
 	}
 	return nil
 }
 
-func (d *RocksDB) disconnectInternalData(btxID []byte, height uint32, addresses map[string]map[string]struct{}, contracts map[string]*unpackedAddrContracts, registryRollback map[string]*bchain.ContractInfo) error {
+func (d *RocksDB) disconnectInternalData(btxID []byte, height uint32, addresses map[string]map[string]struct{}, contracts map[string]*unpackedAddrContracts, registryRollback *contractRegistryRollback) error {
 	internalData, err := d.getEthereumInternalData(btxID)
 	if err != nil {
 		return err
 	}
 	if internalData != nil {
-		// no top-level destruction branch: the persisted tx type is one bit
-		// (CALL|CREATE, see packEthInternalData), so a root SELFDESTRUCT cannot
-		// reach here - destructions always arrive as the transfer frames below
+		// destructions arrive only as transfer frames below: the persisted tx type
+		// is one bit (CALL|CREATE), so a root SELFDESTRUCT cannot round-trip
 		if internalData.Type == bchain.CREATE {
 			contract, err := d.chainParser.GetAddrDescFromAddress(internalData.Contract)
 			if err != nil {
@@ -1402,7 +1421,7 @@ func (d *RocksDB) disconnectInternalData(btxID []byte, height uint32, addresses 
 	return nil
 }
 
-func (d *RocksDB) disconnectBlockTxsEthereumType(wb *grocksdb.WriteBatch, height uint32, blockTxs []ethBlockTx, contracts map[string]*unpackedAddrContracts, registryRollback map[string]*bchain.ContractInfo) error {
+func (d *RocksDB) disconnectBlockTxsEthereumType(wb *grocksdb.WriteBatch, height uint32, blockTxs []ethBlockTx, contracts map[string]*unpackedAddrContracts, registryRollback *contractRegistryRollback) error {
 	glog.Info("Disconnecting block ", height, " containing ", len(blockTxs), " transactions")
 	addresses := make(map[string]map[string]struct{})
 	for i := range blockTxs {
@@ -1467,7 +1486,7 @@ func (d *RocksDB) DisconnectBlockRangeEthereumType(lower uint32, higher uint32) 
 	wb := grocksdb.NewWriteBatch()
 	defer wb.Destroy()
 	contracts := make(map[string]*unpackedAddrContracts)
-	registryRollback := make(map[string]*bchain.ContractInfo)
+	registryRollback := newContractRegistryRollback()
 	for height := higher; height >= lower; height-- {
 		if err := d.disconnectBlockTxsEthereumType(wb, height, blocks[height-lower], contracts, registryRollback); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


db/rocksdb_ethereumtype.go
```diff
@@ -1308,12 +1312,50 @@ func (d *RocksDB) disconnectAddress(btxID []byte, internal bool, addrDesc bchain
...
+	if ci == nil {
+		// already deleted by a creation rollback
+		return nil
+	}
+	if destroyed {
+		if ci.DestructedInBlock == height {
+			ci.DestructedInBlock = 0
+			rollback[key] = ci
+		}
+	} else if ci.CreatedInBlock == height {
```
pragmaxim, line 1359: **Same-block destroy-then-recreate deletes the row of a contract that predates the disconnected range** (low severity, and not a regression — but worth a line in the doc comment).

Contract `X` was created in block `N0`. In block `H`, tx1 `SELFDESTRUCT`s it and a later tx of the *same* block re-deploys it at the same address (pre-EIP-6780 metamorphic `CREATE2`).

- Connect: the records arrive in tx order, so the destruction merges into the `N0` row (`{N0, H}`) and then the sparse creation record overwrites it at `db/rocksdb_contracts.go:194` → `{CreatedInBlock: H, DestructedInBlock: 0}`. The connect path simply cannot represent this lifecycle, and that is what leaves `CreatedInBlock == height` for the disconnect to match on.
- Disconnect: `disconnectBlockTxsEthereumType` walks txs in forward order, so tx1's `SELFDESTRUCT` frame is evaluated first — `ci.DestructedInBlock` is `0`, not `height`, so the key is never marked pending. The later `CREATE` frame then hits this line and sets `rollback[key] = nil`, so the flush at `:1484-1485` **deletes** the row of a contract that exists on the canonical chain from `N0`.

Reproduced on a throwaway copy of this checkout: after `ConnectBlock(block2)` the row reads `{CreatedInBlock: 4321001, DestructedInBlock: 0}` for a contract created in 4321000, and after `DisconnectBlockRangeEthereumType(4321001, 4321001)` `GetContractInfoForAddress` returns `nil`. The ordinary ephemeral case is unaffected (`CREATE` frame first → delete, `SELFDESTRUCT` then short-circuits on `ci == nil`), matching the new test.

Why it is worth a mention rather than a shrug: the deleted row is the only home of the sync-owned `CreatedInBlock`, which per this repo's own note at `db/rocksdb_contracts.go:257-264` a backend re-fetch cannot restore — only a reindex can. And `GetBalanceHistory` gates purely on row existence (`api/worker.go:2084-2092`), so the delete flips a genuine long-lived contract from "correctly refused" to "serves a balance history for a contract" — the mirror image of the bug this PR fixes. Bounded, though: on EIP-6780 chains a pre-range contract cannot be re-deployed at the same address at all, so this needs pre-Dencun history (a deep `-rollback`) or a non-Cancun eth-type chain, and the pre-PR behaviour in this same scenario was also wrong (it left `CreatedInBlock = H` for a contract created at `N0`).

Suggestion: document the limitation here — the helper can reset a destruction to `0` but can never restore a pre-range `CreatedInBlock`, so a same-block destroy→recreate pair rolls back to a deleted row. If you want the stricter behaviour, the frame order is a sufficient discriminator: a destruction rollback seen for this key *before* a creation rollback at the same height means the row predates the range, so skip the delete (leaving a stale `CreatedInBlock` but keeping the "is a contract" flag, which is the less harmful of the two wrong states).

db/rocksdb_ethereumtype_test.go
```diff
@@ -996,6 +996,97 @@ func TestRocksDB_Index_EthereumType(t *testing.T) {
...
+		}
+	}
+
+	// connected state; the ephemeral row proves the same-batch merge
+	assertRegistryRow(topLevel, &bchain.ContractInfo{CreatedInBlock: block2.Height})
+	assertRegistryRow(nested, &bchain.ContractInfo{CreatedInBlock: block2.Height})
+	assertRegistryRow(ephemeral, &bchain.ContractInfo{CreatedInBlock: block2.Height, DestructedInBlock: block2.Height})
+	assertRegistryRow(longLived, &bchain.ContractInfo{CreatedInBlock: block1.Height, DestructedInBlock: block2.Height})
+
+	if err := d.DisconnectBlockRangeEthereumType(block2.Height, block2.Height); err != nil {
```
pragmaxim, line 1079: **The cross-block composition the range-scoped map exists for is untested.** The doc comment at `db/rocksdb_ethereumtype.go:1315-1319` sells exactly this ("a creation in block N and its destruction in N+3 both revert correctly"), but this call disconnects a single height, so `registryRollback` never carries state between heights.

A `-covermode=count` run over the whole `db` package puts numbers on it: `rollbackContractRegistryRow` is entered 7 times in the entire suite, 6 of those take the `if !pending` DB-read path, and the single call that did see a pending entry returned at `if ci == nil`. So the *pending-and-non-nil* arm — both the `destroyed` reset at `:1339-1343` seen again in a lower block, and the docstring's "creation rollback replaces it with nil" at `:1344-1346` — has zero coverage anywhere. It cannot be reached inside one block, because a `CREATE` frame always precedes its `SELFDESTRUCT` in disconnect order; a multi-block range is the only way in. The two pre-existing multi-block calls don't help: `:964` asserts the early "necessary to rebuild index" failure, and `db/rocksdb_protocols_test.go:186-197` seeds empty `blockTxs` rows.

The branch is *correct* as written (I ran the missing scenario against this HEAD via `go test -overlay`: create `X` in `4321001`, destroy in `4321002`, disconnect `[4321001, 4321002]` → `X`'s row deleted, the block-4321000 contract untouched with `DestructedInBlock: 0`), so this is regression risk rather than a defect.

One trap if you add the test: it cannot be written against `ethereumTestnetParser()`, which is `NewEthereumParser(1, true)` → `BlockAddressesToKeep: 1`, so `cleanupBlockTxs` (`db/rocksdb.go:1218-1232`) prunes the older `cfBlockTxs` row and the multi-block disconnect aborts with "Cannot disconnect blocks with height …" before any rollback runs. Use `eth.NewEthereumParser(2, true)` for that test, or seed `cfHeight`/`cfBlockTxs` directly the way `TestDisconnectBlockRangeEthereumType_BumpsReorgGenAndRevertsProtocols` does.

db/rocksdb_ethereumtype.go
```diff
@@ -1106,8 +1106,12 @@ func (d *RocksDB) storeBlockSpecificDataEthereumType(wb *grocksdb.WriteBatch, bl
 				return err
 			}
 		}
+		// registry events of one block may create and destroy the same
+		// (ephemeral) contract - the in-batch map lets the destruction merge
+		// into the creation written a moment ago
+		inBatch := make(map[string]*bchain.ContractInfo)
```
pragmaxim, line 1112: Adjacent, not blocking — just so the PR's scope is unambiguous: the other row family this same function writes, `cfAddressAliases` (`:1104-1107` → `storeAddressAliasRecords` → `wb.PutCF` at `db/rocksdb.go:1918`), is still never reverted on disconnect. `cfAddressAliases` has exactly two non-test uses in the tree, that `PutCF` and the `GetCF` at `db/rocksdb.go:1898` — there is no `DeleteCF` anywhere. So an ENS/alias record registered in a reorged-away block keeps being served as that address's label until the same address is re-observed under a different name, which is the same "connect writes it, disconnect forgets it" shape this PR fixes for `cfContracts`. Fine to leave for a follow-up; worth naming so "registry consistent on disconnect" isn't read as covering all of `blockSpecificData`.

pragmaxim (commented):
Reviewed the diff together with the surrounding connect/disconnect paths and both registry producers (`eth.processCallTrace` and the Tron provider). The fix looks correct and well scoped: the rollback keys off the persisted `cfInternalData` frames whose addresses match exactly what the registry records use (`CREATE` → `to`, `SELFDESTRUCT` → `from`, root `CREATE` → `internalData.Contract`), the higher→lower walk with a range-scoped map gets the composition order right, and reusing the existing `reorgGen` bump instead of inventing a second invalidation mechanism is the right call — the new test's post-disconnect reads go through the LRU, so that path is covered implicitly. `go test -tags unittest ./db/... ./api/...` passes here (299 tests), `gofmt`/`go vet` clean.

Three notes below: one low-severity edge case in `rollbackContractRegistryRow` that I reproduced, one coverage gap in the exact branch the range-scoped map exists for, and one adjacent row family that is still never rolled back.

Things I checked that are **clear**, in case it saves someone else the trip:

- The new `// no top-level destruction branch` comment. The invariant is producer-dependent, but it holds. For eth a root `SELFDESTRUCT` registers no `ContractInfo` at all (`bchain/coins/eth/ethrpc.go:1558-1560`), so there is nothing to revert. Today's Tron provider *does* emit a root destruction record (`tronInternalDataProvider.go:173-177`) whose transfer frame can be dropped by the `cv.CallValue <= 0 || cv.TokenID != ""` guard at `:192` — but that path is gated off (`processInternalTransactions` is absent from `configs/coins/tron.json` and `tron_testnet_nile.json`), and #1682 deletes `detectTopType` and the root destruction record outright, moving destructions to per-frame records plus an unconditional zero-value `CREATE`/`SELFDESTRUCT` frame ("eth processCallTrace parity"). So once Tron enables internal-tx processing, every registry destruction arrives as a rollback-able frame — exactly as the comment asserts.
- `bulkconnect` cannot land a creation and a *next* block's destruction in one uncommitted batch: both paths `WriteBatch` before the next block is processed (`db/bulkconnect.go:455`→`463`, `:481`→`:484`), so the cross-block case still resolves from committed state and does not need `inBatch`.
- The LRU entries that the new pre-commit reads in `rollbackContractRegistryRow` populate are stamped with the pre-bump `reorgGen`, so the bump at `db/rocksdb_ethereumtype.go:1497` invalidates them; a failed `WriteBatch` skips the bump but also wrote nothing.

pragmaxim (approved):
LGTM

PR #1705, opened 2026-08-06
feat(eth)!: heal the internal data error queue automatically

cranycrane:
Replaces the manual **Retry fetch** button on `/admin/internal-data-errors` with a routine that drains the queue on its own. Today a transient `debug_traceBlockByHash` failure leaves a permanent hole in the index — missing internal transfers, and a contract whose `createdInBlock` stays 0 — until an operator happens to open the admin page and press the button.

A simpler alternative to #1693, same goal.

### How it works

- `healInternalDataLoop` in `blockbook.go` starts after the initial sync, first pass 1 minute later, hourly after that. Gated on ethereum-type + `processInternalTransactions`, joined on shutdown before RocksDB is closed.
- The retry cap is gone. The existing `retryCount` byte now schedules the next attempt — a block that has failed `n` times is attempted on passes divisible by `2^min(n,6)`, a wait of up to ~2.7 days — and pass 0 matches every schedule, so a restart retries the whole queue once. Nothing is abandoned permanently, no on-disk format change.
- A failure is only charged when it is about the block — the trace failed again, or the backend definitively does not have the block. A transport error during an outage is logged only, so it cannot back off the whole queue.
- Heal attempts are paced one second apart — after a long outage the queue can hold tens of thousands of blocks, and pass 0 retries all of them after a restart, so back-to-back traces would compete with live sync for the backend. The sleep yields to shutdown.
- Every pass refreshes the `blockbook_internal_data_error_queue` gauge and counts outcomes in `blockbook_internal_data_heals_total` (`healed` / `failed` / `fetch_error` / `reconnect_error`) — every attempted block lands in exactly one, so a pass lost to a backend outage (`fetch_error`, uncharged) is distinguishable from a pass that attempted nothing, and a queue that never drains raises an alert instead of waiting for someone to open the admin page.
- The admin page becomes read-only; the Retries column now shows which blocks have backed off.

### Guards in the reconnect

Healing writes a block's internal data long after the surrounding blocks are indexed, which the reconnect path was never written for:

- reorg check — the height must still hold the block's hash, otherwise the write would land on an orphan and drop another block's queue entry;
- per-tx skip when internal data are already stored — `InternalTxs`/`NonContractTxs` are cumulative and a second store would inflate them permanently;
- contract lifecycle records are overlaid on the stored row instead of overwriting it, so a contract a client already enriched keeps its name/symbol, and the create/destruct heights are guarded against rewinding a reused address (SELFDESTRUCT + CREATE2);
- address aliases are deliberately not re-stored — they come from the logs, which succeeded during sync.

### Testing

Unit tests cover the backoff schedule, the pacing sleep giving way to shutdown, the lifecycle guards, the queue update (no resurrecting a removed entry, no overwriting a different block at the same height), and a healed block compared byte-for-byte against the same block synced normally, across every column family. A full healing pass is driven through a stubbed chain against a real database, pinning the failure classification — transport errors retried without charge, `Block not found` and a repeated trace error charged — together with the metric arithmetic and the backoff skip.

Not yet run against a live chain — the end-to-end pass, a healed `/api/v2/tx`, and SIGTERM during a pass still need a canary.

### DIFF
--- a/api/ethereumtype.go
+++ b/api/ethereumtype.go
@@ -3,6 +3,8 @@ package api
 import (
 	stdErrors "errors"
 	"math"
+	"os"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/juju/errors"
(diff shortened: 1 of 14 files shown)


pragmaxim (approved):
LGTM

PR #1706, opened 2026-08-06
chore(eth):  change alternative mempool tx timeout to 3 hours to alig…

cranycrane:
## What

Sets `alternativeMempoolTxTimeout: "3h"` and `mempoolTxTimeout: "48h"` in `ethereum.json` and `ethereum_archive.json`. Config-only change; the options themselves already exist on master.

## Why

Blink's MEV pool keeps a private transaction mineable for up to 3 hours. With an alternative send-tx provider enabled, Blockbook's defaults (5 min alternative cache, 10 min mempool retention) evict the transaction long before Blink stops trying to mine it — it disappears from the pending view and stops raising the sender's pending-nonce floor while still in flight.

`mempoolTxTimeout` must be set explicitly alongside the cache value: with an alternative provider enabled and no explicit value, mempool retention falls back to the 10-minute default — below the 3-hour cache — and the mempool would evict a Blink transaction while the cache still holds it and keeps raising the pending-nonce floor for a tx no longer shown as pending. `"48h"` satisfies mempool ≥ cache while preserving the current public-tx retention.

## Effects

- **Blink-enabled deployments** (`ETH_ALTERNATIVE_SENDTX_URLS` set): private transactions stay visible as pending, resolvable via `getTransaction`, and counted in the pending nonce for up to 3 hours, matching Blink's own retention. Their lifecycle is bounded by the alternative cache — cache eviction (mined, nonce superseded, timeout) also removes the tx from the mempool — so the 48h mempool ceiling never applies to them. Deterministic early evictions still apply.
- **Public (non-private) transactions**: no behavior change — retention stays at 48 hours, now via the explicit duration instead of `mempoolTxTimeoutHours`.
- `mempoolTxTimeoutHours: 48` is kept as a fallback in case the duration override is ever removed.

## Notes

- A transaction the relay silently drops (never mined, nonce never consumed) now stays visible as pending for the full 3 hours instead of 5 minutes — reconciliation deliberately does not evict on a not-found probe (see #1573). Until it expires, the sender's next send is built on top of its nonce; users can replace/cancel it manually. Watch `EthAlternativeMempoolTxResidence{action="timeout"}` after deploy — evictions clustering at the full 3h would mean the cache is mostly holding dead transactions.
- The reconcile loop probes each cached transaction against the provider once per minute for its whole lifetime, so Blink RPC usage grows with the longer window (bounded by private send volume). Nonce-lookup routing to Blink also holds for up to 3 hours per sender.
- Testnet configs are unchanged; Blink runs on mainnet only.

### DIFF
--- a/configs/coins/ethereum.json
+++ b/configs/coins/ethereum.json
@@ -63,6 +63,8 @@
                 "address_aliases": true,
                 "eip1559Fees": true,
                 "mempoolTxTimeoutHours": 48,
+                "mempoolTxTimeout": "3h",
+                "alternativeMempoolTxTimeout": "3h",
                 "queryBackendOnMempoolResync": false,
                 "fiat_rates": "coingecko",
                 "fiat_rates_vs_currencies": "AED,ARS,AUD,BDT,BHD,BMD,BRL,CAD,CHF,CLP,CNY,CZK,DKK,EUR,GBP,HKD,HUF,IDR,ILS,INR,JPY,KRW,KWD,LKR,MMK,MXN,MYR,NGN,NOK,NZD,PHP,PKR,PLN,RUB,SAR,SEK,SGD,THB,TRY,TWD,UAH,USD,VEF,VND,ZAR,BTC,ETH",
--- a/configs/coins/ethereum.json
+++ b/configs/coins/ethereum.json
@@ -63,7 +63,7 @@
                 "address_aliases": true,
                 "eip1559Fees": true,
                 "mempoolTxTimeoutHours": 48,
-                "mempoolTxTimeout": "3h",
+                "mempoolTxTimeout": "48h",
                 "alternativeMempoolTxTimeout": "3h",
                 "queryBackendOnMempoolResync": false,
                 "fiat_rates": "coingecko",
--- a/configs/coins/ethereum_archive.json
+++ b/configs/coins/ethereum_archive.json
@@ -72,7 +72,7 @@
                 "alternative_estimate_fee": "1inch",
                 "alternative_estimate_fee_params": "{\"url\": \"https://api.1inch.dev/gas-price/v1.5/1\", \"periodSeconds\": 10}",
                 "mempoolTxTimeoutHours": 48,
-                "mempoolTxTimeout": "3h",
+                "mempoolTxTimeout": "48h",
                 "alternativeMempoolTxTimeout": "3h",
                 "processInternalTransactions": true,
                 "queryBackendOnMempoolResync": false,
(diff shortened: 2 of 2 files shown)


pragmaxim (commented):
### The public-mempool cut is a Suite-visible RBF regression

`MempoolTxTimeoutDuration` honors `mempoolTxTimeout` unconditionally, so **every** `ethereum.json` deployment — including ones with no `ETH_ALTERNATIVE_SENDTX_URLS` — drops from 48h to 3h. The PR body calls this intentional; the consequence chain into Suite is worth making explicit, because it is not only "dropped public transactions stop lingering as pending for two days":

- Blockbook stops listing the pending tx → Suite's `analyzeTransactions` ([`transactionUtils.ts:737-760`](https://github.com/trezor/trezor-suite/blob/develop/suite-common/wallet-utils/src/transactionUtils.ts)) slices known pending txs not present in the fresh list into `remove` → the tx **disappears from Suite's local store**.
- `getEthereumRbfParams` requires `isPending(tx)` and the tx to be present in the list, so **the bump-fee / cancel affordance vanishes at 3h** — for exactly the stuck low-gas transaction that RBF exists to rescue, while the node still holds it and can still mine it. (EVM txs always report `rbf: true` — eth vins carry `Sequence: 0`, so `api/worker.go:593` sets it — so pendingness is the only gate.)
- The outgoing amount stops being reflected as pending in the UI while the funds are still committed on-chain.
- `getEvmPrivatePendingHint` (the Suite side of the draft `privatePending` hint, #1639) derives its nonces from that same pruned list, so **the wallet-declared hint can never outlive Blockbook's own mempool retention**. Relevant if #1639 is meant to eventually retire the `recentSenders` heuristic: shortening `mempoolTxTimeout` shortens the hint's usable lifetime with it.

Nonce safety is *not* affected: `account.misc.nonce` is the node's `pending` count, which still includes the stuck tx, and `getEvmNonceInfo` only clamps downward from locally-known pending nonces.

**Suggestion:** the Blink alignment only needs `alternativeMempoolTxTimeout >= 3h` and `mempoolTxTimeout >= that`. Shortening the *public* window is a separate policy decision that this PR doesn't need to make — and 3h is well below the 12h every other EVM chain uses, which is short for mainnet where a cheap tx legitimately sits for hours. If it stays, the reasoning should be recorded in `docs/config.md` so the value isn't silently reverted later.

For balance, the same knob has a real Suite-facing upside worth noting in the description: Blockbook has no replacement detection on the public path, so after an ETH fee bump the *replaced* txid lingers in the mempool until timeout and Suite renders it as a second pending tx at the same nonce. This change shrinks that ghost window from 48h to 3h.

pragmaxim (approved):
LGTM

PR #1711, opened 2026-08-10
fix(api): scale the in-sync tip tolerance to the chain's cadence

pragmaxim:
## Motivation

`blockbookApiInSync` paged twice for `arb-b4.trezor.io` on 2026-08-09 and the recovery was never announced. Investigating that turned up three independent problems; **this PR fixes the two that live in this repo.
What this PR does fix is why a healthy fast chain reports out-of-sync at all.

## 1. The tip tolerance was cadence-blind

`systemInfoSyncedGap` was a fixed **one block**. That is ~12s of slack on Ethereum, 0.25s on Arbitrum and 0.1s on Robinhood. The `/api/` `inSync` field this computes is what `blockbookApiInSync` probes at severity `high`, and the noise tracks chain speed exactly — minutes spent in that alert over 7 days:

| backend | cadence | minutes |
|---|---|---|
| arb-b4 | 0.25s | 203 |
| bsc-b4 | 0.45s | 168 |
| base-b4 | 2s | 111 |
| btc-b11 / ltc-b9 / pol-b4 | 150–600s | 1 each |

The tolerance is now the blocks the chain produces in 30s, floored at one block:

| cadence | coins | old | new | wall clock |
|---|---|---|---|---|
| 0.10s | Robinhood | 1 | 300 | 30s |
| 0.25s | Arbitrum | 1 | 120 | 30s |
| 0.45s | BSC | 1 | 66 | 29.7s |
| 1s | Avalanche, HyperEVM | 1 | 30 | 30s |
| 1.5s | Polygon | 1 | 20 | 30s |
| 2s | Base, Optimism | 1 | 15 | 30s |
| 3s | Tron | 1 | 10 | 30s |
| 12–13s | Ethereum, ETC | 1 | 2 | 24–26s |
| ≥60s | Dogecoin, Zcash, Litecoin, Bitcoin, Bcash | 1 | **1** | unchanged |

**The value changes only for chains at 15s/block or faster** — above 15s, `floor(30s/P) == 1` and the expression is a no-op, so every UTXO coin is untouched by arithmetic rather than by luck. (`floor(30/15) == 2`, so a 15s chain is the last one affected.)

The window is derived from the `systemInfoMinStale` constant, **not** from the computed staleness `threshold`. That distinction is the whole correctness argument: `12*P/P` is 12 blocks on *every* chain, which would silently stretch Bitcoin's tolerance to two hours of lag. A test row pins it.

Detection is not weakened. A full stall is `isFresh` (the index stopped advancing for `max(12P, 30s)`), which this PR does not touch. The 2026-08-09 arb-b4 excursion peaked at 211 blocks — 53s of real lag — and still reads out-of-sync; the 36- and 46-block samples around it no longer do.

## 2. `gap >= 0` was systematically wrong in the metric path

`GetSystemInfo` compares against a **live** `GetChainInfo` tip. `RefreshSyncMetrics` compares against the **cached** `is.BackendInfo.Blocks`, which `db/sync.go` refreshes once per resync iteration (`StartedSync → resyncIndex → updateBackendInfo → FinishedSync`) while `UpdateBestHeight` fires on every `writeHeight`. So the index being past that snapshot is the *designed steady state*, not an anomaly. Fraction of the last 24h with a negative gap:

| coin | negative |
|---|---|
| Avalanche | 99.8% / 99.7% |
| Robinhood | 69.4% / 68.2% |
| BSC | 12.3% / 6.9% |
| Arbitrum | 10.2% / 2.2% |

Rejecting it disabled the tip rescue on exactly those chains, making `blockbook_synchronized` systematically more pessimistic than the `/api/` field it mirrors — which matters now that `BlockbookNotSynchronized` consumes it.

Being past a *cached* tip only means blocks were connected after the snapshot, and `isFresh` still gates the rescue, so the gap is clamped rather than rejected. `backendBlocks > 0` is added alongside: without it the clamp turns `gap` into 0 at cold start and would rescue before any backend height was ever observed. That is the same condition `RefreshSyncMetrics` already applies before publishing `BackendBestHeight`.

Both changes sit inside `systemInfoInSync`, whose only callers are those two. Leaving both call sites untouched is what keeps the `/api/` field and the gauge on one computation so they cannot drift apart — and this change makes them *converge*, since the live path computes a small positive gap and the cached path a clamped 0, both now inside tolerance.

## Behaviour worth weighing

- **Live path only:** a backend rolled back or swapped to a lower height now reads in-sync for up to `max(12P, 30s)` longer, because `isFresh` is what ends it. That is the same window the function already tolerates elsewhere, and `blockbook_best_height` vs `blockbook_backend_best_height` still shows index-above-backend directly.
- A wrong `averageBlockTimeMs` now inflates the tolerance linearly. Not clamped here on purpose: if we want that guard it belongs on `blockPeriod` inside `syncBlockPeriod`, where it would cover all three consumers at once. `BlockbookIndexStalled` is height-vs-height and cadence-free, so it covers the blind spot meanwhile.
- Config-less altcoins whose observed cadence is 15s or faster (DigiByte ~15s → 2, NULS ~10s → 3) also gain tolerance — still ~30s of wall clock, still gated by `isFresh`.

## Testing

- `go build`, `go vet`, `gofmt` clean; `render_grafana.py --check` passes (107 panels, 89 metrics).
- Full suite **2237 tests / 70 packages**.
- `TestSystemInfoInSync` goes 16 → 26 rows. Two existing rows legitimately flipped to `true` under the new tolerance (a 10-block and a 2-block gap at 2s are 20s and 4s of lag, both now inside the window) and were retargeted past the window rather than deleted.
- The rows that would actually catch a wrong implementation: the 600s chain **not** widening past one block (the `threshold`-vs-constant trap and the UTXO constraint), a stale cached tip **not** being rescued (proving `isFresh` still gates the clamp), and no rescue before a backend height is known (the `backendBlocks > 0` guard). Boundaries are pinned on both sides at 250ms (120/121), 2s (15/16), 12s (2/3) and 600s (1/2).

## Rollout note

Only Arbitrum, BSC, Base and Ethereum currently run the post-#1694 build, which is deliberate. Those are also exactly the coins blipping, so this lands where it is needed. After deploy the 7-day minutes-in-alert query should collapse arb/bsc/base toward the ~1 min baseline — if it does not, the residual is a different defect and `for:` should **not** be raised to hide it.

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -2660,10 +2660,15 @@ const (
 	// and the floor is what lets a fast chain's configured cadence be corrected downwards
 	// without tightening this check.
 	systemInfoMinStale = 30 * time.Second
-	// systemInfoSyncedGap is how far the indexed height may trail the backend tip and
-	// still count as synchronized. It covers the one-block window between the tip
+	// systemInfoSyncedGap is the floor for how far the indexed height may trail the backend
+	// tip and still count as synchronized. It covers the one-block window between the tip
 	// advancing and that block being connected, which would otherwise flap the status.
 	systemInfoSyncedGap = 1
+	// systemInfoSyncedGapWindow turns that floor into wall clock. One block is 12s of slack
+	// on Ethereum but 0.25s on Arbitrum, where an excursion peaking at 211 blocks - 53s of
+	// real lag - read as out-of-sync and paged. Same value as systemInfoMinStale on purpose:
+	// under 30s is jitter for both checks.
+	systemInfoSyncedGapWindow = systemInfoMinStale
 )
 
 // systemInfoInSync decides the externally reported in-sync state from the raw
@@ -2692,13 +2697,28 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
 	}
 	isFresh := !lastBlockTime.Add(threshold).Before(now)
 
-	// A sync loop can stay inside ResyncIndex while new blocks keep arriving. If the
-	// indexed height is at (or within one block of) the backend tip and the index was
-	// updated recently, report the externally observable state as synchronized. int64
-	// avoids underflow if the backend momentarily reports a lower tip; gap >= 0 keeps an
-	// "ahead of tip" read from qualifying.
+	// Tolerate the blocks the chain produces inside the window, never fewer than one.
+	// Derived from the constant, not from threshold above: 12 block times divided by the
+	// block time is 12 blocks on every chain, which would stretch Bitcoin to two hours.
+	syncedGap := int64(systemInfoSyncedGap)
+	if lag := int64(systemInfoSyncedGapWindow / blockPeriod); lag > syncedGap {
+		syncedGap = lag
+	}
+
+	// A sync loop can stay inside ResyncIndex while new blocks keep arriving. If the indexed
+	// height is at (or close behind) the backend tip and the index was updated recently,
+	// report the externally observable state as synchronized. int64 avoids underflow if the
+	// backend reports a lower tip.
 	gap := int64(backendBlocks) - int64(bestHeight)
-	if !inSync && !initialSync && gap >= 0 && gap <= systemInfoSyncedGap && isFresh {
+	// A negative gap is the steady state for RefreshSyncMetrics, which compares against
+	// is.BackendInfo.Blocks - refreshed only at the end of a resync iteration: 99.8% of a
+	// day on Avalanche, 69% on Robinhood. Being past a cached tip only means blocks were
+	// connected after that snapshot, and isFresh still gates the rescue, so clamp rather
+	// than reject. backendBlocks > 0 keeps it from rescuing before any tip was observed.
+	if gap < 0 {
+		gap = 0
+	}
+	if !inSync && !initialSync && backendBlocks > 0 && gap <= syncedGap && isFresh {
 		return true
 	}
 
--- a/api/worker.go
+++ b/api/worker.go
@@ -2669,6 +2669,11 @@ const (
 	// real lag - read as out-of-sync and paged. Same value as systemInfoMinStale on purpose:
 	// under 30s is jitter for both checks.
 	systemInfoSyncedGapWindow = systemInfoMinStale
+	// systemInfoMinBlockPeriod floors the cadence the window is divided by, capping the
+	// derived tolerance at 300 blocks. 100ms is the fastest cadence any supported chain
+	// configures (Robinhood); anything smaller is a misconfigured averageBlockTimeMs or
+	// a degenerate observed average.
+	systemInfoMinBlockPeriod = 100 * time.Millisecond
 )
 
 // systemInfoInSync decides the externally reported in-sync state from the raw
@@ -2690,6 +2695,11 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
 	if blockPeriod <= 0 {
 		return inSync
 	}
+	// Neither input path validates the cadence beyond > 0, and the gap tolerance below
+	// scales linearly with it, so floor it to bound what a bad value can buy.
+	if blockPeriod < systemInfoMinBlockPeriod {
+		blockPeriod = systemInfoMinBlockPeriod
+	}
 
 	threshold := systemInfoStaleBlocks * blockPeriod
 	if threshold < systemInfoMinStale {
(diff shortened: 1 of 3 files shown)


api/worker.go
```diff
@@ -2685,13 +2690,28 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
...
+	// report the externally observable state as synchronized. int64 avoids underflow if the
+	// backend reports a lower tip.
 	gap := int64(backendBlocks) - int64(bestHeight)
-	if !inSync && !initialSync && gap >= 0 && gap <= systemInfoSyncedGap && isFresh {
+	// A negative gap is the steady state for RefreshSyncMetrics, which compares against
+	// is.BackendInfo.Blocks - refreshed only at the end of a resync iteration: 99.8% of a
+	// day on Avalanche, 69% on Robinhood. Being past a cached tip only means blocks were
+	// connected after that snapshot, and isFresh still gates the rescue, so clamp rather
+	// than reject. backendBlocks > 0 keeps it from rescuing before any tip was observed.
+	if gap < 0 {
```
cranycrane, line 2727: The comment says `isFresh still gates the rescue`, but on the path the removed `gap >= 0` guard was protecting against, `isFresh` is never false — the disconnect loop refreshes `LastSync` itself.

`db/sync.go` treats `ErrBlockNotFound` from a lagging RPC pool member as a fork and disconnects canonical blocks. Each disconnect runs `writeHeight(opDelete)` -> `is.UpdateBestHeight(height-1)`, which sets `LastSync = now`; `resyncIndex` then reconnects them via `FinishedSync`, which sets it again. So through the whole disconnect/reconnect churn `isFresh` is true on every sample, `gap` is negative, the clamp turns it into 0, `backendBlocks > 0` holds, and this returns `true` indefinitely. `/api/ inSync` and `blockbook_synchronized` both read healthy through a real index/backend divergence, which is exactly what `gap >= 0` used to report.

Worth being deliberate about the asymmetry this introduces: trailing lag is bounded (30s worth of blocks), but after the clamp the ahead direction is unbounded. Two ways to keep the fix and close it:

- clamp only for the cached-tip caller, so a live `ci.Blocks` below the index still fails the rescue in `GetSystemInfo` (where index > tip means a rollback, a replaced/pruned node, or a wrong-chain endpoint, not a stale snapshot); or
- keep the shared clamp but bound it, e.g. only clamp while `bestHeight - backendBlocks <= syncedGap`, so being slightly past a stale snapshot is tolerated and being far past is not.

api/worker.go
```diff
@@ -2685,13 +2690,28 @@ func systemInfoInSync(inSync bool, initialSync bool, chainType bchain.ChainType,
...
 
-	// A sync loop can stay inside ResyncIndex while new blocks keep arriving. If the
-	// indexed height is at (or within one block of) the backend tip and the index was
-	// updated recently, report the externally observable state as synchronized. int64
-	// avoids underflow if the backend momentarily reports a lower tip; gap >= 0 keeps an
-	// "ahead of tip" read from qualifying.
+	// Tolerate the blocks the chain produces inside the window, never fewer than one.
+	// Derived from the constant, not from threshold above: 12 block times divided by the
+	// block time is 12 blocks on every chain, which would stretch Bitcoin to two hours.
+	syncedGap := int64(systemInfoSyncedGap)
```
cranycrane, line 2713: `syncedGap` gets a floor but no ceiling, and `blockPeriod` is unvalidated on both input paths:

- `AverageBlockTimeDuration` only rejects `<= 0` (`bchain/coins/btc/bitcoinrpc.go:122`), so `averageBlockTimeMs: 25` instead of `250` on Arbitrum gives `30s / 25ms` = 1200 blocks — five minutes of real lag reported in sync — with nothing to flag it.
- The coins with no `averageBlockTimeMs` fall back to `AvgBlockPeriodSeconds`, derived from header timestamps that `computeAvgBlockPeriod` documents as non-monotonic. The wrap case is guarded; a small positive span is not, and a millisecond-scale average lands here as thousands of blocks of tolerance.

Before this change a bad `blockPeriod` was harmless, because its only consumer — `threshold` — was floored at `systemInfoMinStale`. Now it scales the tolerance linearly, so a config typo has unbounded reach.

`eth.TipStaleThreshold`, which the `systemInfoMinStale` comment cites as the precedent for clamping this kind of window, clamps `tipWatchdogMinStale` *and* `tipWatchdogMaxStale`. This derivation clamps neither end. A `systemInfoSyncedGapMax` around 120 (the value the fastest chain in the fleet already computes at 250ms) leaves every real chain untouched and makes the blast radius of a bad cadence finite.

cranycrane (commented):
Two findings from a review pass that I think are worth resolving before this merges; the rest were docs drift and polish.

cranycrane (approved):
LGTM

PR #1735, opened 2026-08-19
fix(api): serialize xpub cache updates to prevent data race

cranycrane:
## Problem

The process-wide xpub cache (`cachedXpubs`) stores `xpubData` **by value**, but the value's `addresses [][]xpubAddress` backing arrays are shared. On a cache hit the value is copied out under `cachedXpubsMux`, then the request mutates those shared arrays **outside the lock** (the scan reassigns `data.addresses[i]`, and `xpubCheckAndLoadTxids(&da[i], …)` does a read-modify-write on the per-address struct).

Two concurrent `getAccountInfo` requests with `details >= txids` for the **same descriptor** therefore race. Across a block connect, a straggler can overwrite fresher data, leaving the cached account **missing its newest confirmed txid** with a matching (too-low) `txs` count and a correct balance — a self-consistent but incomplete history served for ~1 block interval. Reachable unauthenticated via `GET /api/v2/xpub/<descriptor>` and WebSocket `getAccountInfo`.

## Fix

- **Per-descriptor serialization**: a 256-way sharded lock around `getXpubData`'s compute/update, so same-descriptor requests can't mutate one cache entry concurrently. Distinct descriptors stay fully parallel, and the caller's expensive per-tx response building runs after `getXpubData` returns, so it stays concurrent.
- **Copy-on-write**: clone the address slices before any mutation, so every published cache entry is an immutable, coherent snapshot and a reader holding a previous snapshot never has its arrays changed underneath.

## Test

`api/xpub_race_test.go` fires concurrent same-xpub `txids` requests over the existing fixtures. It passes under `-race` with the fix; reverting just `xpub.go` reproduces the write/write and read/write races (at `xpub.go:297`/`278`, matching the report).

Note: CI does not currently run `-race`; adding one `-tags 'unittest' -race` job would guard this class going forward.

### DIFF
--- a/api/xpub.go
+++ b/api/xpub.go
@@ -431,6 +431,39 @@ func setIsOwnAddresses(txs []*Tx, xpubAddresses map[string]struct{}) {
 	}
 }
 
+// getXpubDataCached returns the published snapshot without taking the
+// per-descriptor lock when the request cannot mutate it: same tip, same gap,
+// nothing to load. Keeps warm reads of a hot descriptor parallel.
+func (w *Worker) getXpubDataCached(xd *bchain.XpubDescriptor, option AccountDetails, filter *AddressFilter, gap int) (*xpubData, uint32, bool) {
+	cachedXpubsMux.Lock()
+	data, inCache := cachedXpubs[xd.XpubDescriptor]
+	cachedXpubsMux.Unlock()
+	if !inCache || data.gap != gap {
+		return nil, 0, false
+	}
+	bestheight, besthash, err := w.db.GetBestBlock()
+	if err != nil || besthash != data.dataHash {
+		return nil, 0, false
+	}
+	if option >= AccountDetailsTxidHistory {
+		if xpubTxidLoadNeeded(&data) {
+			return nil, 0, false
+		}
+		if data.mergedTxids == nil && isUnfilteredXpubTxidFilter(filter) {
+			return nil, 0, false
+		}
+	}
+	// bump accessed so the evictor does not drop a hot entry served fast-path only
+	now := time.Now().Unix()
+	cachedXpubsMux.Lock()
+	if current, ok := cachedXpubs[xd.XpubDescriptor]; ok {
+		current.accessed = now
+		cachedXpubs[xd.XpubDescriptor] = current
+	}
+	cachedXpubsMux.Unlock()
+	return &data, bestheight, true
+}
+
 func (w *Worker) getXpubData(xd *bchain.XpubDescriptor, page int, txsOnPage int, option AccountDetails, filter *AddressFilter, gap int) (*xpubData, uint32, bool, error) {
 	if w.chainType != bchain.ChainBitcoinType {
 		return nil, 0, false, ErrUnsupportedXpub
@@ -450,6 +483,9 @@ func (w *Worker) getXpubData(xd *bchain.XpubDescriptor, page int, txsOnPage int,
 	gap++
 	if err := validateXpubScanLimits(xd, gap, w.xpubConfig.MaxAddressDerivations); err != nil {
 		return nil, 0, false, err
+	}
+	if data, bestheight, ok := w.getXpubDataCached(xd, option, filter, gap); ok {
+		return data, bestheight, true, nil
 	}
 	// Serialize updates for this descriptor; the expensive per-tx response
 	// building in the caller runs after this returns and stays concurrent.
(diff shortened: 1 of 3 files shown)


api/xpub.go
```diff
@@ -485,6 +492,12 @@ func (w *Worker) getXpubData(xd *bchain.XpubDescriptor, page int, txsOnPage int,
 			}
 		}
 		processedHash = besthash
+		// Copy-on-write before any mutation so a concurrent reader holding the
+		// previously published snapshot never has its arrays changed underneath.
+		if !owned && (data.dataHeight < bestheight || fork || option >= AccountDetailsTxidHistory) {
```
pragmaxim: On a fully warm cache (`dataHeight == bestheight`, no fork), every `option >= AccountDetailsTxidHistory` request still clones the entire `addresses` matrix, even though `xpubCheckAndLoadTxids` will mutate nothing when every address is `complete` with `balance.Txs == txs` (or `balance == nil`). For a hot xpub polled for txid history this is an O(total derived addresses) allocation + copy on every request in steady state — at large gaps that is thousands of `xpubAddress` structs per call.

A cheap read-only pre-pass over `data.addresses` using the same "nothing to load" conditions `xpubCheckAndLoadTxids` checks would let the warm path skip the clone entirely.

api/xpub.go
```diff
@@ -451,10 +451,16 @@ func (w *Worker) getXpubData(xd *bchain.XpubDescriptor, page int, txsOnPage int,
 	if err := validateXpubScanLimits(xd, gap, w.xpubConfig.MaxAddressDerivations); err != nil {
 		return nil, 0, false, err
 	}
+	// Serialize updates for this descriptor; the expensive per-tx response
+	// building in the caller runs after this returns and stays concurrent.
+	mu := xpubUpdateLock(xd.XpubDescriptor)
```
pragmaxim, line 492: The shard mutex is held across the entire load phase — best-block reads, address derivation, and full txid history loading. Two serialization effects that were concurrent on master (racy, but parallel):

1. Warm reads of the same descriptor (`dataHeight == bestheight`, no fork, `option < TxidHistory`) that mutate nothing now run single-file behind any in-flight expensive first load of that xpub.
2. With 256 shards, an unrelated popular xpub that collides in the shard queues behind another xpub's multi-second first-time scan (e.g. gap=10000), which can push colliding warm requests toward client timeouts under load.

Same-descriptor serialization of *updates* is clearly the intent (and doubles as herd protection), so this may be an accepted tradeoff — but a no-mutation fast path (snapshot the entry under `cachedXpubsMux`, return without the shard lock when `besthash == dataHash` and no txid load is needed) would keep pure cache hits parallel, and it composes with the warm-path clone-skip suggested below.

api/xpub.go
```diff
@@ -485,6 +492,12 @@ func (w *Worker) getXpubData(xd *bchain.XpubDescriptor, page int, txsOnPage int,
 			}
 		}
 		processedHash = besthash
+		// Copy-on-write before any mutation so a concurrent reader holding the
+		// previously published snapshot never has its arrays changed underneath.
+		if !owned && (data.dataHeight < bestheight || fork || option >= AccountDetailsTxidHistory) {
+			data.addresses = cloneXpubAddresses(data.addresses)
+			owned = true
+		}
 		if data.dataHeight < bestheight || fork {
```
pragmaxim: `data.dataHeight < bestheight || fork` is now written twice, three lines apart — once inside the CoW guard on line 497 and once here as the rescan trigger. If the rescan condition ever gains a new invalidation trigger without the CoW guard being updated in lockstep, the code silently mutates shared cached arrays again — exactly the race this PR fixes — and the new regression test would not catch it, since its fixture chain is static and only exercises the `option >= AccountDetailsTxidHistory` branch of the CoW condition, never the new-block/fork branch.

Extracting `needsRescan := data.dataHeight < bestheight || fork` and using it in both places removes the drift hazard.

api/xpub_race_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return w, func() {
+		d.Close()
+		os.RemoveAll(tmp)
+	}
+}
+
+func resetXpubCache() {
+	cachedXpubsMux.Lock()
+	cachedXpubs = nil
```
pragmaxim: Setting `cachedXpubs = nil` breaks the nil-as-not-yet-initialized sentinel in `initXpubCache` (api/xpub.go:124-136): every subsequent `NewWorker` call in the test binary sees a nil map and spawns another never-terminating 20s eviction goroutine, each pinning its `*Worker` (including its already-closed RocksDB handle) and redundantly writing the `XPubCacheSize` metric. Test-process-only, but it violates a real invariant for every test that runs after this one. Resetting to an empty map preserves the sentinel:

```suggestion
	cachedXpubs = make(map[string]xpubData)
```

api/xpub_race_test.go
```diff
@@ -0,0 +1,164 @@
...
+			XPubMagic:             70617039,
+			XPubMagicSegwitP2sh:   71979618,
+			XPubMagicSegwitNative: 73342198,
+			Slip44:                1,
+		})
+	chain, err := dbtestdata.NewFakeBlockChain(parser)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tmp, err := os.MkdirTemp("", "xpubracedb")
```
pragmaxim: Only the `NewRocksDB` error path removes `tmp`; every later `t.Fatal` (LoadInternalState, ConnectBlock, GetMetrics, CreateMempool, NewTxCache, NewWorker) leaves the RocksDB handle open and the `xpubracedb*` directory on disk, because the cleanup closure is only returned at the very end. Using `t.TempDir()` (as `api/ethereumtype_test.go` already does) plus `t.Cleanup(func() { d.Close() })` registered right after `NewRocksDB` covers all failure paths and removes the manual cleanup return value.

cranycrane:
### Code review

53gur0 (approved):

PR #1736, opened 2026-08-19
fix(api): match the outpoint when resolving a spending transaction

### LINKED ISSUE #1029: Spending URL falsely showing found if the output is not spent but another output in the tx is
The spending URL will falsely showing `Found` even if the output is not spent but another one is of the same tx.

Not sure if this is intentional but it is misleading when using the /:vout: on the url.

### Scenario

I have a tx with 4 utxos, 0, 1 and 2 are spent but 3 isn't.

https://ltc1.trezor.io/spending/:hash:/1

Returns:
`<a href="/tx/:hash:">Found</a>.`

https://ltc1.trezor.io/spending/:hash:/3

Also returns ``<a href="/tx/:hash:">Found</a>.`` even if the output is not spent.

cranycrane:
Fixes #1029.

## Problem

Without `-extendedindex` there is no direct output → spender index, so `setSpendingTxToVout` finds the spender by walking the address index and matching on address, value and spending txid. It never checked *which* output the candidate input references.

So when one transaction pays the same value to the same address more than once, every one of those outputs resolves to whichever spender is found first. `/spending/:txid/:n` redirects to a transaction that does not spend that output, `spentTxId` on `/api/v2/tx/:txid?spending=true` names the wrong transaction, and `spentIndex` is equally arbitrary.

Reproduced on a public instance: LTC `d4864ba920c13213ac966b6cefc39021d2668d11d7e242e6c10e159ac5ebceeb` pays 800 LTC to `LMXrCNv5EmEpGNxXLJxq6KSoNzNRnRAQz7` at both vout 0 and vout 1. Both resolved to `0cdb7e5ffa9e8b6790762f9e28e53a853b296b381ddf62585c7a516fa6ed823e`, which has a single input and spends vout 0 only.

## Changes

- Require the candidate input to reference this vout, not just this transaction. This also fixes `spentIndex`, since the matched `index` is then really the input that spent this output.
- Return early from `GetSpendingTxid` for an unspent output. It has no spender and the scan has no early exit for that case, so it would walk the address index to the chain tip — and the change above removes the accidental early termination that a false match used to provide. `GetTransactionFromBchainTx` already guards the same call this way.

## Scope

Instances built with `-extendedindex` read the exact per-output `SpentTxid` and were never affected. The per-output `spent` flag is written by exact vout index and was always correct, so this never affected balances or UTXO selection — only which spender was named.

## Test

`api/spending_txid_test.go` runs against both index modes. The existing Fakecoin fixture already had the required shape: `TxidB1T1` pays `SatB1T1A2` to `Addr2` at both vout 1 and vout 2, and only vout 1 is spent. Before the change:

```
--- FAIL: TestGetSpendingTxid/legacyIndex/unspent_sibling_of_a_spent_output_is_not_spent
    GetSpendingTxid(00b2c060...3840, 2) = "7c3be240...9d25", want ""
```

After: pass. Every `extendedIndex` subtest passed both before and after, confirming the fault was isolated to the legacy path.

The test lives in `api` rather than `server` because it exercises the worker directly, and its tx cache is constructed with caching disabled, mirroring `server/public_test.go` — the fixture transactions carry no `Hex`, which `BitcoinLikeParser.PackTx` needs.

## Note for reviewers

Not changed here: `explorerSpendingTx` (`server/public.go`) shadows `err` with `spendingTx, err := s.api.GetSpendingTxid(...)`, so genuine errors from that call are discarded and always reported as "Transaction not found". Worth a separate fix.

### DIFF
--- a/api/spending_txid_test.go
+++ b/api/spending_txid_test.go
@@ -1,0 +1,159 @@
+package api
+
+import (
+	"os"
+	"testing"
+
+	"github.com/trezor/blockbook/bchain"
+	"github.com/trezor/blockbook/bchain/coins/btc"
+	"github.com/trezor/blockbook/common"
+	"github.com/trezor/blockbook/db"
+	"github.com/trezor/blockbook/tests/dbtestdata"
+)
+
+// setupSpendingWorker builds a worker over the two-block Fakecoin fixture. The
+// fixture already contains the shape this test needs: TxidB1T1 pays the same
+// SatB1T1A2 to Addr2 twice (vout 1 and vout 2) and only vout 1 is ever spent.
+func setupSpendingWorker(t *testing.T, extendedIndex bool) (*Worker, func()) {
+	t.Helper()
+	parser := btc.NewBitcoinParser(
+		btc.GetChainParams("test"),
+		&btc.Configuration{
+			BlockAddressesToKeep:  1,
+			XPubMagic:             70617039,
+			XPubMagicSegwitP2sh:   71979618,
+			XPubMagicSegwitNative: 73342198,
+			Slip44:                1,
+		})
+	chain, err := dbtestdata.NewFakeBlockChain(parser)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tmp, err := os.MkdirTemp("", "testdb-spending")
+	if err != nil {
+		t.Fatal(err)
+	}
+	d, err := db.NewRocksDB(tmp, 100000, -1, parser, nil, extendedIndex)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cleanup := func() {
+		d.Close()
+		os.RemoveAll(tmp)
+	}
+	config := &common.Config{CoinName: "Fakecoin", CoinShortcut: "FAKE"}
+	is, err := d.LoadInternalState(config)
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	d.SetInternalState(is)
+
+	bestHeight, err := chain.GetBestBlockHeight()
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	block1, err := chain.GetBlock("", bestHeight-1)
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	for i := uint32(0); i < block1.Height; i++ {
+		is.BlockTimes = append(is.BlockTimes, 0)
+	}
+	if err := d.ConnectBlock(block1); err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	block2, err := chain.GetBlock("", bestHeight)
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	if err := d.ConnectBlock(block2); err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	is.FinishedSync(block2.Height)
+
+	// the tx cache dereferences metrics unconditionally, so they must be real;
+	// the name must be unique or the prometheus registration collides
+	m, err := common.GetMetrics("Fakecoin-api-spending-" + t.Name())
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	// caching disabled, as in server/public_test.go: the fixture txs carry no Hex,
+	// which BitcoinLikeParser.PackTx needs, so a cached tx cannot be read back
+	txCache, err := db.NewTxCache(d, chain, m, is, false)
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	w, err := NewWorker(d, chain, bchain.NewMempoolBitcoinType(chain, 1, 1, 0, "", false, 1), txCache, m, is, nil)
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	return w, cleanup
+}
+
+// TestGetSpendingTxid covers trezor/blockbook#1029: an output must not inherit
+// the spender of a sibling output that happens to share its address and value.
+func TestGetSpendingTxid(t *testing.T) {
+	for _, extendedIndex := range []bool{false, true} {
+		name := "legacyIndex"
+		if extendedIndex {
+			name = "extendedIndex"
+		}
+		t.Run(name, func(t *testing.T) {
+			w, cleanup := setupSpendingWorker(t, extendedIndex)
+			defer cleanup()
+
+			tests := []struct {
+				name string
+				txid string
+				n    int
+				want string
+			}{
+				{
+					// spent by TxidB2T1 input 1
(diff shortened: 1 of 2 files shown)


api/spending_txid_test.go
```diff
@@ -0,0 +1,159 @@
...
+					// spent by TxidB2T1 input 1
+					name: "spent output resolves to its spender",
+					txid: dbtestdata.TxidB1T1,
+					n:    1,
+					want: dbtestdata.TxidB2T1,
+				},
+				{
+					// same address and same value as vout 1, but unspent: the
+					// address+value scan used to match TxidB2T1 here
+					name: "unspent sibling of a spent output is not spent",
```
pragmaxim, line 97: This case no longer exercises the outpoint half of the fix: with the new `!tx.Vout[n].Spent` early return, `GetSpendingTxid(TxidB1T1, 2)` returns before `setSpendingTxToVout` runs, and the n=1 case matches on txid alone (TxidB2T1 vin 1 is its only candidate). Revert only the `&& spentTx.Vin[index].Vout == uint32(vout.N)` condition and the whole suite stays green, so the #1029 regression is pinned only via the guard, not via the outpoint match — and the other caller of `setSpendingTxToVout` (`GetTransactionFromBchainTx` with `spendingTxs=true`) has no guard-based protection at all. To pin it, add a case that reaches the scan with a spent-but-wrong-outpoint sibling: e.g. extend the fixture so vout 2 is spent by a different tx, or call `w.setSpendingTxToVout` directly on vout 2.

api/spending_txid_test.go
```diff
@@ -0,0 +1,159 @@
+package api
```
pragmaxim, line 3: Missing `//go:build unittest` tag. Every other test file in package `api` (worker_test.go, worker_ens_aliases_test.go, xpub_test.go, ...) carries it, and unit tests run as `go test -tags unittest` (contrib/tests/run-unit-tests.sh, build/docker/bin/Makefile). Without the tag this RocksDB-backed test also compiles into untagged `go test ./api` runs, and it cannot use the tagged same-package helpers (e.g. `newRefreshTestMetrics`), which is what forces the hand-rolled metrics setup below.

api/spending_txid_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if err := d.ConnectBlock(block2); err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	is.FinishedSync(block2.Height)
+
+	// the tx cache dereferences metrics unconditionally, so they must be real;
+	// the name must be unique or the prometheus registration collides
+	m, err := common.GetMetrics("Fakecoin-api-spending-" + t.Name())
```
pragmaxim: This re-implements per-test metrics isolation that the package already has: `newRefreshTestMetrics(t)` (api/worker_test.go, behind the unittest tag) swaps in a private prometheus registry and restores it, which is exactly what the comment above works around. Uniquifying by `t.Name()` still registers ~60 collectors on the process-wide `DefaultRegisterer` per subtest and leaks them for the rest of the run, and it collides the moment the same test runs twice in one process (`go test -count=2`, `-shuffle` reruns) — `GetMetrics` then fails with an already-registered error unrelated to what the test covers. Once the `//go:build unittest` tag is added, this can just call `newRefreshTestMetrics(t)`.

api/spending_txid_test.go
```diff
@@ -0,0 +1,159 @@
...
+			XPubMagic:             70617039,
+			XPubMagicSegwitP2sh:   71979618,
+			XPubMagicSegwitNative: 73342198,
+			Slip44:                1,
+		})
+	chain, err := dbtestdata.NewFakeBlockChain(parser)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tmp, err := os.MkdirTemp("", "testdb-spending")
```
pragmaxim: Prefer `t.TempDir()` + `t.Cleanup` over `os.MkdirTemp` and a returned cleanup func — the pattern the neighbouring tests already use (`db.NewRocksDB(t.TempDir(), ...)` in api/ethereumtype_test.go, `t.Cleanup` in api/worker_ens_aliases_test.go). The manual variant forces a `cleanup(); t.Fatal(err)` pair at all ten error sites; any future branch that forgets the prefix leaks an open RocksDB handle and a temp dir, and callers must remember `defer cleanup()`. With `t.Cleanup(func() { d.Close() })` registered right after `NewRocksDB`, every error site collapses to a plain `t.Fatal(err)` and the helper can return just `*Worker`.

api/spending_txid_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	config := &common.Config{CoinName: "Fakecoin", CoinShortcut: "FAKE"}
+	is, err := d.LoadInternalState(config)
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	d.SetInternalState(is)
+
+	bestHeight, err := chain.GetBestBlockHeight()
```
pragmaxim: Lines 52–79 re-derive the fixture blocks through the fake chain (`GetBestBlockHeight` + two `GetBlock` calls plus error branches), but `dbtestdata.GetTestBitcoinTypeBlock1(parser)` / `GetTestBitcoinTypeBlock2(parser)` are exported and already used this way by db/rocksdb_test.go and api/worker_ens_aliases_test.go — two direct calls replace ~25 lines. The `BlockTimes` padding loop also does ~225k single-element appends per subtest (block1.Height is 225493); if it stays, pre-size it (`make([]uint32, block1.Height)`), though the ens-aliases test shows the same setup working from the exported block accessors.

api/spending_txid_test.go
```diff
@@ -0,0 +1,159 @@
...
+		t.Fatal(err)
+	}
+	// caching disabled, as in server/public_test.go: the fixture txs carry no Hex,
+	// which BitcoinLikeParser.PackTx needs, so a cached tx cannot be read back
+	txCache, err := db.NewTxCache(d, chain, m, is, false)
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+	w, err := NewWorker(d, chain, bchain.NewMempoolBitcoinType(chain, 1, 1, 0, "", false, 1), txCache, m, is, nil)
```
pragmaxim: The `NewMempoolBitcoinType(chain, 1, 1, 0, "", false, 1)` literal copies the constants from `fakeBlockChain.CreateMempool` (tests/dbtestdata/fakechain.go) — `chain.CreateMempool(chain)` keeps them in one place. Better yet, pass `nil`: `GetSpendingTxid` never touches the mempool, and `NewMempoolBitcoinType` spawns permanent sync goroutines that `cleanup()` never stops (its doc says it expects to be created once per process) — the two subtests leak them, which is noise under `-race`/goroutine-leak checks.

53gur0 (approved):

PR #1737, opened 2026-08-19
fix(api): page mempool and confirmed txs as one sequence

### LINKED ISSUE #1099: Inconsistent pagination behavior in v2 xpub endpoint
The following txids are returned for vpub5ZbEbAb1g4HMzYbxVj8tbWTjUUDkaRmp3qaQZDqHaHrJAobkvhXUB9BmgtdFAaFgfSTawDCVWccgw9DnN6cLK28zNLyJxEjvKuv8W9dqFJj

["b7a9c32218e1a33b40062f188c6ceafd2182b502779433fa2f2679cbf357208c","5e21743f5c4a8a94b4e0ca47adc6bfc96a3f01aa92c2c9a5ffe3f2f807aea4ad"]

https://blockstream.info/testnet/search?q=b7a9c32218e1a33b40062f188c6ceafd2182b502779433fa2f2679cbf357208c
https://blockstream.info/testnet/search?q=5e21743f5c4a8a94b4e0ca47adc6bfc96a3f01aa92c2c9a5ffe3f2f807aea4ad

Expected Behavior
When requesting a specific number of transactions via the pageSize parameter, the API should strictly adhere to this limit.

Actual Behavior
The API returns more transactions than specified by the pageSize parameter. This behavior is consistent across various settings of pageSize. Moreover, the additional transaction  does not exist.

cranycrane:
Fixes #1099.

## Problem

Account history prepends every mempool tx to page 1 and then appends a full `pageSize` of confirmed txs, so page 1 can hold `pageSize + unconfirmedTxs` items — the behaviour reported in #1099 (the "extra transaction that does not exist" there is a pending tx that was later dropped).

`GetAddress` capped this by truncating page 1 to `pageSize`, but page 2 still started after the dropped tx, so that confirmed tx was unreachable through any page. `GetXpubAddress` had no cap at all.

This matters for clients that store txs indexed by page position — Trezor Suite does, in `transactionsReducer` (`(page - 1) * perPage + i`). An over-long page 1 makes page 2 overwrite the last tx of page 1, which silently disappears from the account history and from CSV/PDF exports.

## Fix

Both paths now page over mempool entries followed by confirmed txids as a single sequence:

```
[ P1 … PM | C1 … CN ]
  page 1: items 0…24
  page 2: items 25…49   <- starts where page 1 stopped
```

- a page never exceeds `pageSize`
- `totalPages` counts pending txs, so the last confirmed tx stays reachable
- walking all pages returns every tx exactly once, pending first
- accounts with no pending txs page exactly as before

`Txs` and `unconfirmedTxs` keep their meaning; only `totalPages` changes. No client change is needed: Suite already pads its page count with the pending count and indexes by response position, so both become correct against a fixed server.

## Tests

- `api/paging_test.go` — window table plus a sweep over `mempool × confirmed × pageSize` asserting a full page walk reproduces the merged sequence with no oversized page.
- `api/worker_paging_test.go` — RocksDB-backed tests with a fake mempool for `GetAddress` (with/without mempool, mempool-only account) and `GetXpubAddress`.

Both regression tests fail on the pre-fix code in the reported ways: the xpub one with `page 1 returned 2 txids, more than the requested pageSize 1`, the address one by returning `[pending, tx3, tx1]` with `tx2` unreachable.

`./api/...` and `./db/...` pass. `./server/...` was not run in my sandbox (httptest cannot bind a port there); its fixture chain has no mempool txs, so those golden responses are expected to be unchanged, but that is worth confirming in CI.


api/worker.go
```diff
@@ -1670,15 +1702,17 @@ func (w *Worker) GetAddress(address string, page int, txsOnPage int, option Acco
 		if err != nil {
 			return nil, errors.Annotatef(err, "GetBestBlock")
 		}
-		var from, to int
-		pg, from, to, page = computePaging(len(txc), page, txsOnPage)
-		if len(txc) >= txsOnPage {
+		var from, to, mempoolFrom, mempoolTo int
+		pg, mempoolFrom, mempoolTo, from, to, page = computeAccountPaging(len(mempoolTxs), len(txc), page, txsOnPage)
```
pragmaxim: **Mempool-only addresses advertise pages they can never serve.** The pre-existing pin at line 1645 — `if ba == nil { ba = &db.AddrBalance{}; page = 0 }` ("if there are only unconfirmed transactions, there is no paging") — still forces `page = 0`, but this call now computes `TotalPages = ceil(mempool/txsOnPage) > 1` for a mempool-only address. A fresh (unindexed) address with 30 pending txs and `pageSize=10` returns `{"page":1,"totalPages":3}`, and `?page=2`/`?page=3` return the *same* first 10 txids (still `page:1`); pending txs 11–30 are unreachable, and a client walking `for page <= totalPages` collects duplicates. Pre-PR this was benign (`TotalPages` stayed 1), and `GetXpubAddress` has no such pin, so the two endpoints now disagree. The pin should be removed (or made conditional on an empty mempool); `TestGetAddressMempoolOnlyAccount` misses this because it uses 2 pending txs with `pageSize=25`.

api/worker.go
```diff
@@ -1670,15 +1702,17 @@ func (w *Worker) GetAddress(address string, page int, txsOnPage int, option Acco
...
+		if len(mempoolTxs)+len(txc) >= txsOnPage {
```
pragmaxim: **Adding the mempool count to this probe makes it fire on complete result sets, degrading exact `totalPages` to `-1`.** The old `len(txc) >= txsOnPage` was a proxy for "txc may have been truncated by `maxResults=(page+1)*txsOnPage`" — that depends only on the confirmed list, which is never truncated below `txsOnPage`, and the mempool is always fetched with `maxInt`. So when `len(txc) < txsOnPage`, `computeAccountPaging` already returned the exact `TotalPages`; adding `len(mempoolTxs)` makes the guard fire anyway, and for filtered queries (`totalResults < 0`, e.g. `?filter=inputs`) it overwrites the exact value with `-1`. Example: address with 1 filter-matching confirmed tx + 1 pending tx, `pageSize=2` → master returns `totalPages:1`, this branch returns `-1`, and the explorer (`getPagingRange` treats `-1` as "unknown") renders a spurious Next link for a one-page result. Keeping the probe as `len(txc) >= txsOnPage` fixes this with no downside. The same applies to `api/xpub.go:677` — where the whole guard is additionally vacuous, since `xpubCheckAndLoadTxids` internally always passes `maxInt`, so xpub `txc` is never truncated. Consider folding the totalResults/`-1` fix-up into `computeAccountPaging` itself: the 8-line block is now duplicated verbatim in both endpoints, and on the xpub side the `computePaging(totalResults+len(mempoolEntries), ...)` recompute provably reproduces the `pg` it discards (unfiltered `totalResults == len(txc)`).

openapi.yaml
```diff
@@ -1184,7 +1184,11 @@ components:
     PageSize:
       name: pageSize
       in: query
-      description: Number of history items per page. The default and maximum for REST account endpoints is 1000.
+      description: >-
+        Number of history items per page. The default and maximum for REST account endpoints
+        is 1000. Unconfirmed transactions take the first slots of the paged history and count
```
pragmaxim, line 1189: **"walking all pages returns every transaction exactly once" overclaims: confirmed offsets now depend on the live mempool count, so mempool churn mid-walk can silently skip a confirmed tx.** The confirmed window for page p is `[p*size - m, (p+1)*size - m)` with `m` re-read per request. A pending tx that confirms mid-walk is self-compensating (m drops, txc gains a head entry), but a pending tx that is RBF-replaced/evicted *without* confirming shifts the window forward and a confirmed tx falls between pages — the silent-skip variant of #1099, where master produced a detectable duplicate instead. Suggest qualifying the sentence ("for an account whose pending set does not change during the walk") and pointing multi-page walkers at `to=<bestHeight>` (which sets `m=0` and freezes the confirmed head) for a race-free walk.

api/worker.go
```diff
@@ -1072,6 +1072,32 @@ func computePaging(count, page, itemsOnPage int) (Paging, int, int, int) {
...
+func computeAccountPaging(mempoolCount, confirmedCount, page, itemsOnPage int) (pg Paging, mempoolFrom, mempoolTo, confirmedFrom, confirmedTo, normalizedPage int) {
+	pg, from, to, normalizedPage := computePaging(mempoolCount+confirmedCount, page, itemsOnPage)
+	mempoolFrom, mempoolTo = clampRange(from, to, mempoolCount)
+	confirmedFrom, confirmedTo = clampRange(from-mempoolCount, to-mempoolCount, confirmedCount)
+	return
+}
+
+// clampRange limits a window of the merged sequence to the bounds of one of its parts.
+func clampRange(from, to, length int) (int, int) {
+	if from < 0 {
```
pragmaxim: nit: go.mod declares `go 1.25.0`, so the four branches reduce to the builtins:

```go
func clampRange(from, to, length int) (int, int) {
	from = min(max(from, 0), length)
	return from, min(max(to, from), length)
}
```

Two of the four branches are unreachable given `computePaging`'s invariants (`0 <= from <= to`), and the min/max form makes the clamp ordering self-evident.

api/worker_paging_test.go
```diff
@@ -0,0 +1,296 @@
...
+	if _, found := c.txs[tx.Txid]; found {
+		return json.RawMessage("{}"), nil
+	}
+	return c.BlockChain.GetTransactionSpecific(tx)
+}
+
+// buildPagingWorker indexes confirmedTxs blocks, each with one tx paying Addr1, and puts
+// pendingTxs txs for the same address in the mempool. Returns the worker and the txids of
+// the pending and the confirmed txs, both newest first - the order the API returns them in.
+func buildPagingWorker(t *testing.T, confirmedTxs, pendingTxs int) (*Worker, []string, []string, func()) {
```
pragmaxim: **Test-fixture duplication:** `buildPagingWorker` (and `buildXpubPagingWorker` below) repeat the RocksDB fixture sequence that `buildHighFanInWorker` in `api/worker_test.go` already implements (same parser config, `MkdirTemp` + `NewRocksDB(tmp, 100000, -1, ...)` + `LoadInternalState`/`SetInternalState` + `InitBulkConnect`/`ConnectBlock`/`Close` + `FinishedSync` + cleanup closure) — three copies in the same package that must be edited in lockstep when the DB fixture changes. Suggest extracting a shared `newIndexedWorker(tb, blocks)` helper and using `t.TempDir()` (as `ethereumtype_test.go` already does), which removes the MkdirTemp/RemoveAll cleanup entirely. The xpub-magic parser literals also duplicate `setupChain` in `server/public_test.go`. Additionally, `pagingMempool` stubs the whole `bchain.Mempool` interface by hand — embedding `bchain.BaseMempool` (or using `CreateMempool` from the fake chain) would keep it compiling when the interface grows.

api/worker.go
```diff
@@ -1647,19 +1674,24 @@ func (w *Worker) GetAddress(address string, page int, txsOnPage int, option Acco
...
 						}
-						if page == 0 {
-							if option == AccountDetailsTxidHistory {
-								txids = append(txids, tx.Txid)
-							} else if option >= AccountDetailsTxHistoryLight {
-								setIsOwnAddress(tx, address)
-								txs = append(txs, tx)
-							}
+						if option >= AccountDetailsTxidHistory {
+							mempoolTxs = append(mempoolTxs, tx)
```
pragmaxim: **Memory retention:** this now keeps every decoded mempool `*Tx` (full Vin/Vout graphs) alive for the whole request on every page, where master let them become garbage inside the loop (the backend fetch itself was already unconditional, so no new I/O — only retention). For a hot address with thousands of pending txs, at most `txsOnPage` of these are ever emitted. Since the window depends only on `len(mempoolTxs)`, collecting txids (plus the count) and re-materializing only `[mempoolFrom:mempoolTo)` afterwards — or at minimum pre-sizing with `make([]*Tx, 0, len(txm))` — bounds the retained set.

53gur0 (approved):

PR #1738, opened 2026-08-19
fix(eth): stop indexing CALLCODE frames as internal transfers

### LINKED ISSUE #1225: Blockbook Incorrectly Indexes Fake ETH Transfers as Internal Transactions
Image

**Issue:**

We encountered a critical issue with Blockbook indexing internal Ethereum transactions. Specifically, Blockbook API reports internal transactions indicating ETH transfers even when no real ETH transfer has occurred on-chain. This problem arises due to the misuse of Ethereum’s delegatecall or deprecated callcode operations by malicious contracts.

Such exploits allow an attacker contract to falsely indicate that ETH has been transferred internally, causing Blockbook to index and report these fake transfers inaccurately as legitimate internal ETH transfers.

**Example:**

Here’s a representative transaction demonstrating this problem:

Transaction hash: [0xb247a9589bc1638dbd157937b121b20105d5d613d802ec0dab98017c861bfe39](https://eth1.trezor.io/tx/0xb247a9589bc1638dbd157937b121b20105d5d613d802ec0dab98017c861bfe39)

Blockbook reported an internal ETH transfer, but checking the Ethereum state explicitly (via account balance at the relevant blocks) confirms no actual ETH transfer occurred.

Image

The only real ETH transfer here is to address 0x66a0E978c0b91034A27D0da7207d5f80f11e86Dc all other ETH transfers originating from this address are fake.

**How to Reproduce:**

1. Locate a transaction involving delegatecall or callcode misuse.
2. Observe internal transfers reported by Blockbook.
3. Verify account balances explicitly at block levels to confirm no ETH has moved.

**Expected Behavior:**

Blockbook should verify or indicate clearly when internal transactions don't reflect actual ETH balance movements to avoid misleading consumers of the API.

**Suggested Fix:**

Implement explicit balance checks or enhance indexing logic to differentiate clearly between genuine ETH transfers and those resulting solely from internal EVM instructions without actual ETH movement.

**Impact:**

**High** - Users relying on Blockbook for Ethereum internal transfers may mistakenly process fake ETH deposits, causing significant financial and operational risks.

Thank you for promptly addressing this issue.

cranycrane:
Fixes #1225.

`processCallTrace` ignored `DELEGATECALL` frames but not `CALLCODE`, so a `CALLCODE`
with a non-zero value fell through to the generic branch and was indexed as an internal
transfer. `CALLCODE` runs foreign code in the caller's own context, so its traced value
never leaves the caller — the reported `from -> to` transfer never happened, and the
victim address ends up with a fabricated incoming transfer in its transaction list.

Confirmed on the tx from the issue, `0xb247a958…`. Its callTracer output contains 22
value-bearing `CALL` frames, 5 `DELEGATECALL` and 5 `CALLCODE`; the five phantom 2.999979 ETH
transfers reported by `/api/v2/tx/…` are exactly the `CALLCODE` frames:

```
CALL           0xa9d1e08c… → 0x66a0e978…   2.999979   (the only real transfer)
  DELEGATECALL 0x66a0e978… → 0x60f760bb…   2.999979   ← already ignored
    CALLCODE   0x66a0e978… → 0xa3b36b1e…   2.999979   ← was indexed
  … ×5, each with a different CALLCODE target
```

Filtering `CALLCODE` loses no real transfer, so no balance diffing is needed — the same
one-branch treatment as `DELEGATECALL` is enough.

The test builds that subtree as a fixture and asserts only the genuine `CALL` transfers survive.

Note: this only affects blocks synced after the change; fake transfers already in the index
remain until a reindex.

### DIFF
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1517,17 +1517,20 @@ func (b *EthereumRPC) processCallTrace(call *rpcCallTrace, d *bchain.EthereumInt
 			To:    call.To,
 		})
 		contracts = append(contracts, bchain.ContractInfo{Contract: call.From, DestructedInBlock: blockHeight})
-	} else if call.Type == "DELEGATECALL" || call.Type == "CALLCODE" {
-		// ignore DELEGATECALL (geth v1.11 the changed tracer behavior)
-		// 	https://github.com/ethereum/go-ethereum/issues/26726
-		// CALLCODE runs foreign code in the caller's own context too, so its traced value never
-		// leaves the caller; indexing from->to would report a fake transfer to the code address
-	} else if err == nil && (value.BitLen() > 0 || b.ChainConfig.ProcessZeroInternalTransactions) {
-		d.Transfers = append(d.Transfers, bchain.EthereumInternalTransfer{
-			Value: *value,
-			From:  call.From,
-			To:    call.To,
-		})
+	} else if call.Type == "DELEGATECALL" || call.Type == "CALLCODE" || call.Type == "STATICCALL" {
+		// DELEGATECALL and CALLCODE run foreign code in the caller's own context, so their
+		// traced value never leaves the caller (geth v1.11 tracer change, issues #26726, #1225);
+		// STATICCALL cannot transfer value at all
+	} else if call.Type == "CALL" {
+		if err == nil && (value.BitLen() > 0 || b.ChainConfig.ProcessZeroInternalTransactions) {
+			d.Transfers = append(d.Transfers, bchain.EthereumInternalTransfer{
+				Value: *value,
+				From:  call.From,
+				To:    call.To,
+			})
+		}
+	} else if err == nil && value.BitLen() > 0 {
+		glog.Warningf("processCallTrace: unknown call type %q with value in block %d, not indexed", call.Type, blockHeight)
 	}
 	if call.Error != "" {
 		d.Error = call.Error
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1517,9 +1517,11 @@ func (b *EthereumRPC) processCallTrace(call *rpcCallTrace, d *bchain.EthereumInt
 			To:    call.To,
 		})
 		contracts = append(contracts, bchain.ContractInfo{Contract: call.From, DestructedInBlock: blockHeight})
-	} else if call.Type == "DELEGATECALL" {
+	} else if call.Type == "DELEGATECALL" || call.Type == "CALLCODE" {
 		// ignore DELEGATECALL (geth v1.11 the changed tracer behavior)
 		// 	https://github.com/ethereum/go-ethereum/issues/26726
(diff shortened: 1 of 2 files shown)


bchain/coins/eth/ethrpc.go
```diff
@@ -1517,9 +1517,11 @@ func (b *EthereumRPC) processCallTrace(call *rpcCallTrace, d *bchain.EthereumInt
 			To:    call.To,
 		})
 		contracts = append(contracts, bchain.ContractInfo{Contract: call.From, DestructedInBlock: blockHeight})
-	} else if call.Type == "DELEGATECALL" {
+	} else if call.Type == "DELEGATECALL" || call.Type == "CALLCODE" {
 		// ignore DELEGATECALL (geth v1.11 the changed tracer behavior)
 		// 	https://github.com/ethereum/go-ethereum/issues/26726
+		// CALLCODE runs foreign code in the caller's own context too, so its traced value never
+		// leaves the caller; indexing from->to would report a fake transfer to the code address
 	} else if err == nil && (value.BitLen() > 0 || b.ChainConfig.ProcessZeroInternalTransactions) {
```
pragmaxim: This fall-through still indexes any *unknown* frame type with a parseable non-zero value as a real from->to transfer, so the ignore list has now grown twice for the same bug class (DELEGATECALL in the geth-1.11 era, CALLCODE here), and STATICCALL is only safe by accident (the tracer omits `value`, so `DecodeBig` errors and `err == nil` skips it). Since the tracer's type set is closed and already enumerated in the `rpcCallTrace` comment above, consider inverting this branch into an explicit `call.Type == "CALL"` allowlist (or drop-and-log unknown types): a future caller-context type carrying a value field (e.g. EIP-7069 EXTDELEGATECALL) would otherwise reopen the #1225 fake-transfer vector until a third patch plus reindex. Trade-off: an allowlist silently drops genuinely new value-moving types, so drop-and-log may be the safer middle ground. Non-blocking hardening suggestion.

bchain/coins/eth/ethrpc_trace_test.go
```diff
@@ -103,3 +105,62 @@ func TestGetInternalDataForBlockOmitsTraceTimeoutWhenUnset(t *testing.T) {
...
+	}
+	for i := range want {
+		if d.Transfers[i].From != want[i].From || d.Transfers[i].To != want[i].To ||
+			d.Transfers[i].Value.Cmp(&want[i].Value) != 0 || d.Transfers[i].Type != want[i].Type {
+			t.Errorf("transfer[%d] = %+v, want %+v", i, d.Transfers[i], want[i])
+		}
+	}
+}
+
+func hexToBig(t *testing.T, hex string) *big.Int {
```
pragmaxim: This helper re-implements `hexutil.MustDecodeBig` (available in the already-imported `hexutil` package, go-ethereum v1.17.0). Both call sites use compile-time constants, so `*hexutil.MustDecodeBig(threeEth)` is a drop-in replacement -- the `t.Fatalf` branch is unreachable for valid literals, and deleting the helper also lets the `math/big` import go.

53gur0 (approved):
lgtm

PR #1742, opened 2026-08-21
fix(sync): don't treat a fetchable block as a reorg without an expected hash

cranycrane:
## Problem

`avax.trezor.io` logged a warning for **every block**, roughly once a second:

```
W0820 09:33:17 sync.go:475] getBlockChain: block 93244979  still missing after 3 retries (last: hash , height 93244979: Block not found); rechecking chain state
```

It is not a missing block. QuickNode answers `eth_getBlockByNumber` for the tip with coreth's `cannot query unfinalized data`, which `avalanche/evm.go` maps to `ErrBlockNotFound`. On a single coreth node this cannot happen — with unfinalized queries disabled, `latest` returns the accepted block and `newHeads` subscribes to accepted heads — so the tip we observe comes from a different pool member than the one serving the numeric fetch.

Three separate defects turned that lag into a problem.

## 1. Empty `expectedHash` read as a reorg — `db/sync.go`

`getBlockChain` passes its loop hash to `shouldRestartSyncOnMissingBlock`, but ethereum-type chains leave `BlockHeader.Next` unset, so the hash is empty for every block after the first of a round. The probe compared a freshly fetched hash against `""`, always found them unequal, and yielded `errResync` — a full resync round per block, plus ~5x the minimum RPC volume against a metered endpoint. The double space in the log line is that empty hash.

Guarded early, ahead of the two probe RPCs. Affects every EVM chain, not just Avalanche.

## 2. Tip patience too short — `configs/coins/avalanche*.json`

`tipRecheckThreshold` 3 → 14 and `recheckThreshold` 10 → 14. At the sequential path's internal 250 ms cap, 3 allowed only ~500 ms before the recheck, inside the normal acceptance skew. Both must move or `ApplyMissingBlockRetryOverride` clamps the tip value back to 10. `retryDelayMs` stays at 1000 deliberately — `getBlockWorker` uses the raw value, so lowering it would speed up bulk-sync retries as a side effect. `avalanche.json` had no override at all and was inheriting the same defaults.

## 3. Sentinel destroyed by annotation — `bchain/coins/eth/ethrpc.go`

Found by deploying commits 1–2 to `blockbook-dev` and reading `/metrics`. The `juju/errors` pin has no `Unwrap`, so `errors.Annotatef` is an `errors.Is` barrier, and `getBlockRaw` annotated `ErrBlockNotFound` on its way out. That silently disabled the end-of-chain exit in `getBlockChain` (gated on `gotNotFound`) and the `IndexBlockNotFoundRetries` counter.

It only surfaces where a tip read fails with an *error* rather than a null result — on Avalanche, every tip read. Measured on dev with commits 1–2 only:

```
blockbook_index_block_not_found_retries          0        ← never fires
blockbook_index_resync_errors{error="get_block"} 1594     ← ~330 blocks, ~4.8/block
blockbook_synchronized                           0        ← with initial_sync = 0
blockbook_backend_best_height            93328239        ← frozen
blockbook_best_height                    93328380        ← advancing
blockbook_index_resync_duration_count            0
```

~4.8 retries per block is one 1.17 s block period divided by the 250 ms retry delay: the loop was spinning at every tip instead of returning. Because no `ResyncIndex` iteration ever completed, `updateBackendInfo` stopped running, so `blockbook_synchronized` sat at 0 with `initial_sync` 0 — the exact condition that metric's HELP text recommends alerting on. `/api/` still looked healthy because `GetSystemInfo` queries the backend live.

Fixed by returning the sentinel unannotated, matching what the null-result branch directly below already does. Genuine skew still retries, since `height > bestHeight` is false in that case. Also removes the `hash , height N:` prefix that made the original log line so hard to read.

## Reorg safety

Forks stay covered by `getBlockChain`'s `prevHash` check and by `resyncIndex`'s own hash comparison at `localBestHeight`. The parallel path never hits the guard — `getBlockWorker` always queues a real hash from `GetBlockHash`. The pre-existing `TestGetBlockChainMissingBlockChangedHashResyncs` still passes, confirming a genuine changed-hash reorg resyncs as before.

## Testing

Five tests added. Each was checked against a reverted fix and fails without it:

- `TestShouldRestartSyncOnMissingBlockIgnoresEmptyExpectedHash` — also asserts `bestHeightCalls == 0`, so the short-circuit cannot regress into "returns false but still probes".
- `TestGetBlockChainByHeightMissDoesNotResync` — drives the AVAX scenario end to end: fetchable on the third attempt, `Next` unset, must connect rather than resync.
- `TestGetBlockRawKeepsBlockNotFoundUnwrappable` — fails with `getBlockRaw error = hash , height 93244979: Block not found`, i.e. the production log line verbatim.
- `TestGetBlockRawAnnotatesOtherErrors` — non-sentinel errors keep their height annotation.
- `TestJujuAnnotatefHidesSentinelFromErrorsIs` — pins the reason the special case exists, and starts failing if juju is ever upgraded so the workaround can be removed.

`db` and `bchain/coins/avalanche` suites pass; `go vet` and `gofmt` clean. `bchain/coins/eth` has a pre-existing sandbox failure (`httptest: bind: operation not permitted`) unrelated to these changes, verified identical on a reverted tree.

## Follow-up (not in this PR)

A typed `ErrBlockNotAccepted` would let the sync loop tell "backend has not accepted this height yet" apart from "block genuinely absent", suppressing the remaining warnings at their source rather than widening the window, and letting `resyncIndex:277` stop returning it as a hard error. With commit 3 in place the sentinel now survives to the sync loop, so that work is unblocked.

The broader debt: ~213 `errors.Annotate*` call sites, 62 in `bchain/coins/eth`, every one an `errors.Is` barrier. `juju/errors` v1.x has `Unwrap`; bumping the 2017 pin is the systemic fix, but it needs an audit of which `Is`/`As` checks currently depend on the string fallbacks in `isRetryableGetBlockError`.

### DIFF
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1450,6 +1450,11 @@ func (b *EthereumRPC) processEventsForBlock(blockNumber string) (map[string][]*b
 	})
 	b.observeEthSyncRpcError(method, err)
 	if err != nil {
+		// juju's Annotatef has no Unwrap; keep the sentinel bare so getBlockChain's
+		// end-of-chain exit and retry accounting still see ErrBlockNotFound.
+		if stdErrors.Is(err, bchain.ErrBlockNotFound) {
+			return nil, nil, err
+		}
 		return nil, nil, errors.Annotatef(err, "%s blockNumber %v", method, blockNumber)
 	}
 	r := make(map[string][]*bchain.RpcLog)
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1419,6 +1419,11 @@ func (b *EthereumRPC) getBlockRaw(hash string, height uint32, fullTxs bool) (jso
 	}
 	b.observeEthSyncRpcError(method, err)
 	if err != nil {
+		// juju's Annotatef has no Unwrap, so annotating a sentinel hides it from
+		// errors.Is and disables getBlockChain's end-of-chain exit and retry accounting.
+		if stdErrors.Is(err, bchain.ErrBlockNotFound) {
+			return nil, err
+		}
 		return nil, errors.Annotatef(err, "hash %v, height %v", hash, height)
 	} else if len(raw) == 0 || (len(raw) == 4 && string(raw) == "null") {
 		return nil, bchain.ErrBlockNotFound
--- a/db/sync.go
+++ b/db/sync.go
@@ -495,8 +495,14 @@ func (w *SyncWorker) onRetryableMiss(retries *int, threshold int, label string, 
 		return false, nil
 	}
 	if *retries == threshold {
-		glog.Warningf("%s: block %d %s still missing after %d retries (last: %v); rechecking chain state",
-			label, height, hash, *retries, err)
+		// Without a hash shouldRestartSyncOnMissingBlock is a no-op; don't claim a recheck.
+		if hash == "" {
+			glog.Warningf("%s: block %d still missing after %d retries (last: %v); waiting for backend tip",
+				label, height, *retries, err)
+		} else {
+			glog.Warningf("%s: block %d %s still missing after %d retries (last: %v); rechecking chain state",
+				label, height, hash, *retries, err)
+		}
 	}
 	return w.shouldRestartSyncOnMissingBlock(height, hash)
 }
--- a/db/sync.go
+++ b/db/sync.go
@@ -458,6 +458,11 @@ func (w *SyncWorker) sendHashHeight(hch chan<- hashHeight, abortCh <-chan error,
 }
 
 func (w *SyncWorker) shouldRestartSyncOnMissingBlock(height uint32, expectedHash string) (bool, error) {
+	// Without an expected hash (the ethereum-type path walks by height) a fetchable
+	// block is no evidence of a reorg; getBlockChain's prevHash check catches forks.
+	if expectedHash == "" {
+		return false, nil
+	}
 	// When a block hash disappears at a given height, it can indicate a
 	// reorg/rollback, but on load-balanced EVM RPCs a single lagging backend can
 	// also report an older tip. Only restart immediately when another probe can
--- a/bchain/coins/eth/ethrpc_blocknotfound_test.go
+++ b/bchain/coins/eth/ethrpc_blocknotfound_test.go
@@ -1,0 +1,68 @@
+//go:build unittest
+
+package eth
+
+import (
+	"context"
+	stdErrors "errors"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/juju/errors"
+	"github.com/trezor/blockbook/bchain"
+)
+
+// mockBlockRawRPC returns a fixed error from CallContext, standing in for a backend
+// that rejects a tip read (Avalanche's "cannot query unfinalized data" is mapped to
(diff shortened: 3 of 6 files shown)


bchain/coins/eth/ethrpc.go
```diff
@@ -1419,6 +1419,11 @@ func (b *EthereumRPC) getBlockRaw(hash string, height uint32, fullTxs bool) (jso
 	}
 	b.observeEthSyncRpcError(method, err)
 	if err != nil {
+		// juju's Annotatef has no Unwrap, so annotating a sentinel hides it from
+		// errors.Is and disables getBlockChain's end-of-chain exit and retry accounting.
+		if stdErrors.Is(err, bchain.ErrBlockNotFound) {
```
pragmaxim, line 1424: The sentinel passthrough here fixes the block-fetch path, but the same laundering survives one method over: `processEventsForBlock` (line ~1453) wraps every error with `errors.Annotatef(err, "%s blockNumber %v", ...)`, and Avalanche's `AvalancheRPCClient.CallContext` maps `cannot query unfinalized data` to `bchain.ErrBlockNotFound` for **every** RPC method, `eth_getLogs` included. `GetBlock` returns that annotated error verbatim, so in `getBlockChain` `gotNotFound` is false again: the end-of-chain exit is skipped, `IndexBlockNotFoundRetries` is not counted, and (since `expectedHash` is empty on the by-height walk) the loop can only escape via the 60s `maxStall` -> `errResync` — the same symptom this PR fixes, reachable whenever the pool member serving `eth_getBlockByNumber` has accepted height H but the one serving `eth_getLogs` hasn't.

Minimal fix: the same sentinel carve-out in `processEventsForBlock`. (Cosmetic sibling: `GetBlockHeader` at ~1379 re-annotates the now-bare sentinel from `getBlockRaw`, though only API-path callers see that.) The deeper alternative is consumer-side — the two direct `stdErrors.Is(err, bchain.ErrBlockNotFound)` checks in `db/sync.go` (~702, ~970) could use `stdErrors.Is(errors.Cause(err), ...)` like `api/ethereumtype.go:69` already does, which would cover every juju annotator in one place.

db/sync.go
```diff
@@ -458,6 +458,11 @@ func (w *SyncWorker) sendHashHeight(hch chan<- hashHeight, abortCh <-chan error,
 }
 
 func (w *SyncWorker) shouldRestartSyncOnMissingBlock(height uint32, expectedHash string) (bool, error) {
+	// Without an expected hash (the ethereum-type path walks by height) a fetchable
+	// block is no evidence of a reorg; getBlockChain's prevHash check catches forks.
+	if expectedHash == "" {
```
pragmaxim, line 463: With this guard, the threshold-crossing warning in `onRetryableMiss` becomes misleading on exactly the path this PR targets: at `retries == threshold` it still logs `"...block %d %s still missing after %d retries (last: %v); rechecking chain state"` and only then calls into this function, which now short-circuits — so on the by-height walk the log claims a chain-state recheck that never happens, and the `%s` hash placeholder prints empty, keeping the double-space artifact (`block 93244979  still missing`) the PR description calls out. An operator debugging a tip stall will look for probe RPCs that were never sent.

Suggest making `onRetryableMiss` aware of the empty hash: skip or reword the message (e.g. "waiting for backend to accept height %d; resync only via stall deadline") when `hash == ""`, so the log matches what the code now does.

53gur0 (approved):

PR #1748, opened 2026-08-26
perf(evm): serve GetChainInfo backend identity from a 60s cache

cranycrane:
## Motivation

`EthereumRPC.GetChainInfo` calls `net_version` and `web3_clientVersion` on **every** invocation, and it sits on the request path of `/`, `/api/` and `/api/v2/` as well as of every sync iteration (`SyncWorker.updateBackendInfo`).

On EVM chains those two calls are its *only* backend round trips — the tip (`Blocks`, `Bestblockhash`, `Difficulty`) already comes from the subscription-fed `bestHeader` cache. And neither value feeds `systemInfoInSync`; they only fill in `ChainInfo.Chain` and `ChainInfo.Version`.

Both are near-immutable: a chain id cannot change for a live endpoint, and a client version changes only when the node is restarted onto a new build. Re-asking for them per request is pure overhead — and the cheapest amplification vector in the public HTTP API, since an unauthenticated hit on `/api/` costs two serialized RPCs against a rate-limited or paid provider.

## Change

A generic `TTLValue[T]` (`common/ttlvalue.go` — domain-free, so a future non-EVM user such as btc's `GetChainInfo` can reuse it without importing a coin package) serves `{chainID, clientVersion}` from a snapshot refreshed at most once per 60s:

- **fresh** → zero RPCs;
- **stale** → one background goroutine refreshes while every caller (including the one that triggered it) gets the previous snapshot immediately — no request ever blocks on an RPC that can sit until `b.Timeout`;
- **cold** → the first caller fetches synchronously and its error propagates as before; concurrent cold callers neither pile on (single-flight) nor block;
- **failing** → last-known-good is served and retries are paced at 5s (instead of a full TTL, so recovery from a long outage is noticed within seconds), until refreshes have been failing continuously for 5 minutes — then the error is reported.

It lives on `EthereumRPC` (constructed once in `NewEthereumRPC`), so every derived chain (eth, Arbitrum, Optimism, Base, BSC, Polygon, Tron, HyperEVM, Robinhood, Avalanche) is covered with no per-chain edits — no changes to the ten `Initialize()`/`Shutdown()` copies. Avalanche's extra `info.getNodeVersion` reuses the same `TTLValue`, with its own policy (the version is a label, never a reason to fail `GetChainInfo`) in its fetch func. A panicking fetch resets the in-flight flag via `defer` (and a panic in a background refresh is contained), so it cannot wedge the cache.

Two deliberate non-reuses: `singleflight.Group` (as in `multicall.go`) blocks every concurrent caller until the in-flight call returns, which contradicts the never-block contract here; and `consensusVersionMonitor` is an *eager* poller with a goroutine lifecycle for facts that must stay fresh with zero readers, while `TTLValue` is pull-driven and needs no shutdown hook. Dropping the unused fields of the `info.getNodeVersion` response also removes the repo's only `ava-labs/avalanchego` import (`go mod tidy` will drop the requirement; not run here).

## Before / after

On Arbitrum, whose ~250ms blocks drive the resync loop at the `resyncindexdebounce` floor of ~1.25s:

| | before | after |
|---|---|---|
| sync loop (~1.25s cadence) | 2 RPCs each → **~140k/day** | 2 RPCs/min → **2,880/day** |
| each `/`, `/api/`, `/api/v2/` request | **+2 RPCs** | **0** |
| unreachable backend | 2 failing RPCs per request | 2 per 5s, single-flight |
| worst snapshot age | live | ~61s |

Slower chains gain less on the sync path but the same on the request path.

## Deliberate behaviour changes

**A dead backend is still reported, 5 minutes late.** Because these were the only calls in EVM `GetChainInfo` that could observe an unreachable node, caching them unconditionally would have deleted the "backend down → `backendError` → `inSync=false`" signal `api/worker.go` relies on. So once refreshes have been *failing continuously* for 5 minutes, `GetChainInfo` returns the error, annotated with the failing duration. Below that, a transient blip no longer flips a whole instance out of sync over a cosmetic version string — mirroring what `blockbookAppInfoMetric` already does deliberately for the metric path. (Keying on failure duration rather than snapshot age means a quiet instance whose snapshot simply aged out never reports a healthy backend as down.)

**`inSyncMempool` rides on the same 5-minute window.** On `GetChainInfo` error, `GetSystemInfo` also forces `inSyncMempool=false` (api/worker.go). With standard EVM configs (`queryBackendOnMempoolResync: false`) the mempool loop never touches the backend itself, so the removed per-request RPCs were the only mechanism flipping `inSyncMempool` on a dead backend. Post-PR it reads true for up to 5 minutes of backend death, until the failing-refresh cutoff trips (`inSync` itself is rescued sooner by the tip-staleness demotion, ~12 block periods).

**A changed chain id is refused, not adopted.** The per-request `net_version` incidentally caught "this endpoint is now answering for a different chain". The refresh latches the first id it validated and treats any later flip as a failed refresh: the old identity keeps serving through the 5-minute grace (an LB briefly misrouting to a wrong-chain node self-heals silently), after which the error reaches `backendError`/`inSync=false` — consistent with `Initialize` refusing to start on an unexpected network id. Each detection is logged.

**Metrics volume.** `blockbook_rpc_sync_latency{method="net_version"|"web3_clientVersion"}` now gets ~1 sample/min instead of one per request. Same latency semantics, far fewer samples — worth a look at any panel that assumed request-rate volume there.

## Tests

`TestTTLValue` (`common`) — 7 subtests with an injected clock covering the state machine: TTL reuse, non-blocking background refresh, failure keeping the last value with 5s retry pacing, cold failure propagation, cold single-flight, and panic containment (cold and background). `TestIdentityFromCache` / `TestValidateBackendChainID` (`eth`) cover the eth policy (5-minute failing cutoff carrying both duration and backend error; chain-id latch). Passes under `-race`. `common`, `avalanche`, `tron` and `api` suites pass; `go vet` is quiet.


bchain/coins/eth/backendidentity.go
```diff
@@ -0,0 +1,121 @@
...
+		c.mu.Unlock()
+		return snap, nil
+	}
+	// One caller at a time talks to the backend and the rest keep serving the previous
+	// snapshot, so no request ever queues behind an RPC that may sit until the timeout.
+	// Suppressing retries for a whole TTL caps an unreachable backend at one call a minute.
+	if warm && (c.fetching || now.Sub(c.lastAttempt) < backendIdentityTTL) {
+		lastErr := c.lastErr
+		c.mu.Unlock()
+		return serveStaleIdentity(snap, now, lastErr)
```
pragmaxim: **Spurious `backend identity not refreshed` error against a healthy backend after an idle gap.**

`serveStaleIdentity` keys liveness on snapshot **age**, but the snapshot only refreshes on request traffic. On a quiet instance (EVM `GetChainInfo` refreshers are per-request, the sync loop only with `-sync`, and the app-info tick is 15 min), the snapshot routinely ages past `backendIdentityMaxAge`. When a burst then arrives, request A starts the inline fetch, and every concurrent request B hits this `c.fetching` branch with `lastErr == nil` and `age > 5m` — and gets `backend identity not refreshed for 6m…` while the backend is perfectly healthy. `GetSystemInfo` (api/worker.go:2798-2806) turns that into `backendError` + `inSync=false` + `inSyncMempool=false` for those responses → health-check/monitoring flap on every burst-after-idle.

Liveness should key on *duration of consecutive failed attempts*, not snapshot age — e.g. only report the error when `lastErr != nil`, or track `failingSince` and compare that against the max age.

bchain/coins/eth/backendidentity.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	c.fetching = true
+	c.lastAttempt = now
+	c.mu.Unlock()
+
+	fresh, err := fetch()
```
pragmaxim: **Once per TTL, one unlucky caller blocks inline for up to `b.Timeout` (25s in eth/arbitrum/avalanche configs) — and on Avalanche two such windows stack to ~50s.**

The triggering caller runs `fetch()` synchronously on the request path while everyone else gets the stale snapshot instantly. That unlucky caller can be a public `/api/` request or the sync goroutine (`db/sync.go` `updateBackendInfo`). On Avalanche it is worse: `AvalancheRPC.GetChainInfo` first pays the parent identity refresh (shared ctx, up to 25s), then `probeNodeVersion` with a **fresh** `b.Timeout` ctx (another 25s); both TTLs are 60s and seeded together, so the same request tends to draw both.

Since the machinery for serving stale already exists, the fix is small: once warm, set `fetching=true`, kick the fetch off in a goroutine, and return the stale snapshot immediately (keep the synchronous path only for the cold first call).

bchain/coins/eth/backendidentity.go
```diff
@@ -0,0 +1,121 @@
...
+	snap := c.snapshot
+	warm := snap != nil
+	if warm && now.Sub(snap.fetchedAt) < backendIdentityTTL {
+		c.mu.Unlock()
+		return snap, nil
+	}
+	// One caller at a time talks to the backend and the rest keep serving the previous
+	// snapshot, so no request ever queues behind an RPC that may sit until the timeout.
+	// Suppressing retries for a whole TTL caps an unreachable backend at one call a minute.
+	if warm && (c.fetching || now.Sub(c.lastAttempt) < backendIdentityTTL) {
```
pragmaxim: **Recovery after a >5min outage is delayed up to a full TTL.**

Once the snapshot is past `backendIdentityMaxAge`, every `GetChainInfo` errors — but this suppression branch keys only on `lastAttempt`, never on snapshot age. If the last failed attempt was 1s before the backend came back, all requests for the next ~59s keep getting `backend identity not refreshed` without re-probing, so blockbook reports `backendError`/`inSync=false` for up to a minute while every other backend RPC already succeeds. Pre-PR it recovered on the very next request; automated failback/monitoring now resolves ~1 min late.

Consider skipping (or shortening) the `lastAttempt` suppression once `now.Sub(snap.fetchedAt) > backendIdentityMaxAge`.

bchain/coins/eth/backendidentity.go
```diff
@@ -0,0 +1,121 @@
...
+		lastErr := c.lastErr
+		c.mu.Unlock()
+		return serveStaleIdentity(snap, now, lastErr)
+	}
+	c.fetching = true
```
pragmaxim: **A panic in `fetch()` wedges the cache permanently: `fetching` is never reset.**

`c.fetching = true` is cleared only after `fetch()` returns — there is no `defer`. `GetChainInfo` runs in contexts that recover panics (net/http per-connection recovery; explicit `recover()` in server/websocket.go), so a single panic escaping the fetch leaves `fetching=true` forever: every later call takes the suppression branch, serves stale for 5 minutes, then returns `backend identity not refreshed` errors until restart — a healthy backend reported down permanently. The same pattern wedges `nodeVersionProbing` in `avalancherpc.go` (freezing the reported avm version).

Fix: clear the flag in a `defer` (re-lock, `c.fetching = false`, unlock) installed right after setting it, so any panic in `fetch()` still resets the state before propagating. Same treatment for `nodeVersionProbing` in `nodeVersionCached`.

bchain/coins/eth/backendidentity.go
```diff
@@ -0,0 +1,121 @@
...
+		// A cold cache has nothing to fall back on, so the failure propagates as before.
+		if !warm {
+			return nil, err
+		}
+		return serveStaleIdentity(snap, now, err)
+	}
+	fresh.fetchedAt = now
+	// A changed chain id means the endpoint is answering for a different chain than the one
+	// Initialize validated - the whole index is then being built against the wrong backend.
+	if warm && snap.chainID != fresh.chainID {
```
pragmaxim: **A chain-id flip is adopted into the snapshot and then served as healthy.**

The comment on this branch says the condition means "the whole index is then being built against the wrong backend" — yet the response is one `glog.Errorf` per transition, after which `c.snapshot = fresh` adopts the new id, so `/api` immediately reports the new chain with no `backendError`, no metric, and subsequent refreshes are silent (snapshot now equals fresh). That is inconsistent with the startup policy (`Initialize` refuses to start on an unexpected network id) and with the severity the comment itself assigns.

Since a live endpoint never legitimately changes chain id, consider latching the validated id and returning an error (or at least keeping the old snapshot + re-logging), which routes the misconfiguration through the existing `backendError`/`inSync=false` machinery instead of a single log line scrolling past.

bchain/coins/avalanche/avalancherpc.go
```diff
@@ -139,6 +151,51 @@ func (b *AvalancheRPC) GetChainInfo() (*bchain.ChainInfo, error) {
...
+// cachedNodeVersion returns the avm version, re-querying the info endpoint at most once per
+// nodeVersionTTL: GetChainInfo serves / and /api/ on every request, and this value changes
+// only when the node is restarted onto a new build. Empty means never probed successfully.
+func (b *AvalancheRPC) cachedNodeVersion() string {
+	return b.nodeVersionCached(time.Now(), b.probeNodeVersion)
+}
+
+// nodeVersionCached holds the TTL rules; now and probe are injected so they can be tested
+// without sleeping or a live node.
+func (b *AvalancheRPC) nodeVersionCached(now time.Time, probe func() string) string {
```
pragmaxim: **Second hand-rolled copy of the eth TTL + single-flight + serve-stale state machine.**

`nodeVersionCached` is field-for-field isomorphic to `backendIdentityCache.get` introduced in this same PR (`nodeVersionMu`~`mu`, `nodeVersionProbing`~`fetching`, `nodeVersionAttempt`~`lastAttempt`, same 60s TTL, same lock dance), each with its own mirror test file. Both copies already share the same latent defect (no `defer` resetting the in-flight flag), which is exactly the class of fix that will land in one copy and not the other. And the package now carries a third in-flight-dedup idiom next to `singleflight` already used in `multicall.go`.

Since `AvalancheRPC` embeds `*eth.EthereumRPC`, one small exported helper in eth (e.g. `ttlValue[T]` with `get(now, fetch)`) covers both call sites — Avalanche's policy differences (never error, keep last value on empty probe) fit entirely in its fetch func.

bchain/coins/eth/backendidentity.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	// net_version and web3_clientVersion answer with values that change only when the
+	// backend is restarted onto a new build, so they are refreshed on this period instead
+	// of on every GetChainInfo - which serves /, /api/ and /api/v2/ on every request.
+	backendIdentityTTL = 60 * time.Second
+	// Past this age the cached identity stops standing in for a live backend: callers turn
+	// a GetChainInfo error into backendError and inSync=false, and on EVM chains these two
+	// calls are the only part of GetChainInfo that touches the backend at all.
+	backendIdentityMaxAge = 5 * time.Minute
```
pragmaxim: **Unstated consequence of the 5-minute window: `inSyncMempool` stays true while the backend is dead.**

The PR describes the delayed `backendError`/`inSync` signal, but there's a second flag riding on the same error path: on GetChainInfo error, `GetSystemInfo` also forces `inSyncMempool=false` (api/worker.go:2806). With standard EVM configs (`queryBackendOnMempoolResync: false`), `MempoolEthereumType.Resync` never touches the backend and `syncMempoolLoop` re-asserts `FinishedMempoolSync` every minute even with the backend down — so the removed per-request RPCs were the *only* mechanism flipping `inSyncMempool` on a dead backend. Post-PR it reads true for up to 5 minutes of backend death (`inSync` itself is rescued sooner by the tip-staleness demotion, ~12 block periods). Worth stating in the PR description at minimum, or shortening the effective blindness by having the mempool loop verify backend reachability.

bchain/coins/eth/backendidentity.go
```diff
@@ -0,0 +1,121 @@
...
+	lastAttempt time.Time
+	lastErr     error
+}
+
+// get returns the cached identity, refreshing it through fetch when stale. now is passed in
+// so the staleness rules can be tested without sleeping.
+func (c *backendIdentityCache) get(now time.Time, fetch func() (*backendIdentity, error)) (*backendIdentity, error) {
+	c.mu.Lock()
+	snap := c.snapshot
+	warm := snap != nil
```
pragmaxim: **Cold cache: single-flight and retry suppression don't apply, contradicting the documented invariant.**

Both guards are gated on `warm`, so while `snapshot == nil` (startup, or a backend unreachable since startup) N concurrent callers all set `fetching=true` and fetch in parallel — the "caps an unreachable backend at one call a minute" comment doesn't hold in exactly the state it describes, and every `/api/` request against a cold+dead backend still costs two RPCs waiting up to `b.Timeout` each. A failing cold caller also returns its error based on the stale `warm` local even when a concurrent fetcher has meanwhile populated `c.snapshot` (needless `inSync=false` blip), and overwrites `c.lastErr` next to a fresh snapshot.

Dropping `warm` from the suppression gate (cold suppressed callers returning `c.lastErr`) restores the invariant in the cold state too — no worse than pre-PR, but the comment currently promises more than the code delivers.

53gur0 (approved):

PR #1749, opened 2026-08-27
chore(eth): keep private transactions pending for the whole mempool retention

cranycrane:
A privately relayed transaction stuck behind a nonce gap has been mined ~40 hours after broadcast. At the previous `alternativeMempoolTxTimeout` of `3h` the cache dropped it long before that, so the sender saw nothing pending and had nothing to cancel while the transaction was still mineable.

This raises the alternative cache retention to each coin's `mempoolTxTimeout` — `48h` on ethereum, `12h` elsewhere — the longest value the retention ordering permits, so a private transaction stays visible and keeps its nonce slot reserved for as long as its address index lives.

### Why the guard had to move

`mempoolRetentionInverted` rejected `alternativeMempoolTxTimeout == mempoolTxTimeout`, so the two could never be aligned. Equality is not the inverted case it exists for: `SendRawTransaction` stores the cache entry before `AddTransactionToMempool`, so the cache timestamp is never the younger one, and every cache exit clears the wrapped mempool too. Only a cache configured to *strictly* outlive the mempool can leave a private transaction indexed nowhere while still served as pending (#1573), and that stays rejected.

### Observability

Whether a relay keeps answering `eth_getTransactionByHash` past its advertised pending window decides whether the raised retention is reachable at all, or whether the missing eviction (`alternativeMissingTxTimeout`, 2 min) takes the transaction first. That was previously unobservable — while the cache retention was shorter than the relay's window, every entry left before the question could arise.

Reconcile now logs the surfacing transitions — first null of a run, a run the relay ends by answering again, and each eviction — distinguishing a cache timeout reached with the relay still answering from one reached on nulls, with the sender and nonce so the gap is visible. `alternativeMissingTxTimeout` is deliberately left at its default until those logs say whether it needs to change.

Residence histogram buckets now run to 48h; they stopped at 12600s, so under the new retention every long-lived exit would have collapsed into `+Inf`.

### Commits

- `fix(eth)` — accept an alternative cache retention equal to the mempool's, plus docs
- `chore(configs)` — align `alternativeMempoolTxTimeout` with `mempoolTxTimeout` in all seven configs that set them
- `feat(eth)` — log alternative-provider surfacing transitions
- `chore(metrics)` — extend cache residence buckets


bchain/coins/eth/ethrpc.go
```diff
@@ -201,9 +201,10 @@ func (c *Configuration) AlternativeMissingTxTimeoutDuration() (time.Duration, er
 // sweep is the one exit that does NOT clear the cache: inverted, it drops a private transaction's
 // address index while the cache keeps serving its body as pending, and nothing reconciles the two. Only
 // an explicit mempoolTxTimeout can invert the pair, which is why that is rejected rather than warned
-// about.
+// about. Equal timeouts are not inverted: the cache stores a transaction before handing it to the
+// mempool, so its entry is never younger, and both stores stop serving it at the same age.
 func mempoolRetentionInverted(alternativeTimeout, mempoolTimeout time.Duration) bool {
-	return alternativeTimeout >= mempoolTimeout
+	return alternativeTimeout > mempoolTimeout
```
pragmaxim, line 207: **Equality relaxation re-opens a #1573-class window on rebroadcast of the same txid.** The justification "the cache stores a transaction before handing it to the mempool, so its entry is never younger" is false for a re-send of the same raw tx: `SendRawTransaction` unconditionally re-caches with a fresh gen (`cacheMempoolTransaction`, alternativesendtx.go:301), `insertMempoolTx`'s newer-slot scan skips the same txid (line 927) and overwrites the entry with `time = now` (line 935), while `AddTransactionToMempool` is a no-op for an existing txid and never refreshes `entry.time` (mempool_ethereum_type.go:161-181). So after a rebroadcast at t1 = t0 + gap, cache time t1 > mempool time t0. With the shipped equal retentions (48h/48h), the mempool sweep — the one exit that does not clear the cache — drops the address index at t0+48h while `GetTransaction` keeps serving the body as pending (its self-check at alternativesendtx.go:1034 uses t1) and `cachedNoncesFor` keeps the nonce floor raised until t1+48h: pending by txid, gone from the address view, for the whole rebroadcast gap. Rebroadcasting a stuck tx is ordinary wallet behavior, and this PR's own comments say txs sit pending for tens of hours. Pre-PR the derived +30min (or larger explicit) margin absorbed realistic gaps; the equal configs shrink it to zero. Fix directions: refresh (remove + re-add) the wrapped mempool entry on cache overwrite, or keep a strict margin (see the derived-default suggestion on ethereum.json).

docs/config.md
```diff
@@ -107,10 +107,10 @@ Good examples of coin configuration are
...
-            * `alternativeMissingTxTimeout` – How long a cached private transaction may stay missing from the relay's `eth_getTransactionByHash` (consecutive null answers) before Blockbook treats it as dropped or cancelled and evicts it, as a positive Go duration string. Defaults to **2 minutes** — the relay answers from one consistent store, so a null means the transaction is gone, and the tolerance only absorbs a transient relay fluke. Set it at or above `alternativePendingTxWindow` for a relay that does not surface accepted transactions over its whole window — that restores eviction by cache timeout only.
+            * `alternativeMempoolTxTimeout` – Optional alternative-provider transaction cache retention as a positive Go duration string. Defaults to `alternativePendingTxWindow`; override only to hold private transactions for a *different* period than the one in which they can still land. The coin configs set it equal to their explicit `mempoolTxTimeout`, the longest value the ordering above permits, so a private transaction is served as pending for as long as its index survives.
```
pragmaxim: **This sentence overpromises: with `alternativeMissingTxTimeout` left at its 2m default in every config, the raised retention is unreachable under the relay's documented behavior.** Per this same file, the relay answers `eth_getTransactionByHash` only over its ~3h window; past it every probe is null, so a still-mineable nonce-gapped tx exits as `provider_missing` at ~3h + 2m (+ the 15m probe backoff) — never reaching 12h/48h, misclassified under the dropped/cancelled exit, and the new 21600-172800s residence buckets stay empty. The motivating case (mined at ~40h, nothing pending after ~3h) reproduces exactly, and the very next bullet's own guidance (set `alternativeMissingTxTimeout` at or above the cache retention for a relay that does not surface as long as the cache holds) is applied by no config. The PR body says the deferral is deliberate pending log evidence — fine, but then this sentence (and the PR title's promise) should be scoped: "provided the relay keeps surfacing the transaction" / "once `alternativeMissingTxTimeout` is raised per the guidance below".

bchain/coins/eth/alternativesendtx.go
```diff
@@ -1161,28 +1171,44 @@ func (p *AlternativeSendTxProvider) reconcileMempoolTxs() {
...
 				continue
 			}
-			missingSince := p.markMissing(tx.txid, tx.tx)
+			missingSince, startedMissing := p.markMissing(tx.txid, tx.tx)
+			// Logged on the transition, not per probe: this line and the two below are the record of when
+			// the relay stopped answering for a transaction that may still be mineable - a nonce gap has
+			// delayed inclusion by tens of hours - so a relay outliving or undercutting its advertised
+			// window can be told apart from a genuine drop without reading metrics.
+			if startedMissing {
+				glog.Warningf("alternative mempool: %s no longer surfaced by the alternative provider%s, age %s; evicting after %s of null answers unless it reappears", tx.txid, slotLabel(tx.tx), age.Round(time.Second), p.missingTimeout())
```
pragmaxim: **This log announces a deadline the code does not honor for exactly the entries it targets.** The missing-run eviction check (line 1195) sits below the probe-backoff gate (`skipped_backoff`, lines 1137-1140), so it is only re-evaluated at the next due probe — 15 min for entries older than 1h (`probeInterval`, lines 86-95). Every entry at the ~3h relay-window boundary is in the 15m bucket, so this line promises "evicting after 2m0s of null answers" while the eviction log three lines later then reports "null answers for 15m0s" — a routine 7.5x contradiction between two adjacent lines, corrupting the relay-behavior measurement these logs were shipped to take. Log the actual horizon instead, e.g. "evicting at the next probe (no sooner than %s) unless it reappears" using `max(missingTimeout, probeInterval(age))`.

bchain/coins/eth/ethrpc.go
```diff
@@ -817,7 +818,7 @@ func (b *EthereumRPC) CreateMempool(chain bchain.BlockChain) (bchain.Mempool, er
 		if cacheEnabled && mempoolRetentionInverted(b.alternativeSendTxProvider.mempoolTxsTimeout, mempoolTxTimeout) {
 			// wrapped in ErrConfiguration so startup fails fast: this cannot resolve on a retry, and
 			// the retry loop would otherwise spend two minutes reporting it once a second as an RPC fault
-			return nil, fmt.Errorf("%w: mempoolTxTimeout=%s must be longer than the alternative-provider cache retention of %s, or the wrapped mempool drops a private transaction's address index while the provider cache still serves it as pending; raise mempoolTxTimeout or lower alternativePendingTxWindow", bchain.ErrConfiguration, mempoolTxTimeout, b.alternativeSendTxProvider.mempoolTxsTimeout)
+			return nil, fmt.Errorf("%w: mempoolTxTimeout=%s must not be shorter than the alternative-provider cache retention of %s, or the wrapped mempool drops a private transaction's address index while the provider cache still serves it as pending; raise mempoolTxTimeout or lower alternativePendingTxWindow", bchain.ErrConfiguration, mempoolTxTimeout, b.alternativeSendTxProvider.mempoolTxsTimeout)
```
pragmaxim: **The suggested remedy cannot work on any shipped config.** `AlternativeMempoolTxTimeoutDuration` (lines 181-186) prefers an explicit `alternativeMempoolTxTimeout` and falls back to the pending window only when it is unset — and after this PR all seven configs set it explicitly. An operator who hits this error (e.g. `mempoolTxTimeout: "24h"` on ethereum, where the cache retention is an explicit 48h) and follows "lower alternativePendingTxWindow" keeps failing with the same message. Name the knob that actually feeds the comparison: "raise mempoolTxTimeout or lower alternativeMempoolTxTimeout (or alternativePendingTxWindow when no explicit cache retention is set)".

configs/coins/ethereum.json
```diff
@@ -64,7 +64,7 @@
                 "eip1559Fees": true,
                 "mempoolTxTimeoutHours": 48,
                 "mempoolTxTimeout": "48h",
-                "alternativeMempoolTxTimeout": "3h",
+                "alternativeMempoolTxTimeout": "48h",
```
pragmaxim, line 67: **The "longest value the ordering permits" intent is encoded as seven hand-copied equal literals, and the guard only catches drift in one direction.** `mempoolRetentionInverted` rejects only `alternative > mempool`, so a later bump of `mempoolTxTimeout` (48h → 72h) that misses this line passes silently and detaches the retention from the documented "as long as its index survives" intent. A deeper form was available without relaxing the guard at all: `MempoolTxTimeoutDuration` (ethrpc.go:156-168) already derives the mempool retention as the explicit `alternativeMempoolTxTimeout` + 30 min when `mempoolTxTimeout` is unset — so setting only `alternativeMempoolTxTimeout` here (and dropping the explicit `mempoolTxTimeout`) yields 48h/48h30m, keeps the strict `>=` guard and its margin (which also absorbs the rebroadcast case flagged on ethrpc.go:207), and expresses the coupling once instead of in seven config pairs. If the explicit pair is preferred for operator visibility, consider at least a startup log/warning when the two detach.

bchain/coins/eth/alternativesendtx.go
```diff
@@ -1161,28 +1171,44 @@ func (p *AlternativeSendTxProvider) reconcileMempoolTxs() {
...
+			// that stops answering while a tx is still mineable is accommodated by configuring
+			// alternativeMissingTxTimeout at or above the cache retention, which restores timeout-only
+			// eviction.
+			// An entry at the cache timeout leaves as timeout whatever the final probe answered, so
+			// provider_missing stays the missing-run rule's own exit and its residence reads as "how
+			// long after the drop". With the retention configured past the relay's advertised window, an
+			// entry reaching the timeout was surfaced for nearly all of it - a null on the final probe is
+			// a coincidence, not the window closing - which is what the two timeout logs tell apart.
 			if timedOut {
+				glog.Warningf("alternative mempool: evicting %s at the cache timeout, last answer was null%s, age %s", tx.txid, slotLabel(tx.tx), age.Round(time.Second))
```
pragmaxim, line 1197: **A third timeout-eviction site stays outside the new boundary log stream.** The probe-error branch (lines 1142-1150) also evicts as `"timeout"` when `timedOut`, logging only the generic "eth_getTransactionByHash ... failed" line — no "evicting at the cache timeout", no slotLabel, no age, and the same generic line fires when a pre-timeout error retains the entry, so the two cases are indistinguishable. During a relay outage every entry crossing the retention boundary exits through that path, i.e. the new log record goes silent precisely when the relay's behavior is most in question. One line inside that `if timedOut` (e.g. "evicting %s at the cache timeout, provider unreachable%s, age %s") completes the record.

configs/coins/ethereum.json
pragmaxim, line 67: **16x retention without a matching probe-backoff rung multiplies long-tail relay traffic for exactly the population this PR targets.** `probeInterval` (alternativesendtx.go:86-95) still caps at 15 min for entries older than 1h — a cadence calibrated for the old 3h window — so a tx the relay keeps surfacing for tens of hours is probed ~190 times over 47h (vs ~45 over 3h), each probe a fresh `rpc.DialContext` + TLS handshake (`callHttpRawResult`) plus, via `transactionSupersededByNonce`, an `eth_getTransactionCount` to every configured URL whose memoization is only cycle-scoped. The ladder's own doc comment already argues this quota trade-off one rung down, and relay quota is treated as scarce (#1629). One more rung (e.g. `age > 3h → 1h`) cuts the long tail ~4x at the cost of up to 1h of drop-detection latency for a tx already stuck for hours; the timeout eviction is unaffected (checked before the backoff gate).

53gur0 (approved):
lgtm

cranycrane:
### Code review

Found 2 issues:

1. Stale operator guidance contradicts this PR's own corrected wording. The PR decouples the cache retention (now 12h/48h) from the 3h pending window and updates `docs/config.md` to say `alternativeMissingTxTimeout` should be set "at or above the cache retention" — but the doc comments on `defaultAlternativeMissingTxTimeout` and `AlternativeMissingTxTimeoutDuration`, plus two passages in `docs/evm-send.md`, still say to set it "at [or above] the pending window" to restore timeout-only eviction. On the raised coins that advice no longer achieves the documented effect (3h < 48h cache retention on ethereum).

https://github.com/trezor/blockbook/blob/abef906e6d3367f359b730569f81bec4322e1f7f/bchain/coins/eth/ethrpc.go#L69-L74

https://github.com/trezor/blockbook/blob/abef906e6d3367f359b730569f81bec4322e1f7f/bchain/coins/eth/ethrpc.go#L187-L192

https://github.com/trezor/blockbook/blob/abef906e6d3367f359b730569f81bec4322e1f7f/docs/evm-send.md#L120-L124

https://github.com/trezor/blockbook/blob/abef906e6d3367f359b730569f81bec4322e1f7f/docs/evm-send.md#L156-L160

2. Stale diagrams in `docs/evm-send-mempools.md`, a file this PR edits. The mermaid nodes still hardcode the cache at "timeout 3 h" and the wrapped mempool at "cache retention + 30 min", contradicting the retention table this PR updated a few lines above; the edge label still names `AddTransactionToMempool` where the cache path now calls `AddOrRefreshTransactionInMempool` (the restamp coupling this PR introduces); and the probe-interval node ("then 15 min") omits the new hourly tier past 3h.

https://github.com/trezor/blockbook/blob/abef906e6d3367f359b730569f81bec4322e1f7f/docs/evm-send-mempools.md#L53-L56

https://github.com/trezor/blockbook/blob/abef906e6d3367f359b730569f81bec4322e1f7f/docs/evm-send-mempools.md#L75-L77

https://github.com/trezor/blockbook/blob/abef906e6d3367f359b730569f81bec4322e1f7f/docs/evm-send-mempools.md#L114-L117

PR #1751, opened 2026-08-31
fix(tron): stop fabricating confirmed nonce

cranycrane:
Tron has no account nonces — replay protection uses a ref-block reference plus expiration, and java-tron's `eth_getTransactionCount` is a stub that always returns `0x0`. Blockbook's `TronClient.NonceAt` mirrors that as a local no-op, so `EthereumTypeGetNonces` never had a real value to report; it nevertheless returned `confirmedOK=true`, making the API fabricate `confirmedNonce: "0"` whenever a client requested a confirmed nonce.

This change turns `TronRPC.EthereumTypeGetNonces` into an honest constant stub (`0, 0, false, nil`): `nonce` stays `"0"` in responses (matching the backend's own stub), while `confirmedNonce` is no longer claimed. No wire traffic is affected — the old path never made an RPC call. Suite's Tron outdated check reads only balance + tronResources, so nothing consumes this field today.

### DIFF
--- a/bchain/coins/tron/tronrpc.go
+++ b/bchain/coins/tron/tronrpc.go
@@ -1184,20 +1184,11 @@ func (b *TronRPC) normalizeMulticallCalls(calls []bchain.EthereumMulticallCall) 
 	return normalizedCalls, nil
 }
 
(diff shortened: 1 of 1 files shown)


OriginalEveres (approved):
lgtm

PR #1752, opened 2026-08-31
feat(eth): bound input-data decoding to a max calldata size

cranycrane:
## Summary

`ParseInputData` decodes EVM transaction calldata into `ethereumSpecific.parsedData` on every transaction-detail response. The decode work (and per-signature allocations) grows linearly with the calldata length, which is bounded only by chain gas — a single transaction carrying a large dynamic array (e.g. `uint256[]`) can materialize one output value per element, costing tens of MB of transient allocation and hundreds of ms of CPU per request.

This adds a size budget so parsing cost is bounded by an operator-controlled limit rather than by the maximum calldata a chain can include.

## Changes

- Add `maxParseInputDataLen` (~64 KB calldata) and short-circuit `ParseInputData` above it, before the signature loop. Oversized input returns only the method id.
- Add a `Truncated` field to `EthereumParsedInputData`, serialized as `parsedDataTruncated`, so clients can tell an elided decode from an unknown selector.
- Add a regression test covering the oversized (truncated) and small (fully decoded) cases.

## Compatibility

The threshold is far above any normal call (transfers, swaps, stake/unstake are a few KB at most), so only pathological large-array inputs are affected. The truncated response — method id only, empty `name`, no `params` — is the same shape already returned for an unrecognized selector, and the raw `data` field is unchanged, so downstream clients degrade gracefully. The new field is additive and optional.

### DIFF
--- a/bchain/coins/eth/dataparser.go
+++ b/bchain/coins/eth/dataparser.go
@@ -96,6 +96,11 @@ func GetSignatureFromData(data string) uint32 {
 
 const ErrorTy byte = 255
 
+// maxParseInputDataLen bounds the hex length of calldata that ParseInputData will
+// decode. Real calls rarely exceed a few KB; ~64 KB of calldata is generous while
+// keeping parse work and per-signature allocations away from the chain gas limit.
+const maxParseInputDataLen = 128 * 1024
+
 func processParam(data string, index int, dataOffset int, t *abi.Type, processed []bool) ([]string, int, bool) {
 	var retval []string
 	d := index << 6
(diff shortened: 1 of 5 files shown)


cranycrane:
### Code review

Found 1 issue:

1. The new `parsedDataTruncated` field is added to the `ts_doc`-tagged `EthereumParsedInputData` struct, but the generated `blockbook-api.ts` (and the `openapi.yaml` schema) were not updated, so the checked-in TS types silently omit a field the API now returns. Prior field additions updated these in the same commit (e.g. 0ba2a066: "openapi.yaml and the generated blockbook-api.ts are updated in lockstep and verified by the OpenAPI parity typecheck"); because both files are missing the field, the one-directional parity typecheck passes and won't catch the drift.

New field: https://github.com/trezor/blockbook/blob/e9c43c6e447908543e72fc2f46d211ee41dbaa2f/bchain/types_ethereum_type.go#L51-L56

Stale generated interface: https://github.com/trezor/blockbook/blob/e9c43c6e447908543e72fc2f46d211ee41dbaa2f/blockbook-api.ts#L91-L100

Also verified against the underlying report: the `maxParseInputDataLen` guard runs before the signature loop and all decode work, `ParseInputData` is the single choke point for every reported trigger path (REST `/api/v2/tx/`, explorer, WebSocket `getTransaction`, plus Tron's parser which delegates to it), and the API response serializes `bchain.EthereumParsedInputData` directly so the `Truncated` flag reaches clients. The report's PoC calldata (~11.5M hex chars) is well over the 128 KiB cap.

OriginalEveres:
@coderabbitai review

53gur0 (approved):
as calldata can be re-fetched if needed.

PR #1753, opened 2026-08-31
feat(eth): batch contract metadata reads via Multicall3

cranycrane:
## Summary

- Resolve `name`/`symbol`/`decimals` for all index-unknown contracts of an address in **one Multicall3 `aggregate3` call** instead of up to three serialized `eth_call`s per contract, interleaved with the balance work. Chains without Multicall3 keep the sequential path unchanged.
- A **gas-starvation canary** (a cheap call to the zero address closing every chunk) distinguishes a genuine "not a token" revert from a starved sub-call; starved elements are re-read individually instead of being mis-recorded.
- Conclusive "no token at this address" verdicts go into an in-memory **negative probe cache** (LRU 8192, 15 min TTL expressed in blocks, invalidated on reorg generation change), so repeated requests stop re-probing dead/spam contracts the index can never describe. Chain read errors are never cached — `fetchContractInfo` now returns an explicit verdict instead of swallowing every eth_call error into "not a token".
- The ERC-4626 negative cache is generalized into the shared `negativeProbeCache`; the ERC-4626 path is a pure refactor onto it.
- Concurrent probes of the same contract are collapsed via singleflight; batched metrics keep `eth_call_contract_info` counts accurate.

## Test plan

- [x] `go test -tags unittest ./api/ ./bchain/coins/eth/ ./db/` — new coverage for the aggregate3 layout/canary/fallbacks (`contract_info_batch_test.go`), the prefetch + probe cache behavior (`contract_probe_test.go`), and the shared negative cache (`probe_cache_test.go`)


bchain/coins/eth/contract.go
```diff
@@ -446,55 +446,205 @@ func erc20BalanceOfCallData(addrDesc bchain.AddressDescriptor) string {
...
+func contractInfoFromAggregate3(address string, slots []bchain.EthereumMulticallResult) bchain.EthereumContractInfoResult {
+	var name string
+	if slots[0].Success {
+		name = strings.TrimSpace(parseSimpleStringProperty(slots[0].Data))
+	}
+	// a failed or blank name() is how the chain reports that there is no token here, and a
+	// reverting symbol() has never been surfaced as one either
+	if name == "" || !slots[1].Success {
+		return bchain.EthereumContractInfoResult{}
+	}
```
53gur0, line 637: Might make sense to rely on https://eips.ethereum.org/EIPS/eip-165 in upcoming PR instead of checking specific call

53gur0 (approved):

PR #1756, opened 2026-09-02
fix(sync): un-wedge steady-state EVM sync that falls behind the backend

cranycrane:
## Problem

On EVM chains that produce blocks faster than the sequential loop can index them, Blockbook diverges from the tip indefinitely. Observed live on Robinhood Chain (dev2): the backend produces ~10 blk/s while the sequential loop tops out at ~8.3 blk/s (serial `eth_getBlockByNumber` + `eth_getLogs` + `debug_traceBlockByHash` ≈ 121 ms/block), so the gap grew ~130k blocks/day with the instance reporting healthy sync activity.

Root cause: once `resyncIndex` enters the sequential `connectBlocks` round, `getBlockChain` only exits on reaching the tip, a fork, or a stalled block fetch. It never re-checks whether it has fallen behind, so the parallel path is never reconsidered — a restart temporarily fixes it (parallel catch-up), then it wedges again the moment it touches the tip.

## Changes

- **Fall-behind yield**: `getBlockChain` probes the backend tip every 100 connected blocks; if it is more than `max(4×workers, 64)` blocks behind, it yields `errResync` so `resyncIndex` re-evaluates and takes the parallel path. Yields are visible as `blockbook_index_sync_yields{reason="fell-behind"}`. Gated to EVM chains with `-workers > 1` (Bitcoin-type chains have no steady-state parallel path to yield to); probe errors are ignored and re-tried on the next batch.
- **Parallel pool size from config**: the steady-state EVM parallel path used a hardcoded pool of 4 workers, ignoring the `-workers` flag. It now uses the configured count capped by the block range. The trigger threshold stays at 4 blocks, decoupled from the pool size, so small steady-state rounds keep their current behavior.

## Testing

- New unit tests: yield-when-behind (exactly one tip probe, `errResync` after 100 blocks), no-yield-near-tip, gating (single worker / non-EVM never yield), and pool-capping table.
- Full `go test -tags unittest ./db/` and `go vet` pass.

### DIFF
--- a/configs/grafana/panels.yaml
+++ b/configs/grafana/panels.yaml
@@ -181,7 +181,7 @@ panels:
         legend: '{{type}}'
   sync.index_sync_yields:
     title: Index sync yields
-    description: The retry loop gave control back to outer resync because a deadline elapsed or chain-state probes kept failing.
+    description: Sync gave control back to outer resync because a deadline elapsed, chain-state probes kept failing, or the sequential loop fell too far behind the backend tip.
     queries:
       yields:
         promql: sum(increase({{name:index_sync_yields}}{job="blockbook", coin="$coin"}[$__interval])) by (reason)
--- a/configs/metrics.yaml
+++ b/configs/metrics.yaml
@@ -213,7 +213,7 @@ metrics:
   index_sync_yields:
     name: blockbook_index_sync_yields
     type: counter_vec
-    help: 'Number of times sync yielded to the outer resync machinery for non-reorg reasons (reason=deadline: MaxStallDuration elapsed in retry loop; reason=probe_failed: chain-state probe failed for the configured streak)'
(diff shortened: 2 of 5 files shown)


cranycrane:
### Code review

Found 1 issue:

1. The new `reason="fell-behind"` label value on `blockbook_index_sync_yields` is not reflected in the metric's documented enumeration. `configs/metrics.yaml` (the single source of truth for metric metadata) still describes only `deadline` and `probe_failed`, and the operator docs and Grafana panel description enumerate only those two causes, so the new yield path is invisible to anyone following them. The value is also hyphen-case while every sibling label value in this family is snake_case (`deadline`, `probe_failed`, `resync`, `get_block_hash`, ...) — consider `fell_behind` and updating the three doc sites.

https://github.com/trezor/blockbook/blob/99b5378f2e672341e189349dab106202c072ee68/db/sync.go#L1079-L1081

Documentation sites that still enumerate only the two original values:
- https://github.com/trezor/blockbook/blob/99b5378f2e672341e189349dab106202c072ee68/configs/metrics.yaml#L213-L217
- https://github.com/trezor/blockbook/blob/99b5378f2e672341e189349dab106202c072ee68/docs/sync.md#L173-L175
- https://github.com/trezor/blockbook/blob/99b5378f2e672341e189349dab106202c072ee68/configs/grafana/panels.yaml#L182-L188

53gur0 (approved):

Issue #1720, opened 2026-08-12
Improve 1inch fees for EVMs (focus on Ethereum)


tomasklim:
**Problem**

Blockbook supports two alternative EIP-1559 fee providers for EVM chains (`alternative_estimate_fee`): 1inch and Infura. Neither works well for Ethereum:

- **1inch (used previously) — too low.** Even after our tier remap (medium→low, high→medium, instant→high in `oneinchfees.go`), priority fees on a calm network are ~0.002 / 0.005 / 0.01 gwei and `maxFeePerGas` is only ~1.2 × baseFee — enough headroom when the base fee is high (10+ gwei), but not when it is low. The estimates track the current network state with virtually no margin, so any load uptick right after broadcast leaves the transaction stuck.
- **Infura (current) — too high.** Its medium tier suggests a ~1–2 gwei priority fee, an order of magnitude above what the network clears at most of the time, so users systematically overpay. We switched to it to stop the stuck transactions, at the cost of fees.

**Proposal**

Return to 1inch as the primary provider and post-process its response in blockbook:

1. **Priority-fee floor with jitter.** Per request, draw a random floor `F` uniformly from **[0.1, 0.15] gwei** with wei granularity (configurable via `alternative_estimate_fee_params`). Derive tiered floors from the single draw so tier ordering is preserved: low `max(tip, F)`, medium `max(tip, 1.5×F)`, high `max(tip, 2.5×F)` (multipliers open for discussion).
2. **Normalize `maxFeePerGas` with level-aware base-fee headroom.** 1inch's ~1.2 × baseFee padding survives less than 2 full blocks of base-fee growth (max +12.5 % per block) — then the transaction is unincludable until the base falls back. A flat multiplier like the common 2× fixes that but overshoots at high base fees: it inflates the displayed max cost and locks up balance (send-max). One possible shape — an absolute headroom floor combined with a moderate multiple:

   `maxFee' = baseFee + min( max(H, m × baseFee), c × baseFee ) + tip'`

   Illustration with `H = 0.5 gwei`, `m = 0.25`, `c = 1.5` (caps the effective multiple at 2.5×):

   | baseFee | headroom | maxFee (excl. tip) | effective multiple |
   |---|---|---|---|
   | 0.1 gwei | 0.15 | 0.25 | 2.5× (cap) |
   | 0.2 gwei | 0.3 | 0.5 | 2.5× (cap) |
   | 0.5 gwei | 0.5 | 1.0 | 2× |
[...]

PR #1770, opened 2026-09-07
feat(eth): fetch alternative fee provider data on demand

### LINKED ISSUE #1746: Fetch alternative fee provider data on demand instead of polling
Blockbook polls the alternative fee provider (Infura / 1inch) every `periodSeconds` (10 s on Ethereum ⇒ 8,640 requests/day per instance, ~26k/day across instances) and serves everyone the same cached values — all Trezor transactions within a window carry identical fees, and we burn API quota even when nobody is sending.

**Proposal**

Hit the provider only when a client actually requests an estimate:

- request coalescing (singleflight) + a short TTL (a few seconds) caps provider QPS at today's polling level even under burst; coalesced responses still get independent jitter draws (#1720)
- the existing fallback chain stays: provider down/slow → last fresh value within the stale window → on-chain `eth_feeHistory` estimation

**Benefits**

- per-user temporal fee diversity — each user samples the provider at their own moment, complementing the per-request jitter from #1720
- unless clients request more estimates per day than we currently poll, strictly fewer API requests — would let us drop back to a lower Infura plan tier again

cranycrane:
Closes #1746.

Replaces the per-instance polling goroutine of the EVM alternative fee providers (Infura, 1inch) with on-demand fetching:

- `GetEip1559Fees` serves the cache while it is fresh (within `periodSeconds`, now the cache TTL — no config changes); on expiry, one coalesced upstream fetch (`singleflight`) serves all concurrent callers, so upstream traffic stays capped at the old polling rate but only occurs under demand.
- Fallback chain unchanged: failed fetch → last value within `staleSeconds` → on-chain `eth_feeHistory`. Failed fetches are retried at most every 5s so an outage is not hammered at client-request rate.
- Returned fees are deep copies (`Eip1559Fees.Copy`), preparing for per-request jitter (#1720).
- One warm-up fetch at startup keeps the first request fast and surfaces a bad URL/key in the log at boot.
- New metric `blockbook_alternative_fee_provider_cache{provider,result}` (fresh/fetched/coalesced/stale/miss).
- Bitcoin providers (mempool.space) are out of scope and keep polling.

**Note for ops:** `blockbook_alternative_fee_provider_last_sync_timestamp_seconds` now only advances on demand — its age legitimately grows on a quiet instance, so correlate age alerts with the new cache counter.


cranycrane:
### Code review

Found 1 issue:

1. After a failed fetch, upstream retries are paced only by the fixed 5s `feeFetchRetryDelay`, not by `periodSeconds`. `refresh()` skips a fetch while `time.Since(lastSync) < ttl`, but a failure never advances `lastSync`, so once the provider errors (outage or HTTP 429) the TTL gate stays open and the only throttle is the 5s constant. With steady demand a chain like `bsc_archive` (`periodSeconds: 60`) is hit every 5s, 12x the configured cadence, exactly while the provider is throttling; the old poller retried once per period. This contradicts the docs added in this PR (`docs/config.md`: "upstream traffic is capped at one request per `periodSeconds`"; `docs/env.md`: "capping upstream traffic at one request per period") and the `PeriodSeconds` field comment ("contacted at most once per period"). Suggest pacing the post-failure retry off `max(feeFetchRetryDelay, ttl)`, or tracking a separate next-attempt deadline that advances on failure too.

https://github.com/trezor/blockbook/blob/2090c5b15fbe88b4ba2abebac72a398a97505e77/bchain/coins/eth/alternativefeeprovider.go#L141-L147

https://github.com/trezor/blockbook/blob/2090c5b15fbe88b4ba2abebac72a398a97505e77/docs/config.md#L108-L110

53gur0 (approved):

PR #1771, opened 2026-09-08
fix(sync): return from parallel connect when a worker exits without its block

### LINKED ISSUE #1767: sync: ParallelConnectBlocks deadlocks when a worker exits on a tail block without producing it (bsc-b4 stuck 2026-09-06)
## Summary

Steady-state EVM sync can deadlock silently in `ParallelConnectBlocks` (and `BulkConnectBlocks`). When a `getBlockWorker` hits a **non-retryable `GetBlock` error after the hash queue is closed**, it logs `getBlockWorker N connect block error ... Exiting...` and returns without producing its block. The writer goroutine then waits for that height forever, sibling workers holding later blocks are parked on their unbuffered `bch` sends, and the coordinator sits in a bare `wg.Wait()` that never returns. `abortCh` is only read inside the connect loop (already finished) and in the drain *after* `wg.Wait()`, so the error is never observed.

## Observed

`bsc-b4.trezor.io` (build `60ef2a1`, `-workers=16`) stopped at block 120372158 on 2026-09-06 21:11Z while the backend was at tip and the newHeads subscription, mempool loop and API all stayed healthy. Two `/metrics` scrapes six minutes apart showed every sync-side counter frozen (`rpc_sync_latency_count{eth_getBlockByHash|eth_getLogs|debug_traceBlockByHash}`, `index_resync_duration_count`), `go_goroutines` stable, and no `index_reorg_events` / `index_sync_yields` series at all — which rules out the reorg, stall-deadline and probe-failure exits (they all increment a metric) and leaves only the silent tail exit. A restart is the only recovery.

The stuck height being the round's first block matches the mechanism: the writer connected nothing before waiting on the missing height. The trigger is most likely an `eth_getLogs` rpc -32000 failure, which fails the whole `GetBlock` and is classified non-retryable.

## Reproduction

Unit test: 4 workers, range 1..4, `GetBlock` for height 1 returns a persistent non-retryable error, heights 2..4 succeed. `ParallelConnectBlocks(nil, 1, 4, 4)` never returns (the worker logs `Exiting...`, no `WriteBlock exiting...` follows). With a coordinator that selects on `abortCh` while waiting for the workers, it returns the worker error immediately.

## History

- 2018-09-21 `d2d5df88` (v0.1.0): original bulk-sync worker already had the return-without-block tail exit and the bare `wg.Wait()`; only reachable during initial sync.
- 2025-03-17 `0790f881` PR #1383 (v0.5.0): same code copied into `ParallelConnectBlocks` and run in steady state on every EVM chain whenever the tip is ≥4 blocks ahead — this made it reachable in production.
- 2026-02-26 `0703a69e` (v0.6.0): added the `abortCh <- err` send for the tail exit and a "Hardening" drain, but placed the drain after `wg.Wait()`, so it only helps when the exiting worker held the last block.
- #1756 (16 workers instead of 4) raises the odds of a sibling being parked when a worker exits.

## Proposed fix

Replace the bare `wg.Wait()` in both coordinators with a wait that also selects on `abortCh`; on abort, close `terminating` (guarded against double close, since the connect loop may already have closed it), then wait for the workers and return the error so `resyncIndex` retries. Also drop the writer closure's read of the coordinator's `err` variable, which is a data race and logs a misleading `ParallelConnectBlocks.Close error`.

## Ops fingerprint

`grep -E "connect block error .* Exiting|WriteBlock exiting|parallel connect aborted" blockbook.INFO` — an `Exiting...` line with no following `WriteBlock exiting...` is this bug.

cranycrane:
Fixes #1767

## Problem

`ParallelConnectBlocks` and `BulkConnectBlocks` wait for their workers with a bare `wg.Wait()`. A `getBlockWorker` that aborts after the hash queue is closed returns without producing its block. The writer waits on that height forever, sibling workers park on their unbuffered block sends, and the coordinator never reads `abortCh`. Steady-state EVM sync wedged this way on bsc-b4 on 2026-09-06 and needed a restart.

This affects every abort-then-return exit at the tail, not only the non-retryable error seen in the incident: probe failure, stall deadline and vanished-hash resync go through the same path.

## Fix

- New `waitWorkersOrAbort` helper used by both coordinators: waits for the workers while also selecting on `abortCh`. On abort it terminates the round, lets the parked workers drain, and returns the error so the sync loop retries. The `errResync` distinction is kept so reorgs still restart the resync.
- `terminating` is now closed through a `sync.OnceFunc`, since the connect loop may already have closed it when the tail abort arrives.
- Dropped the parallel writer's read of the coordinator's `err`, which was a data race and logged a misleading `ParallelConnectBlocks.Close error`.

## Validation

- `TestParallelConnectBlocksReturnsWhenTailWorkerExitsWithoutBlock` reproduces the issue scenario (4 workers, heights 1..4, height 1 fails non-retryably). It times out after 5 s on master with the `Exiting...`-without-`WriteBlock exiting...` fingerprint and returns the worker error in 40 ms with the fix.
- `TestWaitWorkersOrAbort` covers the helper in isolation.
- `go test -race -tags unittest ./db` passes; `go vet` clean under `unittest` and `integration` tags.

### DIFF
--- a/db/sync.go
+++ b/db/sync.go
@@ -601,6 +601,25 @@ func isRetryableGetBlockError(err error) bool {
 	return cause != nil && isRetryable(cause)
 }
 
+// waitWorkersOrAbort returns once every worker has exited, or as soon as a worker reports
+// an abort after the hash queue closed. Without the abort branch a worker that exits on a
+// tail block without producing it parks the writer and its siblings forever (#1767).
+func waitWorkersOrAbort(wg *sync.WaitGroup, abortCh <-chan error, terminate func()) error {
+	workersDone := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(workersDone)
+	}()
+	select {
+	case <-workersDone:
+		return nil
+	case abortErr := <-abortCh:
+		terminate()
+		<-workersDone
+		return abortErr
+	}
+}
+
 // ParallelConnectBlocks uses parallel goroutines to get data from blockchain daemon but keeps Blockbook in
 func (w *SyncWorker) ParallelConnectBlocks(onNewBlock bchain.OnNewBlockFunc, lower, higher uint32, syncWorkers uint32) error {
 	var err error
@@ -614,6 +633,8 @@ func (w *SyncWorker) ParallelConnectBlocks(onNewBlock bchain.OnNewBlockFunc, low
 	hchClosed.Store(false)
 	writeBlockDone := make(chan struct{})
 	terminating := make(chan struct{})
+	// The connect loop and the worker wait can both decide to stop the round; guard the close.
+	terminate := sync.OnceFunc(func() { close(terminating) })
 	// abortCh is used by workers to signal a resync-worthy reorg or a terminal worker error.
 	// Keep it buffered so the first worker can report without blocking while the
 	// coordinator is closing channels/terminating.
@@ -651,9 +672,6 @@ func (w *SyncWorker) ParallelConnectBlocks(onNewBlock bchain.OnNewBlockFunc, low
 				break WriteBlockLoop
 			}
 		}
-		if err != nil {
-			glog.Error("sync: ParallelConnectBlocks.Close error ", err)
-		}
 		glog.Info("WriteBlock exiting...")
 	}
 	for i := 0; i < int(syncWorkers); i++ {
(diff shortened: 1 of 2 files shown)


cranycrane:
### Code review

53gur0 (approved):

PR #1774, opened 2026-09-08
fix(eth): surface per-tx debug_traceBlockByHash errors instead of storing empty traces

### LINKED ISSUE #1760: Per-transaction debug_traceBlockByHash errors are silently swallowed, permanently under-indexing internal data
## Summary

`debug_traceBlockByHash` returns one envelope per transaction, carrying either `result`
or `error`. Blockbook's `rpcTraceResult` (`bchain/coins/eth/ethrpc.go:1579`) decodes only
`result`. When a single transaction fails to trace — most typically a tracer timeout on a
heavy transaction — its envelope decodes to a zero-value trace, the block is committed as
fully processed, and that transaction's internal data is lost permanently and without any
log line, metric, or queue entry.

## Current behaviour

geth returns (`eth/tracers/api.go`, mirrored by Erigon's rpcdaemon):

```go
type txTraceResult struct {
	TxHash common.Hash `json:"txHash"`
	Result interface{} `json:"result,omitempty"`
	Error  string      `json:"error,omitempty"`
}
```

Blockbook decodes:

```go
type rpcTraceResult struct {
	Result rpcCallTrace `json:"result"`
}
```

Walking a failed envelope through `getInternalDataForBlock`:

1. It decodes to a zero-value `rpcCallTrace`: `Type == ""`, `Calls == nil`.
2. The `len(trace) != len(data)` guard (`ethrpc.go:1657`) passes — the count is correct,
   one envelope per transaction; only its content is an error.
3. The loop at `ethrpc.go:1684` does nothing for that index: no `CREATE` row, no
   `processCallTrace` call, no error string. Note `rpcCallTrace.Error` (checked at
   `ethrpc.go:1696`) is the EVM-level revert reason of a call frame — a different thing;
   the envelope-level tracing failure has no representation at all.
4. `return data, contracts, nil` — success.
5. `GetBlock` therefore leaves `blockSpecificData.InternalDataError` unset
   (`ethrpc.go:1795`), and the block is stored as complete.
6. Nothing is written to `cfBlockInternalDataErrors`, so `HealInternalData` never retries
   it, `internal_data_error_queue` stays at 0, and `/internal` lists nothing.

Because the `timeout` in `traceConfig` is applied per transaction, the transactions that
fail are the heaviest ones — MEV bundles, batch routers, mass airdrops, exchange sweeps —
which are exactly the ones carrying the most internal transfers.

## Consequences

- **Transactions vanish from address histories.** `processInternalData`
  (`db/rocksdb_ethereumtype.go:602`) is what links an address to a transaction when the
  only involvement is an internal transfer. No internal data, no link. Native balance is a
  live `eth_getBalance` (`api/worker.go:1273`), so the balance stays correct while the
  transaction explaining it is absent — a "missing deposit" with no error to correlate.
- **Balance history is understated.** `processInternalTransactionsForBalanceHistory`
  (`api/worker.go:1816`) sums stored internal transfers into received/sent.
- **Transaction detail looks inert**: `internalTransfers: []`, and the internal error
  string is gone.
- **Contract registry holes**: no `CREATE` row (no `createdInBlock`, repeated chain
  probes, `not found in DB` warnings) and no `SELFDESTRUCT` row. It compounds — a later
  destruction for a contract with no row is dropped by `storeContractInfo`.
- **Invisible and permanent**: no retry, no metric, no queue entry; only a reindex
  repairs it, and nothing records that it happened, so past damage cannot be audited.
- **Instances diverge**: whether the data exists depends on whether that particular
  server's tracer timed out, so two Blockbooks behind a load balancer can return
  different histories for the same address.

No corruption: disconnect/reorg only undoes what was stored. This is pure silent
under-indexing.

## Proposed fix

```go
type rpcTraceResult struct {
	TxHash string       `json:"txHash"`
	Result rpcCallTrace `json:"result"`
	Error  string       `json:"error"`
}
```

On a non-empty envelope `Error`, have `getInternalDataForBlock` return an error naming the
block and the offending txHash. Everything downstream already exists: `GetBlock` sets
`InternalDataError`, the block is stored with its transactions and logs intact,
`StoreBlockInternalDataErrorEthereumType` queues it, and `HealInternalData` re-traces it
on the hourly pass, overlaying the result via `ReconnectInternalDataToBlockEthereumType`.

Block-granular retry means one bad transaction re-traces its whole block. That is the
right trade against under-indexing it forever, and the queue is already block-keyed.

Expected outcomes after the fix:

- Transient failures (node restart, momentary overload, load-balanced peer without the
  block) heal on the next pass.
- Deterministic failures (a transaction that genuinely exceeds `trace_timeout`) keep
  failing, but are now *visible* — queue gauge, `internal_data_heals{result="failed"}`,
  `/internal` listing — and, because the backoff caps at 2^6 passes and never abandons,
  they heal automatically once `trace_timeout` is raised or a faster backend is used. No
  reindex.

## Scope

`rpcTraceResult` is used only by `EthereumRPC.getInternalDataForBlock`, so this affects
every ethereum-type coin with `processInternalTransactions: true` (ethereum + sepolia/
hoodi, bsc, polygon, base, optimism, arbitrum, arbitrum nova, avalanche, hyperevm,
robinhood, robinhood testnet). Tron has its own internal data provider and does not use
this decode.

## Status

Code-level finding, not a confirmed production incident. It was found while investigating
`Contract ... not found in DB` warnings on eth.trezor.io, which turned out to have a
different cause (EIP-7702 delegated EOAs — see #1759); the transaction
examined there had zero `CREATE` frames, so nothing had been lost. By construction the
defect leaves no trace, so confirming it requires either adding the field and logging, or
re-tracing a sample of heavy transactions and diffing against stored internal data.

cranycrane:
Fixes #1760

## Problem

`debug_traceBlockByHash` returns one envelope per transaction that carries either `result` or `error`. Blockbook's `rpcTraceResult` decoded only `result`. When a single transaction failed to trace, its envelope decoded to a zero-value trace, the length guard passed, and the block was committed as fully processed.

Verified against client sources: geth's native callTracer fails the whole call on a per-tx error, so that path was already handled. **Erigon** writes the per-tx `error` field and keeps tracing the rest of the block, which is the live case for our Erigon-backed instances. bor does the same for the Polygon state-sync transaction.

## Consequences

- The failing transaction's internal transfers, `CREATE`/`SELFDESTRUCT` rows and address links were lost. Since the native balance is a live RPC, the account showed the right balance with the transaction that explains it missing.
- No `InternalDataError` was set, so nothing reached the heal queue, `internal_data_error_queue` stayed at 0, and `/internal` listed nothing. Only a reindex repaired it, and nothing recorded that it happened.
- Because the tracer timeout is per transaction, the losses concentrate on the heaviest transactions, which carry the most internal transfers.

## Fix

Decode the envelope `error` and return an error from `getInternalDataForBlock` naming the block and transaction. The field is decoded through a small `rpcTraceError` type because the two client families disagree on its shape: geth forks send a plain string, Erigon's `rpc.HandleError` sends a `{code, message}` object. Decoding Erigon's object into a string field would fail the whole call, which still queues the block for healing but loses the transaction hash and tracer reason from the log line and the `/internal` listing. Everything downstream already exists: `GetBlock` stores the block with its transactions and logs intact and sets `InternalDataError`, the block is queued, and `HealInternalData` re-traces it on the hourly pass.

The Polygon bridging transaction (from and to both zero address) keeps the exemption it already has in the trace-length adjustment, extracted into a small shared helper. Without it, a deterministic bor state-sync trace failure would queue every Polygon block containing one, and the heal loop never abandons entries.

## Why this solution

- **Reuses existing machinery end to end.** No new storage, metric, or retry path.
- **Transient failures self-heal** on the next pass. Deterministic ones become visible in the queue gauge, `internal_data_heals{result="failed"}` and `/internal`, and heal automatically once `trace_timeout` is raised or the backend gets faster. No reindex.
- Alternatives considered: recording the failure only in the transaction's error string keeps the data lost with no retry; re-tracing inline with `debug_traceTransaction` adds code on the sync hot path and still needs a fallback; failing `GetBlock` would stall sync.

## Tests

New cases in `ethrpc_trace_test.go`: a failing envelope returns an error naming the transaction, the same for Erigon's exact object-shaped wire form, a failing envelope for the bridging transaction is tolerated, and a decode table covering string, object, object without message, null and absent error fields. The `bchain/coins/eth` package passes locally apart from tests that bind a local HTTP server, which the sandbox forbids.

### DIFF
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1594,7 +1594,30 @@ type rpcCallTrace struct {
 type rpcTraceResult struct {
 	Result rpcCallTrace `json:"result"`
 	// tracer failure for the whole tx (typically a timeout), distinct from the EVM revert in Result.Error
-	Error string `json:"error"`
+	Error rpcTraceError `json:"error"`
+}
+
+// rpcTraceError tolerates both envelope shapes: geth forks send a plain string, Erigon a {code,message} object
+type rpcTraceError string
+
+func (e *rpcTraceError) UnmarshalJSON(b []byte) error {
+	var s string
+	if json.Unmarshal(b, &s) == nil {
+		*e = rpcTraceError(s)
+		return nil
+	}
+	var o struct {
+		Code    int    `json:"code"`
+		Message string `json:"message"`
+	}
+	if err := json.Unmarshal(b, &o); err != nil {
+		return err
+	}
+	if o.Message == "" {
+		o.Message = fmt.Sprintf("error code %d", o.Code)
+	}
+	*e = rpcTraceError(o.Message)
+	return nil
 }
 
 // isBridgingTx identifies the Polygon state-sync tx, which bor traces separately and unreliably
--- a/bchain/coins/eth/ethrpc.go
+++ b/bchain/coins/eth/ethrpc.go
@@ -1593,6 +1593,13 @@ type rpcCallTrace struct {
 
 type rpcTraceResult struct {
 	Result rpcCallTrace `json:"result"`
+	// tracer failure for the whole tx (typically a timeout), distinct from the EVM revert in Result.Error
+	Error string `json:"error"`
+}
+
+// isBridgingTx identifies the Polygon state-sync tx, which bor traces separately and unreliably
+func isBridgingTx(tx *bchain.RpcTransaction) bool {
+	return tx.To == "0x0000000000000000000000000000000000000000" && tx.From == "0x0000000000000000000000000000000000000000"
 }
 
 func (b *EthereumRPC) getCreationContractInfo(contract string, height uint32) *bchain.ContractInfo {
@@ -1682,8 +1689,7 @@ func (b *EthereumRPC) getInternalDataForBlock(ctx context.Context, blockHash str
 				for i := range transactions {
 					tx := &transactions[i]
 					// bridging transactions in Polygon do not create trace and cause mismatch between the trace size and block size, it is necessary to adjust the trace size
-					// bridging transaction that from and to zero address
-					if tx.To == "0x0000000000000000000000000000000000000000" && tx.From == "0x0000000000000000000000000000000000000000" {
+					if isBridgingTx(tx) {
 						if i >= len(trace) {
 							trace = append(trace, rpcTraceResult{})
 						} else {
@@ -1702,6 +1708,12 @@ func (b *EthereumRPC) getInternalDataForBlock(ctx context.Context, blockHash str
 			}
 		}
 		for i, result := range trace {
+			// a failed per-tx trace must not be committed as an empty one; the error queues the block for HealInternalData
+			if result.Error != "" && !isBridgingTx(&transactions[i]) {
+				e := fmt.Sprintf("trace of tx %s failed: %s", transactions[i].Hash, result.Error)
+				glog.Error("debug_traceBlockByHash block ", blockHash, ", error: ", e)
+				return data, contracts, errors.New(e)
+			}
 			r := &result.Result
 			d := &data[i]
 			if r.Type == "CREATE" || r.Type == "CREATE2" {
--- a/bchain/coins/eth/ethrpc_trace_test.go
+++ b/bchain/coins/eth/ethrpc_trace_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"strings"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
@@ -13,6 +14,7 @@ import (
 type mockTraceRPC struct {
 	method string
 	args   []interface{}
+	trace  []rpcTraceResult // canned debug_traceBlockByHash reply
 }
 
 func (m *mockTraceRPC) EthSubscribe(ctx context.Context, channel interface{}, args ...interface{}) (bchain.EVMClientSubscription, error) {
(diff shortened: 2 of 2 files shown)


bchain/coins/eth/ethrpc.go
```diff
@@ -1702,6 +1731,12 @@ func (b *EthereumRPC) getInternalDataForBlock(ctx context.Context, blockHash str
 			}
 		}
 		for i, result := range trace {
+			// a failed per-tx trace must not be committed as an empty one; the error queues the block for HealInternalData
+			if result.Error != "" && !isBridgingTx(&transactions[i]) {
+				e := fmt.Sprintf("trace of tx %s failed: %s", transactions[i].Hash, result.Error)
+				glog.Error("debug_traceBlockByHash block ", blockHash, ", error: ", e)
+				return data, contracts, errors.New(e)
+			}
```
OriginalEveres, line 1739: please check if the already processed transactions are being thrown away if there is a error in the tx...

eg.: block has 10 tx, first 5 are ok, 6th has error

cranycrane: yea the consumer of this method is doing
`if internalErr != nil || len(ens) > 0 || len(contracts) > 0 {`
therefore if there is an error, all the processed transactions are thrown away as desired

bchain/coins/eth/ethrpc.go
```diff
@@ -1593,6 +1593,36 @@ type rpcCallTrace struct {
...
+	}
+	if o.Message == "" {
+		o.Message = fmt.Sprintf("error code %d", o.Code)
+	}
+	*e = rpcTraceError(o.Message)
+	return nil
+}
+
+// isBridgingTx identifies the Polygon state-sync tx, which bor traces separately and unreliably
+func isBridgingTx(tx *bchain.RpcTransaction) bool {
```
53gur0, line 1638: nit: I wouldn't call it bridging transaction

53gur0 (approved):

PR #1775, opened 2026-09-08
fix(evm): undo token holdings in reverse order when disconnecting a block

### LINKED ISSUE #1617: NFT Rollback Bug Causes Incorrect Owner State and DB Corruption on Block Reorganization
## Overview
A bug in Blockbook's block-reorganization (rollback) handling for Ethereum-type blockchains (including shared storage blockchains like Tron) causes database state corruption. When a block is reorganized out, Blockbook rolls back the transactions in the **wrong order** (forward instead of reverse) and uses **lossy inverse arithmetic**. This can result in permanent database corruption where multiple addresses are recorded as simultaneously owning the same NFT, or ERC1155 token balances are incorrect.

---

## Affected Code
* **File:** [db/rocksdb_ethereumtype.go](db/rocksdb_ethereumtype.go)
* **Locations:**
  * [db/rocksdb_ethereumtype.go:402](db/rocksdb_ethereumtype.go#L402) - `addToContract()` (Handles ERC721/ERC1155 balance arithmetic)
  * [db/rocksdb_ethereumtype.go:1326](db/rocksdb_ethereumtype.go#L1326) - `disconnectBlockTxsEthereumType()` (Iterates forward over block transactions)

---

## Root Cause

1. **Forward-Order Rollback:** [disconnectBlockTxsEthereumType](db/rocksdb_ethereumtype.go#L1326) iterates through block transactions in the original forward order. To properly roll back a block, transactions must be processed in **strict reverse order**.
2. **Lossy Inverse Arithmetic:** 
   - **ERC721 (NFTs):** Disconnects call `c.Ids.remove()`. If the token is already absent (e.g., from a subsequent transfer in the same block), this removal is a no-op.
   - **ERC1155 (Multi-Tokens):** Negative balances are clamped to zero during subtraction (`s.SetUint64(0)`). This clamping causes information loss that cannot be recovered by simple inverse addition during rollback.

---

## Concrete Scenario (A -> B -> C)

With initial owner **A**, and consecutive transactions in an orphaned block:
1. `Tx 1: A → B`
2. `Tx 2: B → C`

Under the current forward-order disconnect logic:
* **Disconnect Tx 1:** Restores `X` to `A`. Tries to remove `X` from `B` (no-op, as `B` doesn't have it).
* **Disconnect Tx 2:** Restores `X` to `B`. Removes `X` from `C`.
* **Result:** Both `A` and `B` now record ownership of token `X` in RocksDB.

In contrast, a **reverse-order disconnect** correctly restores only `A` as the owner.

---

## Impact
* **State Corruption:** The corrupted RocksDB index survives service restarts and remains incorrect until a full, manual reindex of the database is performed.
* **Incorrect Balances:** Client applications (such as Trezor Suite) and public APIs will show incorrect NFT ownership and ERC1155 balances.

---

### 1. Reverse-Order Disconnection (Immediate Fix)
Modify [disconnectBlockTxsEthereumType](db/rocksdb_ethereumtype.go#L1326) to loop over block transactions in reverse:
```diff
-	for i := range blockTxs {
-		blockTx := &blockTxs[i]
+	for i := len(blockTxs) - 1; i >= 0; i-- {
+		blockTx := &blockTxs[i]
```

### 2. Block-Scoped Prior State Records (Robust Fix)
Since inverse arithmetic alone remains unsafe for lossy operations (like ERC1155 zero-clamping), store block-scoped snapshots of the affected addresses' prior states before applying changes, and restore them exactly during a rollback.

cranycrane:
Fixes #1617

## Problem

`disconnectBlockTxsEthereumType` walked the transactions of a block, and the token transfers inside each transaction, in forward order. Connect applies them chronologically, so for ERC-721 and ERC-1155 the disconnect was not a valid inverse. A token moving `A -> B -> C` inside one block (or inside one transaction, e.g. mint-then-transfer or a marketplace router hop) was re-inserted at `B` after its removal had already been a no-op. Whenever `B` had other history with the collection, the phantom holding survived in the address contracts index until a reindex.

## Fix

Split the disconnect into two passes:

- the existing forward pass keeps handling base tx data, internal data and the row deletes. The contract registry rollback from #1703 relies on seeing a same-block destruction before the re-creation, so this walk must stay forward.
- a new reverse pass over transactions and over the transfers within each transaction undoes the token holdings.

All counters (`Txs`, `TotalTxs`, `NonContractTxs`, `InternalTxs`) are order-independent, so nothing else changes. Reversing the whole walk, as the issue proposes, would instead have inverted the registry heuristic and dropped the row of a metamorphic contract redeployed in a later transaction of the disconnected block.

## Tests

`TestDisconnectBlockRange_MultiHopTransferRestoresHoldings` covers ERC-721 across two transactions, ERC-721 within one transaction, and ERC-1155 across two transactions, with the middle address holding prior history so the phantom is observable. All three sub-tests fail on master and pass with this change. `go test -tags unittest ./db/ ./api/` is green.

### DIFF
--- a/db/rocksdb_ethereumtype.go
+++ b/db/rocksdb_ethereumtype.go
@@ -1515,14 +1515,20 @@ func (d *RocksDB) disconnectBlockTxsEthereumType(wb *grocksdb.WriteBatch, height
 				return err
 			}
 		}
-		// internal data
+		// internal data - the contract registry rollback relies on this forward walk
 		err := d.disconnectInternalData(blockTx.btxID, height, addresses, contracts, registryRollback)
 		if err != nil {
 			return err
 
 		}
-		// contracts
-		for j := range blockTx.contracts {
+		wb.DeleteCF(d.cfh[cfTransactions], blockTx.btxID)
+		wb.DeleteCF(d.cfh[cfInternalData], blockTx.btxID)
+	}
+	// token holdings are undone in reverse chronological order, otherwise a token that
+	// moved A->B->C within the block ends up recorded at both A and B
+	for i := len(blockTxs) - 1; i >= 0; i-- {
+		blockTx := &blockTxs[i]
+		for j := len(blockTx.contracts) - 1; j >= 0; j-- {
 			c := &blockTx.contracts[j]
 			if err := d.disconnectAddress(blockTx.btxID, false, c.from, c, addresses, contracts); err != nil {
 				return err
@@ -1533,8 +1539,6 @@ func (d *RocksDB) disconnectBlockTxsEthereumType(wb *grocksdb.WriteBatch, height
 				}
 			}
 		}
-		wb.DeleteCF(d.cfh[cfTransactions], blockTx.btxID)
-		wb.DeleteCF(d.cfh[cfInternalData], blockTx.btxID)
 	}
 	for a := range addresses {
 		key := packAddressKey([]byte(a), height)
--- a/db/rocksdb_ethereumtype_test.go
+++ b/db/rocksdb_ethereumtype_test.go
@@ -255,6 +255,116 @@ func TestERC721SelfTransferConnectDisconnectRoundTrip(t *testing.T) {
 	}
 }
 
+// TestDisconnectBlockRange_MultiHopTransferRestoresHoldings verifies that a token moving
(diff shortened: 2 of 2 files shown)


cranycrane:
### Code review

53gur0 (approved):

PR #1776, opened 2026-09-08
fix(eth): stop reporting pending transferFrom as an NFT transfer (#1714)

### LINKED ISSUE #1712: eth: pending ERC-20 transfer with trailing calldata is not parsed, so the payee never sees it
`contractGetTransfersFromTx` derives a token transfer from calldata only when the payload is **exactly** `10+128` chars (`bchain/coins/eth/contract.go:305`). That path is the only source of token transfers for a **pending** transaction — with no receipt there are no logs to read (`bchain/coins/eth/ethparser.go:582-586`).

Stablecoin payment rails append a payment reference after the ABI arguments. Solidity ignores the extra bytes and the transfer executes normally, but Blockbook reports `tokens: []` for as long as the transaction sits in the mempool.

## Consequences

- Because no transfer is parsed, the recipient is never indexed (`bchain/mempool_ethereum_type.go:83-84`), so **the payee does not see the pending deposit at all**.
- Blockbook is internally inconsistent: its own ABI parser reports `transfer(address, uint256)` for the very payload that `contractGetTransfersFromTx` rejects on length.
- It also breaks Trezor Suite — see trezor/trezor-suite#31026, which carries the measured frequency (2.28% of `0xa9059cbb` calls over 60 mainnet blocks; ~1,700 genuine single ERC-20 transfers/day) and a live example tx.

## Why this is not a one-character fix

Relaxing `==` to `>=` on its own was simulated over the same 60 blocks in trezor/trezor-suite#31026 and produces **8.4 phantom transfers per correct one**: other functions' ABI heads decode as `(address,uint256)`, where the "recipient" is a dynamic-type offset word. Blockbook would then index those phantom recipients, injecting fake pending deposits into unrelated address histories.

## Suggested fix

Relax the length check **and** require `tx.To` to be a known fungible-token contract (`db/contract_info_cache.go` / `RocksDB.GetContractInfo`) before emitting the transfer.

Also bound the amount slice: `contract.go:311` currently slices `tx.Payload[10+64:]` to end of string, so it would parse the trailing reference into the amount as soon as the length check is relaxed. It needs to be `[10+64 : 10+128]`.

### LINKED ISSUE #1714: eth: pending ERC-20 transferFrom is reported as an NFT transfer (shared 0x23b872dd selector)
`bchain/coins/eth/contract.go:17` maps `0x23b872dd` to `erc721TransferFromMethodSignature`, and the branch at `contract.go:322-344` emits `Standard: bchain.NonFungibleToken` with `Value` read from the third argument.

ERC-20 and ERC-721 **share that selector** — `transferFrom(address,address,uint256)`. For ERC-20 the third argument is the transferred **amount**, not a token id.

## Consequence

A pending ERC-20 `transferFrom` is reported as an NFT transfer whose token id is the raw amount — so a history row reads as an NFT with id `100000000000000000000` instead of `100 TOKEN`. Put differently: pending ERC-20 `transferFrom` is not supported at all.

This path only runs while there is no receipt, so it self-corrects on confirmation (logs then give the real `Transfer` event). But with the private send-tx pending window at 3 h (#1638), the wrong row can now persist for hours instead of seconds.

## Second-order effect worth confirming

The wrong standard is passed to `GetContractInfo`, and both `api/worker.go:726-732` and `db/rocksdb_contracts.go:111-119` **persist** `standardFromContext` when the stored entry's standard is `UnknownTokenStandard`. So for a token whose standard the contract probe could not determine, a single pending `transferFrom` could permanently stamp it as ERC-721 in RocksDB, affecting every later transfer of that token. Worth verifying before deciding severity.

## Suggested fix

Choose the standard from the contract's known info rather than from the selector: fungible ⇒ `FungibleToken` with the amount as `Value`, ERC-721 ⇒ token id, unknown ⇒ emit nothing rather than guess.

cranycrane:
Fixes #1714.

## Problem

`contractGetTransfersFromTx`, the only source of token transfers for a **pending** transaction (no receipt, no logs), mapped `transferFrom(address,address,uint256)` = `0x23b872dd` to ERC-721 and reported the third argument as a token id. ERC-20 shares that selector, and on mainnet it is the common case: in a 30-block sample (25,900,000–25,900,029), 75 of 77 `transferFrom` calls were ERC-20 (USDC 34, USDT 28), 2 were ERC-721.

Without a receipt the standard cannot be told apart, so the shared selector is no longer parsed from calldata. The ERC-721-only `safeTransferFrom` selectors (`0x42842e0e`, `0xb88d4fde`) still are. Confirmed transactions always parse from receipt logs and are unaffected.

## Consequences for Suite users

While such a transaction was pending, Suite received `standard: "ERC721"` with the amount in the id slot, so:

- the history row rendered through the NFT formatter as e.g. "Token ID: 1717292743 USDC" instead of "1717.29 USDC";
- because the standard was ERC721, the fake-token phishing detector looked the contract up in the NFT definitions, did not find it, and flagged the whole receive as a potential fraud (blurred amount, warning icon);
- a token contract with no stored `cfContracts` row yet was stamped `ERC721` permanently via `standardFromContext` (`api/worker.go`, `db/rocksdb_contracts.go`), which nothing later corrects.

After this change a pending `transferFrom` shows no token transfer until it confirms, the same as a pending suffixed `transfer` today. Suite maps this selector to `'unknown'` in `getEvmTransactionTextSignature`, so nothing on the Suite side depends on the removed row. Pending ERC-721 transfers via plain `transferFrom` (the 2-of-77 minority) lose their pending row.

## Follow-up

- Resolve the standard from the stored contract info instead of the selector, so pending ERC-20 `transferFrom` gets a correct row and pending ERC-721 `transferFrom` comes back. Needs a DB-backed lookup reachable from the parser and the mempool address indexer; same hook fixes #1712 (suffixed `transfer` calldata) safely.
- Pass `UnknownTokenStandard` as `standardFromContext` for receipt-less transactions so no calldata-derived guess can ever be persisted.

## Testing

`go test -tags unittest ./bchain/ ./bchain/coins/eth/ ./bchain/coins/tron/ ./api/ ./db/` (httptest-bound tests skipped in the sandbox). The Tron parser test that pinned the old guess is updated the same way, since TRC-20 shares the selector too.

### DIFF
--- a/bchain/coins/eth/contract.go
+++ b/bchain/coins/eth/contract.go
@@ -14,7 +14,6 @@ import (
 )
 
 const erc20TransferMethodSignature = "0xa9059cbb"                  // transfer(address,uint256)
-const erc721TransferFromMethodSignature = "0x23b872dd"             // transferFrom(address,address,uint256)
 const erc721SafeTransferFromMethodSignature = "0x42842e0e"         // safeTransferFrom(address,address,uint256)
 const erc721SafeTransferFromWithDataMethodSignature = "0xb88d4fde" // safeTransferFrom(address,address,uint256,bytes)
 const erc721TokenURIMethodSignature = "0xc87b56dd"                 // tokenURI(uint256)
@@ -316,8 +315,9 @@ func contractGetTransfersFromTx(tx *bchain.RpcTransaction) (bchain.TokenTransfer
 			Value:    t,
 		})
 	} else if len(tx.Payload) >= 10+192 &&
-		(strings.HasPrefix(tx.Payload, erc721TransferFromMethodSignature) ||
-			strings.HasPrefix(tx.Payload, erc721SafeTransferFromMethodSignature) ||
+		// transferFrom(address,address,uint256) 0x23b872dd is deliberately absent: ERC-20 shares the
+		// selector and is the common case, so without a receipt the standard cannot be told apart
+		(strings.HasPrefix(tx.Payload, erc721SafeTransferFromMethodSignature) ||
 			strings.HasPrefix(tx.Payload, erc721SafeTransferFromWithDataMethodSignature)) {
 		from, err := addressFromPaddedHex(tx.Payload[10 : 10+64])
 		if err != nil {
--- a/bchain/coins/eth/contract_test.go
+++ b/bchain/coins/eth/contract_test.go
@@ -317,8 +317,21 @@ func Test_contractGetTransfersFromTx(t *testing.T) {
 			},
 		},
 		{
-			name: "ERC721 transferFrom",
+			// 0x23b872dd is shared by ERC-20 and ERC-721, so no standard can be derived from calldata
+			name: "transferFrom with the shared ERC-20/ERC-721 selector yields nothing",
 			args: (b2.Txs[2].CoinSpecificData.(bchain.EthereumSpecificData)).Tx,
+			want: bchain.TokenTransfers{},
+		},
+		{
+			name: "ERC721 safeTransferFrom",
+			args: &bchain.RpcTransaction{
+				From: "0x837e3f699d85a4b0b99894567e9233dfb1dcb081",
+				To:   "0xcda9fc258358ecaa88845f19af595e908bb7efe9",
+				Payload: erc721SafeTransferFromMethodSignature +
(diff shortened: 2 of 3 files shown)


53gur0 (approved):
EIP-165 would probably be the ultimate way to check the asset type

PR #1778, opened 2026-09-08
fix(ui): replace proprietary TT Hoves font with Inter (OFL-1.1)

### LINKED ISSUE #1060: Request for Font Replacement Due to Licensing Concerns
I am reaching out on behalf of the community with a pressing concern regarding the use of the "TT Hoves" font within the Blockbook instance. It has come to our attention that the licensing terms set forth by the owner of the "TT Hoves" font may not align with the open-source ethos that Blockbook embodies. Furthermore, the stance taken by the font's owner has unfortunately escalated into potential legal complexities that could affect the platform's development and its use by the community.

Issue Overview:

- Licensing Misalignment: The "TT Hoves" font, as it currently stands, is governed by licensing terms that are not compatible with the open-source nature of Blockbook. This misalignment raises concerns about the sustainability and freedom of use of the platform.
- Legal Risks: The owner of the "TT Hoves" font has expressed a reluctance to support the open-source model, which Blockbook prides itself on. This has led to potential legal challenges that could impede the platform's development and its mission to provide a free, open space for communication.

My tip for the replacement:
- [Satoshi font](https://www.fontshare.com/fonts/satoshi)

OriginalEveres:
Closes #1060.

TT Hoves is a proprietary TypeType typeface shipped in-tree with no accompanying licence grant, in an AGPL-3.0 repo. Whatever licence covers Trezor's own use doesn't settle it: committing the binaries redistributes them to every operator who runs an instance and self-hosts the files on their own domain, and foundry webfont licences are per-licensee and don't sublicense.

Replaced with [Inter](https://rsms.me/inter/) 4.1 under the **SIL Open Font License 1.1**, which explicitly permits bundling and redistribution provided the notice ships alongside (`OFL.txt` included).

- `static/css/TTHoves/` (18 files, ~1 MB) → `static/css/Inter/` (`InterVariable.woff2`, 352 KB)
- One variable file covers the full 100–900 range. `main.css` only used `normal` and `bold` and nothing uses italics, so it's a superset of what was there, at a third of the size.
- Cache-bust token `.min.4` → `.min.5` (convention from #1162) so clients don't keep a cached stylesheet pointing at removed font paths. Test fixtures updated.

**Not Satoshi**, suggested in the issue: ITF FFL v2.0 permits self-hosting for your own sites but forbids distributing the font via a *"repository"* or *"publicly accessible servers"*, and forbids subsetting or format conversion. Same blocker as TT Hoves.

Visual change — worth loading a page to check the rendering before merge.


server/public.go
```diff
@@ -457,7 +457,7 @@ func (s *PublicServer) newTemplateData(r *http.Request) *TemplateData {
 		t.MultiTokenName = bchain.EthereumTokenStandardMap[bchain.MultiToken]
 	}
 	if !s.debug {
-		t.Minified = ".min.4"
+		t.Minified = ".min.5"
```
cranycrane: we might update it again as I have already changed it in https://github.com/trezor/blockbook/pull/1777

PR #1784, opened 2026-09-10
Add CurveZMQ support for encrypted message queue binding

jhawk28:
ref #1318

At Quicknode, we support blockbook for a number of chains. In order to expose ZeroMQ publicly, it needs an auth based mechanism. Curve provides this capability. By adding support for curve to blockbook, it enables the use case where a customer can run their own blockbook and get the benefits of a zeromq sync without the complexity of running their own node. Our immediate benefit is that we can simplify configuration by pointing at one of our standard endpoints instead of going directly to nodes.

Curve is part of the ZMQ standard and most of the implementations support it. I suspect that the clients don't use it because they haven't needed to run in more corporate environments. As blockchain adoption increases, I expect that they will start including curve in their implementations.


cranycrane (approved):
LGTM, thank you for your contribution!

PR #1788, opened 2026-09-14
feat(eth): index the wallet-declared privatePending txids

### LINKED ISSUE #1773: eth: privatePending txids are ignored, so a pending tx vanishes when the client lands on a Blockbook that never saw it
## Summary

Suite sends `privatePending.txids` with every EVM `getAccountInfo` request so that Blockbook can
surface the wallet's in-flight transactions in history. Blockbook parses the field and ignores it.
The result is a second way a live pending transaction disappears from the account view, independent
of the null-eviction bug in #1709: a client that lands on a Blockbook instance which has never seen
the transaction gets a history page without it, and the wallet prunes it locally.

## Scenario

1. A user broadcasts a transaction through Blockbook A. A indexes it from the
   `newPendingTransactions` subscription and lists it as pending.
2. A restarts, or the load balancer moves the user, and the client reconnects to Blockbook B.
3. B's mempool index only holds what it learned since its own start: the node's **pending block**
   fetched once in `InitializeMempool` (`bchain/coins/eth/ethrpc.go:863`) plus subscription
   events after that. The pending block excludes queued / nonce-gapped transactions, and B knows
   nothing that lives only in a private relay's pool or on another node.
4. Suite's refresh sends the `basic` and then the `txs` request, both carrying the txid in
   `privatePending.txids` (built by `getEvmPrivatePendingHint`,
   `suite-common/wallet-utils/src/transactionUtils.ts:249`).
5. B answers a history page without the transaction. Suite's `analyzeTransactions` removes every
   locally known pending transaction that is absent from the fresh page
   (`suite-common/wallet-utils/src/transactionUtils.ts:733`). The transaction vanishes from the UI
   while it is still in the pool and still fetchable by hash.

Once Suite has dropped it, the hint no longer carries it either, so the loss is permanent from the
wallet's point of view until the transaction mines.

## Root cause

`server/websocket.go:1122` copies only the nonces out of the request:

```go
PrivatePendingNonces: privatePendingNonces(req.PrivatePending),
```

`WsPrivatePending.Txids` (`server/ws_types.go:56`) is documented as "reserved for future use" and
`docs/evm-send.md:204` states that only `nonces` drives behavior. Nothing reads the hashes, nothing
adds them to the mempool index, nothing injects them into the history page.

Suite's side of the contract was written against the intended behavior: the comment on
`getEvmPrivatePendingHint` says the txids "let it fetch-back each body (eth_getTransactionByHash)
so the tx shows up in history".

## Proposed fix

Honor the declared txids in the websocket `getAccountInfo` handler, before the account info is
built, so the same response already lists the transaction:

- Deduplicate and cap the txids the same way `privatePendingNonces` caps the nonces.
- For each txid **not already in the mempool index**, fetch it once through the chain's normal
  `GetTransaction`. If it is pending (no block), add it via
  `AddOrRefreshTransactionInMempool` (`bchain/mempool_ethereum_type.go:221`) so it is indexed and
  restamped. If it is mined or unknown, do nothing.
- Expose this as one EthereumType method on `EthereumRPC` with a no-op stub on `BaseChain`, and
  call it from the handler where the nonces are extracted.
- Update the `Txids` doc string and `docs/evm-send.md`.

Cost: one `eth_getTransactionByHash` per unknown txid per request, zero once indexed (the
existence check needs no RPC). A genuinely gone transaction costs two lookups, after which the
wallet stops declaring it.

### Requirements

- **Mined guard is mandatory.** `createTxEntry` does not check whether a transaction is mined, so
  adding blindly would index confirmed transactions.
- **Depends on #1709 / #1772.** Without it the next null answer from the backend evicts the
  re-added entry again. With it, a null no longer evicts and the nonce watermark retires the entry
  when its slot mines.

## Known limitation

A transaction that exists only in a private relay's pool cannot be recovered by B this way: the
alternative provider's `GetTransaction` serves only its own cache and does not ask the relay for
arbitrary hashes, so the public node answers null. Covering that means teaching the provider to
query the relay by hash, which is separate work.

## Out of scope

- The REST API. Suite uses the websocket, and the `privatePending` field exists only there.
- Any Suite change. Suite already sends the txids.

## Related

- #1709, #1772 (null-eviction fix this depends on)
- #1639 (introduced `privatePending`)

cranycrane:
Fixes #1773

**Stacked on #1772** (`fix/eth-mempool-nonce-watermark`) — this PR targets that branch, not master. Review only the last commit.

Suite sends `privatePending.txids` — the hashes of its own pending sends — with every EVM `getAccountInfo`, on both the `basic` probe and the `txs` page of each account refresh. Blockbook parsed the field and read only the nonces. A client that lands on an instance whose mempool index never saw the transaction gets a history page without it, and Suite's `analyzeTransactions` removes every locally known pending tx absent from the fresh page, with no re-fetch by hash. The transaction vanishes from the wallet while it is still in the pool and still fetchable by hash.

That instance is not exotic: its index holds only what it learned since its own start, and `disableMempoolSync` is set for Arbitrum, Base and other L2 configs, where it holds nothing but this instance's own broadcasts.

**Change**
- `privatePendingTxids` validates (`0x` + 64 hex), lowercases, dedupes and caps the declared list.
- `EthereumTypeAddPendingTransactions` fetches each unknown hash once — the relay cache first, then `eth_getTransactionByHash` — and indexes it only when it comes back without a block **and** sent from the queried address. Mined, unknown, foreign-sender and malformed hashes are ignored. A hash already in the index costs no backend call.
- Called from the `getAccountInfo` handler before the account info is built, so the same response lists the transaction and `subscribeAddresses` clients are notified. Best effort: the hint never fails the request.
- `createTxEntry` splits into a pure `txEntryFromTx` builder, so the fetched body is indexed without a second round trip. `GetTransaction` is deliberately not reused: its mined branch spends a block-header and a receipt lookup this path does not need.
- New counter `blockbook_eth_private_pending_txids_total{result}`.

**Deviation from the issue**: the issue proposed `AddOrRefreshTransactionInMempool`, which restamps. This uses a non-restamping add. Because the response then lists the transaction, Suite keeps declaring it, so a restamp would hold a permanently dropped transaction pending forever. Without it `mempoolTxTimeout` stays the server-side expiry: after it the next declaration re-verifies the hash at one round trip, and once the backend stops returning it the entry is not re-added and the wallet prunes it.

**Trust boundary**: the declared nonces stay per-request, but an indexed txid does enter the shared mempool index. Blockbook indexes only what the backend or relay actually returns as pending, so a client can name a transaction and never fabricate one; it must be sent from the queried address; the cap bounds the round trips under the connection's existing rate limit; and the result is what the `newPendingTransactions` subscription would have produced anyway. `docs/evm-send.md` carries the full argument.

Tron overrides the method as a no-op — its pending bodies come from the HTTP full node.

**Known limitation** (from the issue): a transaction living only in a private relay's pool still cannot be recovered by an instance that never saw it, since the provider serves only its own cache.

**Follow-up, not addressed here**: an address whose only activity is the declared pending transaction now shows the transaction, but still reports no nonce — `EthereumTypeGetNonces` runs only when the address has indexed data (`api/worker.go`).

**Tests**
- `bchain/coins/eth/private_pending_txids_test.go` — unknown pending tx indexed with exactly one `eth_getTransactionByHash` and no receipt call, `OnNewTx` fired; known txid costs zero calls and keeps its first-seen time; mined, unknown, foreign-sender and undecodable bodies all ignored; relay-cached body indexed with no backend call; no-op before the mempool is initialized.
- `server/websocket_privatepending_test.go` — `TestPrivatePendingTxids` covers normalization, validation, dedupe, cap and the defensive copy.
- `server/public_ethereumtype_test.go` — websocket round trip: declaring a hash the fake backend holds as pending returns it in `txids` with `unconfirmedTxs: 1`, while a malformed and an unknown hash alongside it change nothing.


cranycrane:
### Code review

Issue #1757, opened 2026-09-02
EVM sync: 3 of 4 block-fetch RPCs per new block are redundant (eth_getBlockByNumber)


cranycrane:
# EVM steady-state sync makes 3 redundant block-fetch RPCs per block

## Current state

On EVM chains, every new block notification (from the `newHeads` subscription, or the tip watchdog poll on chains without websockets) triggers `resyncIndex` in `db/sync.go`. For the common case — the chain advanced by exactly one block, no reorg — the per-block RPC sequence is:

| # | Step | Call | Backend RPC |
|---|------|------|-------------|
| 1 | get remote tip hash | `GetBestBlockHash()` | none (cached tip header) |
| 2 | fork check at local best height | `chain.GetBlockHash(localBestHeight)` in `resyncIndex` | `eth_getBlockByNumber(h, false)` |
| 3 | hash of the next block to sync | `chain.GetBlockHash(startHeight)` in `resyncIndex` | `eth_getBlockByNumber(h+1, false)` |
| 4 | gap size check | `GetBestBlockHeight()` | none (cached) |
| 5 | fetch the block | `getBlockChain` → `GetBlock(startHash, …)` | `eth_getBlockByHash(hash, true)` |
| 6 | probe for a following block | `getBlockChain` → `GetBlock("", h+2)` → `null` | `eth_getBlockByNumber(h+2, true)` |

That is **4 block-fetch RPCs per block, of which only step 5 delivers data Blockbook does not already have**:

- **Step 2** re-proves something the pushed tip header already proves: if the new tip's `parentHash` equals the locally indexed best hash, the indexed chain is by definition still canonical.
- **Step 3** downloads a whole header only to learn the new block's hash — which is the hash of the tip header that was just pushed.
- **Step 6** is an end-of-chain probe that almost always returns `null`.

The parent hash is already in memory on every tip update: the `newHeads` feed delivers full `*types.Header` objects (wrapped in `EthereumHeader`, `bchain/coins/eth/evm.go`), and the watchdog's HTTP poll gets the same header. It is just not exposed through the `bchain.EVMHeader` interface, which only has `Hash()/Number()/Difficulty()`.

## Proposed optimization

**Change 1 — gap==1 fast path (skips steps 2 and 3).** Expose `ParentHash()` on the EVM header (extend `bchain.EVMHeader`, or a narrow optional interface + type assertion in the sync worker), then in `resyncIndex` for `ChainEthereumType`:

```go
tip := cached best header
[...]

PR #1804, opened 2026-09-21
perf(sync): fetch parallel-sync blocks by height on EVM chains

cranycrane:
Follow-up to #1766, which deferred "the by-height parallel path"; continues #1757.

## Problem

`ParallelConnectBlocks` and `BulkConnectBlocks` resolved `GetBlockHash(h)` for every height on the single producer goroutine, then a worker fetched the same block again by hash. On EVM `GetBlock("", h)` is one `eth_getBlockByNumber` that already carries the body, so the header read was a wasted RPC per block, issued serially. #1766 measured that BSC takes the parallel path in ~47% of steady-state cycles, so this is where the remaining header reads on sub-second chains live.

## What changed

| Path | Before | After |
|---|---|---|
| EVM parallel/bulk producer | `eth_getBlockByNumber(h, false)` per height, serial | nothing: bare heights are queued |
| EVM worker | `eth_getBlockByHash(hash, true)` | `eth_getBlockByNumber(h, true)` |
| Bitcoin-type producer | `getblockhash` per height | unchanged (`GetBlock("")` would issue it internally anyway) |

Per block on the EVM parallel path this is 2 header+body RPCs → 1 body RPC; `eth_getLogs` and the trace are untouched.

**Reorg detection moves to the writer.** With no expected hash a missing block is no longer evidence of a reorg (`shouldRestartSyncOnMissingBlock` already short-circuits on an empty hash; the stall cap still bounds a lagging node). `writeBlockWorker` is the only goroutine that sees blocks in height order, so it now seeds the previous hash from the local best block and, when a block's `Prev` does not match, counts `index_reorg_events{type="fork"}`, yields `errResync` and stops before connecting the unlinked block. The next `resyncIndex` fails the `linked` shortcut, its `GetBlockHash(localBestHeight)` fork check finds the orphan and `handleFork` unwinds it: the existing round trip, no new code path. This also covers a load-balanced pool where one worker read height *h* from a lagging node and a sibling read *h+1* from a current one. The check runs only on by-height (EVM) rounds: a hash-pinned bitcoin-type round already follows the node's `getblockhash` chain, and the integration `HandleFork` fixtures splice non-linking blocks under real hashes and expect them to connect (this is what the first CI run tripped on).

**Late abort drain reordered.** Both coordinators drained `abortCh` before waiting for the writer, so a fork found while the writer drained the last blocks could be lost. The drain now runs after `writeBlockDone`. (A missed error would still have self-healed via the post-round `resyncIndex`, but now it is returned.)

## Tests

- `TestResyncIndexEthereumTypeParallelPathFetchesByHeight` (was `...ResolvesStartHashOnce`): a five-block parallel round makes zero `GetBlockHash` calls and one by-height `GetBlock` per block.
- `TestParallelConnectBlocksDetectsUnlinkedParent`: writer stops at the unlinked block, linked prefix is committed, `errResync` returned, fork counter +1.
- `TestResyncIndexEthereumTypeParallelForkRecovers`: end-to-end reorg between worker fetches → abort → fork check → `handleFork` → second parallel round lands the canonical branch; no header lookups above the fork height.
- `TestBulkConnectBlocksBitcoinTypeKeepsHashLookup` / `TestBulkConnectBlocksEthereumTypeFetchesByHeight`: the chain-type gate on the bulk path (first direct tests of `BulkConnectBlocks`).
- `TestParallelConnectBlocksBitcoinTypeIgnoresUnlinkedParent`: the fixture shape, non-linking parents under pinned hashes connect with no fork event.
- `TestUnlinkedParent` table test.
- Two existing tests that built a `SyncWorker` without a DB now get one, since the writer reads the local best block once per round.
- `go test -race -tags unittest ./db/` passes.

## Docs

`docs/sync.md` gains an invariant bullet for the by-height producers and the writer's linkage check; the `RecheckThreshold` row notes it is a no-op for hashless EVM fetches.

## Not in this PR

The sequential general-gap variant from #1757 (gaps of 2–3 on polled chains) is the only remaining item there.


cranycrane:
### Code review

PR #1807, opened 2026-09-22
feat(eth): index wrapped-native Deposit/Withdrawal as token transfers

### LINKED ISSUE #1702: eth: wrapping native asset (WETH deposit/withdraw) produces no tokenTransfers — wrapped token is not auto-discovered on the address
## Summary

Wrapping the native asset (e.g. `WETH.deposit()`) produces no `tokenTransfers` entry in Blockbook, because canonical wrapped-native contracts (WETH9 and its ports) mint without emitting an ERC-20 `Transfer` event — they emit `Deposit(address indexed dst, uint256 wad)` instead. Blockbook's log parser only recognizes `Transfer` / `TransferSingle` / `TransferBatch`, so:

1. the wrap transaction is indexed as a plain ETH send to the contract, with no indication the address received WETH,
2. **token auto-discovery never happens** — the address↔contract association is built from token transfers only, so an address whose only WETH interaction is wrapping never gets WETH into its `tokens` list and its WETH balance stays invisible until some unrelated `Transfer` event touches the address,
3. the same applies to unwrapping: `withdraw()` emits `Withdrawal(address indexed src, uint256 wad)`, also ignored, and per-address `transfers` counts for WETH exclude all wraps/unwraps.

In clients this means e.g. Trezor Suite renders a wrap as nothing but an outgoing ETH payment to an unknown contract — there is no "received X WETH" line to show, while Etherscan displays the same tx as "Wrap 0.000123 ETH into 0.000123 WETH" (it parses the `Deposit` event itself).

Image
Image

Temporary fix for Trezor Suite https://github.com/trezor/trezor-suite/pull/30797, https://github.com/trezor/trezor-suite/issues/29875

## Examples

- https://eth.trezor.io/tx/0xdb59b35f7ba6c9545e14d2b2493120572f50b0da308e1c131971b543e1c65aa1
- https://eth.trezor.io/tx/0x77706244fde779e24a5d51ab497d423703eec99f77228b63966aca445832dbe0

Both are `deposit()` (`0xd0e30db0`) calls sending ETH to WETH `0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2`.

`/api/v2/tx-specific/0xdb59b3…` shows the receipt contains exactly one log — the `Deposit` event, no `Transfer`:

```json
"logs": [{
  "address": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
  "topics": [
    "0xe1fffcc4923d04b559f4d29a8bfc6cda04eb5b0d3c460751c2402c5c5cc9109c",
    "0x0000000000000000000000009ea3721b5bf3b64b4418c38b603154d2d597fae3"
  ],
  "data": "0x00000000000000000000000000000000000000000000000000038d7ea4c68000"
}]
```

`/api/v2/tx/…` consequently returns no `tokenTransfers` field at all for either tx (verified on both), even though `parsedData` correctly identifies the call as `deposit()`.

(The example address does show WETH in `tokens` today, but only because later swap/send txs emitted real `Transfer` events; an address that only ever wraps gets nothing.)

## Root cause

`contractGetTransfersFromLog` dispatches solely on the three transfer topics and `continue`s on everything else:

https://github.com/trezor/blockbook/blob/89963a3972e363aaa054e3fe46953309990b5dca/bchain/coins/eth/contract.go#L274-L300

WETH9 `deposit()`/`withdraw()` mint/burn without `Transfer`:

```solidity
function deposit() public payable {
    balanceOf[msg.sender] += msg.value;
    Deposit(msg.sender, msg.value);
}
```

## Suggested fix

Translate the wrapped-native contract's events into synthetic mint/burn token transfers:

- `Deposit(dst, wad)` → `TokenTransfer{from: 0x0, to: dst, value: wad}`
- `Withdrawal(src, wad)` → `TokenTransfer{from: src, to: 0x0, value: wad}`

Since any contract can emit events with these signatures (same spoofing concern as #1225), the translation should be restricted to the canonical wrapped-native contract per chain, e.g. a per-coin config field (`wrappedNativeContract`). That would cover WETH on Ethereum and the same WETH9-pattern contracts on other EVM chains Blockbook indexes (WBNB, WPOL, WXDAI, …).

cranycrane:
## Description

WETH9-style contracts mint and burn without a `Transfer` event, so a wrap was indexed as a plain native send: no `tokenTransfers`, and the wrapped token was never auto-discovered on the address (#1702).

- New per-coin `wrappedNativeContract` (`additional_params`), validated at startup and applied to the parser via `ApplyToParser`.
- `Deposit(dst, wad)` → ERC-20 mint `0x0 → dst`, `Withdrawal(src, wad)` → burn `src → 0x0`, only for the configured contract (any contract can emit these signatures). The address compare runs after the topic match, so the hot path is unchanged.
- Same for pending `deposit()` / `withdraw(uint256)` calls to that contract; a plain `transfer()` of the wrapped token still takes the ERC-20 path.
- Configured for Ethereum, Polygon, Avalanche, Optimism, Base, BSC, Ethereum Classic, HyperEVM and Sepolia after checking one wrap receipt per chain. **Arbitrum, Robinhood Chain and Hoodi are deliberately not configured**: their wrapped token already emits a mint `Transfer`, so the key would double count every wrap.

## Notes

- Receipt logs are stored per tx, so `/api/v2/tx` shows the new legs for existing wraps immediately. Address-level discovery and the per-token `transfers` count apply to newly synced blocks; existing blocks need a reindex.
- Client-visible change: wrap/unwrap txs gain a `tokenTransfers` entry with the zero address as counterparty.

Closes #1702


cranycrane:
### Code review

